PR #229, opened 2016-06-14
Case studies

saralohr:
Add case study core and single page designs and images to library.

I exported the images to multiple sizes and was going to use `srcset` and `sizes` but then I remembered that Wordpress does that for you. Currently just using two different size images on the single page vs the core.

case-study-core
case-study-single


tylersticka:
Nit: Can we put the case study images in a subdirectory of `images/`?

Apologies if I'm a dummy, but what's the origin of the `@0,5x` naming convention?

saralohr:
> Nit: Can we put the case study images in a subdirectory of images/?

Sure.

> Apologies if I'm a dummy, but what's the origin of the @0,5x naming convention?

This is the file format that exports from Photoshop to reference the size as a percentage of the original, where the comma is a decimal. However, it actually breaks when used in a `srcset` list since commas are part of the syntax. I had changed these to dashes, but then I moved stuff around and that must have gotten lost. I'm going to change them to something different.

tylersticka:
LGTM after those changes!

saralohr:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;S1EMXmbO&#x2F;174-mockups-prototypes-what-we-ve-done-core"> Mockups&#x2F;Prototypes: What We&#39;ve Done (Core)</a>

saralohr:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;1INQ9zmB&#x2F;175-mockups-prototypes-case-study-single"> Mockups&#x2F;Prototypes: Case Study (Single)</a>

PR #230, opened 2016-06-14
Update team photo

saralohr:
our-team


src/assets/toolkit/styles/utils/text.css
```diff
@@ -16,6 +16,16 @@
...
 
+
+/**
+ * Text color
+ */
+
+.u-textWhite {
+  color: var(--color-white);
+}
+
```
tylersticka: Was this meant to be included? Not sure how it relates to the photo change.

tylersticka:
One question inline. Otherwise LGTM.

src/assets/toolkit/styles/utils/text.css
saralohr: Doesn't relate to photo change, but I had removed it from my WIP stylesheet in the last PR, and it was being used by the header text, since the light purple doesn't have enough contrast.

tylersticka: Does that mean this text is now inconsistent with other instances of header text?

If we need to adjust contrast for _all_ the headers, I'd rather do that.

saralohr: Okay, I removed that change from this PR and will address contrast in another one.

PR #233, opened 2016-06-15
WIP: Home prototype header

tylersticka:
This PR adds a WIP home prototype to the sandbox that showcases the current state of the homepage animation and accompanying header text.

The animation is entirely separate from the rest of the toolkit. The imagery exists in an external SVG, which then references a JS file that is separate from our main `toolkit.js`.

For performance and, honestly, usability of constructing a timeline-based animation, I chose to use GSAP. This means the animation JavaScript adds up to a whopping `318kb` uncompressed. This compresses down to `138kb` with Uglify, and further down to `37kb` when gzipped. The gzipped SVG is only `918b`, so we're looking at a total `38kb` transferred for the animation currently, which seems like a reasonable amount to load (and far lighter than the average full-bleed photo).

home-2016-06-15-0934

---

@saralohr @nicolemors @erikjung @mrgerardorodriguez

### DIFF
--- a/src/assets/toolkit/scripts/home-animation.js
+++ b/src/assets/toolkit/scripts/home-animation.js
@@ -5,7 +5,11 @@
  * See: https://github.com/greensock/GreenSock-JS/issues/71
  */
 
-import TweenMax from 'gsap';
+import TweenLite from 'gsap/src/uncompressed/TweenLite';
+import TimelineMax from 'gsap/src/uncompressed/TimelineMax';
+import EasePack from 'gsap/src/uncompressed/easing/EasePack';
+import CSSPlugin from 'gsap/src/uncompressed/plugins/CSSPlugin';
+import AttrPlugin from 'gsap/src/uncompressed/plugins/AttrPlugin';
 
 /**
  * Settings
@@ -49,7 +53,7 @@ const dom = (function (nodeList) {
   return result;
 })(document.querySelectorAll('*[id]'));
 
-TweenMax.set('.js-scaleIn', {
+TweenLite.set('.js-scaleIn', {
   visibility: 'visible',
   scale: 0,
   transformOrigin: '50% 50%'
--- a/src/assets/toolkit/scripts/home-animation.js
+++ b/src/assets/toolkit/scripts/home-animation.js
@@ -39,7 +39,6 @@ const timelines = {
 
 const dom = (function (nodeList) {
   const result = {};
-  nodeList = Array.prototype.slice.call(nodeList);
 
   for (let i = 0; i < nodeList.length; i++) {
     let node = nodeList[i];
--- a/src/assets/toolkit/scripts/home-animation.js
+++ b/src/assets/toolkit/scripts/home-animation.js
@@ -1,0 +1,296 @@
+/**
+ * GSAP is powerful and performant, but its organization is a little old-school.
+ * This imports a bunch of functions to the global scope.
+ *
+ * See: https://github.com/greensock/GreenSock-JS/issues/71
+ */
+
+import TweenMax from 'gsap';
+
+/**
+ * Settings
+ */
+
+const settings = {
+  morph: {
+    delay: '+=1',
+    duration: 0.5,
+    ease: Elastic.easeOut.config(0.5, 0.4)
+  }
+};
+
+/**
+ * Timelines
+ */
+
+const timelines = {
+  morph: new TimelineMax({
+    repeat: -1
+  })
+};
+
+/**
+ * Elements
+ */
+
+const dom = (function (nodeList) {
+  const result = {};
+  nodeList = Array.prototype.slice.call(nodeList);
+
+  for (let i = 0; i < nodeList.length; i++) {
+    let node = nodeList[i];
+    let id = node.getAttribute('id');
+
+    if (id) {
+      result[id] = node;
+    }
+  }
+
+  return result;
+})(document.querySelectorAll('*[id]'));
+
+TweenMax.set('.js-scaleIn', {
+  visibility: 'visible',
+  scale: 0,
+  transformOrigin: '50% 50%'
+});
+
+/**
+ * Device morphing
+ */
+
+timelines.morph.addLabel('toTablet', settings.morph.delay);
+
+timelines.morph.to(dom.bezel, settings.morph.duration, {
+  attr: {
+    x: 94,
+    y: 66,
+    width: 132,
+    height: 188
+  },
+  ease: settings.morph.ease
+}, 'toTablet');
+
+timelines.morph.to(dom.screenMaskMain, settings.morph.duration, {
+  attr: {
+    x: 100,
+    y: 80,
+    width: 120,
+    height: 160
+  },
+  ease: settings.morph.ease
+}, 'toTablet');
+
+timelines.morph.addLabel('toTabletLandscape', settings.morph.delay);
+
+timelines.morph.to([dom.bezel, dom.screenMaskMain], settings.morph.duration, {
+  attr: {
+    transform: 'rotate(-90 160 160)'
+  },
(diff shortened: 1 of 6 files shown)


tylersticka:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;NcrR6BI2&#x2F;173-mockups-prototypes-homepage"> Mockups&#x2F;Prototypes: Homepage</a>

saralohr:
Looking great so far! 👍

src/assets/toolkit/scripts/home-animation.js
```diff
@@ -0,0 +1,300 @@
...
+ * This imports a bunch of functions to the global scope.
+ *
+ * See: https://github.com/greensock/GreenSock-JS/issues/71
+ */
+
+import TweenLite from 'gsap/src/uncompressed/TweenLite';
+import TimelineMax from 'gsap/src/uncompressed/TimelineMax';
+import EasePack from 'gsap/src/uncompressed/easing/EasePack';
+import CSSPlugin from 'gsap/src/uncompressed/plugins/CSSPlugin';
+import AttrPlugin from 'gsap/src/uncompressed/plugins/AttrPlugin';
```
erikjung: That is unfortunate. The following does work, but because it uses the main library (`TweenMax`) the resulting file is 1300 lines longer.

```js
import {
  default as TweenLite,
  TimelineMax,
  EasePack,
  CSSPlugin,
  AttrPlugin
} from 'gsap';
```

tylersticka: Yeah, that's because the `default` is `TweenMax`, which includes all that other stuff.

src/assets/toolkit/scripts/home-animation.js
```diff
@@ -0,0 +1,300 @@
...
+  })
+};
+
+/**
+ * Elements
+ */
+
+const dom = (function (nodeList) {
+  const result = {};
+  nodeList = Array.prototype.slice.call(nodeList);
```
erikjung: Are we polyfilling `Array.from()`?

tylersticka: Yes, but not in this context. The SVG is its own document.

That said, I'm glad you mentioned this because the line isn't even necessary... I'm just using a `for` loop, so I don't need to convert the `nodeList` at all.

src/assets/toolkit/scripts/home-animation.js
```diff
@@ -0,0 +1,299 @@
...
+
+timelines.morph.to(dom.bezel, settings.morph.duration, {
+  attr: {
+    x: 94,
+    y: 66,
+    width: 132,
+    height: 188
+  },
+  ease: settings.morph.ease
+}, 'toTablet');
```
mrgerardorodriguez: Could instances, such as `'toTablet'`, be pulled out into a `const` setting object maybe?

tylersticka: I'm not sure I understand the suggestion.

mrgerardorodriguez: I was wondering if these string values could be DRYed out at all? :)

src/static/images/sandbox/home-animation.svg
```diff
@@ -0,0 +1,60 @@
...
+
+  <g id="remote" class="u-hidden js-scaleIn">
+    <rect class="u-fillDark" x="236" y="216" width="36" height="88" rx="4" ry="4"/>
+    <circle class="u-strokeMid" stroke-width="4" cx="254" cy="242" r="9"/>
+    <circle class="u-fillMid" cx="245" cy="264" r="3"/>
+    <circle class="u-fillMid" cx="254" cy="264" r="3"/>
+    <circle class="u-fillMid" cx="263" cy="264" r="3"/>
+  </g>
+
+  <script xlink:href="/assets/toolkit/scripts/home-animation.js"/>
```
mrgerardorodriguez: Interesting! I hadn't considered doing this. Neat idea.

src/assets/toolkit/scripts/home-animation.js
```diff
@@ -0,0 +1,299 @@
...
+    let node = nodeList[i];
+    let id = node.getAttribute('id');
+
+    if (id) {
+      result[id] = node;
+    }
+  }
+
+  return result;
+})(document.querySelectorAll('*[id]'));
```
erikjung: Assuming we do have `Array.from()` (it appears we do via polyfill.io), this could be simplified down to:

```js
return Array.from(nodeList).reduce((result, node) => {
  result[node.id] = node;
  return result;
}, {});
```

If the check for `id` is needed, this chain could also include:

```js
.filter(node => node.hasAttribute('id'))
```

But because the query selector includes `[id]`, I'm not sure this is necessary (?).

src/pages/sandbox/tyler-home-a.hbs
```diff
@@ -0,0 +1,45 @@
...
+
+{{{embed "patterns.components.sky.base"}}}
+
+<main role="main" id="main">
+  {{#embed "patterns.components.sky.base" class="Sky--clouds"}}
+    {{#content "top"}}
+      <div class="Grid Grid--alignMiddle">
+        <div class="Grid-cell u-md-size1of2 u-md-flexOrderLast">
+          <object data="/images/sandbox/home-animation.svg" type="image/svg+xml" style="width: 100%; height: 20em;">
+            {{! TODO: Fallback (GIF, maybe?) }}
```
mrgerardorodriguez: A GIF seems reasonable. :)

src/assets/toolkit/scripts/home-animation.js
erikjung: Ah, then feel free to disregard my other code comment about this function.

src/assets/toolkit/scripts/home-animation.js
tylersticka: I dunno. Unlike the other `settings`, these are really just shorthands relative to the current chunk of animation. I feel like moving those labels to somewhere else that has to be managed separately (like in the `settings` object) will make the individual animation components tougher to mess with.

There are ways of instead adding an array of tweens to the timeline all at once, but they're all more verbose.

src/assets/toolkit/scripts/home-animation.js
mrgerardorodriguez: Oh, nice @erikjung. 🔥 🔥 🔥

erikjung:
@tylersticka 👍 LGTM as far as the JS goes.

I'm not sure what to suggest at the moment regarding any _better_ way to package and deliver the JS for this animation. It's a lot of JS, but as you mentioned, the payload isn't bad. 

I wish GSAP had a more modular organization (or the ability to make subset builds).

src/assets/toolkit/scripts/home-animation.js
mrgerardorodriguez: Cool, just curious. Thanks for explaining. 👍

mrgerardorodriguez:
You are doing _awesome_ work @tylersticka! LGTM! 👍

src/assets/toolkit/scripts/home-animation.js
tylersticka: But remember, this is in _a separate document_. So we don't have _any_ polyfills initialized in the HTML page. This script is loaded by and only exists in the SVG document, which is embedded via `<object>`.

I don't think it's worth including an additional dependency within the SVG document to save 8 lines of JavaScript. But maybe I'm wrong about that.

tylersticka:
> I wish GSAP had a more modular organization (or the ability to make subset builds).

Yeah, me too!

FWIW, though, what I'm doing here _is_ basically a subset build. Originally I was doing `import TweenMax from 'gsap';`, which includes the whole she-bang. The `import` lines I'm using now include _only_ what I need.

PR #235, opened 2016-06-15
Team page tweaks; height utilities

saralohr:
A few more team page tweaks:
- Simplify text to a single paragraph
- Make image full-width at small sizes
- Crop image height based on `vh`
- Add some utilities to do this ^^ 

Open to suggestions on naming conventions for these. I wanted to distinguish between our viewport _width_ prefixes, so I went with `.u-[prefix]vh-limitHeight[value]`

screen shot 2016-06-15 at 1 20 26 pm
screen shot 2016-06-15 at 1 20 33 pm
screen shot 2016-06-15 at 1 20 53 pm

cc @tylersticka @mrgerardorodriguez @erikjung


saralohr:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;HMlPSX0j&#x2F;182-address-medium-viewport-team-page-feedback"> Address medium viewport Team page feedback</a>

src/assets/toolkit/styles/utils/height.css
```diff
@@ -0,0 +1,11 @@
+/* Set height limits with vh units */
+
+.u-limitHeight75 {
```
tylersticka: Maybe use `3of4` style instead of `75` for consistency with size utilities?

saralohr: Oh, good idea!

src/assets/toolkit/styles/utils/height.css
```diff
@@ -0,0 +1,11 @@
...
+  max-height: 75vh !important;
+}
+
+@media (min-height: 550px) {
+  .u-mdvh-limitHeight50 {
```
tylersticka: I'm not sure `vh` is the right addition, because the media query itself has nothing to do with `vh` (only the rule that's applied).

I'd be fine with `mdh`. It's not _great_, but it's only one more character.

We should probably define the media query where the others ones are in our theme, though.

src/assets/toolkit/styles/utils/index.css
```diff
@@ -4,6 +4,7 @@
 @import "./contain.css";
 @import "./display.css";
 @import "./flex.css";
+@import "./height.css";
```
tylersticka: Would we consider these to be `size` utilities?

tylersticka:
Some suggestions inline. Otherwise LGTM 👍

src/assets/toolkit/styles/utils/index.css
saralohr: Sure, that makes sense.

src/assets/toolkit/styles/base/theme.css
```diff
@@ -178,9 +178,13 @@
...
+/* Width */
 @custom-media --sm-viewport (width >= 30em);
 @custom-media --md-viewport (width >= 48em);
 @custom-media --lg-viewport (width >= 64em);
 @custom-media --xl-viewport (width >= 80em);
 @custom-media --xxl-viewport (width >= 100em);
 @custom-media --xxxl-viewport (width >= 120em);
+
+/* Height */
+@custom-media --mdh-viewport (height >= 35em);
```
erikjung: Not crazy about this prefix, but also unable to propose a better alternative for abbreviations.

Since we're likely to not have as many height-based queries, what about this:

```css
@custom-media --short-viewport (height < 35em);
@custom-media --tall-viewport (height >= 35em);
```

erikjung: Eh, this is also flawed though. Once we need a third.

PR #238, opened 2016-06-17
Add inline-svg postcss plugin and use it for sky clouds

saralohr:
Use PostCSS plugin to create inline SVG on the fly. Specifically for use on sky clouds since they are a background image.

screen shot 2016-06-17 at 8 05 44 am


src/assets/toolkit/styles/components/sky.css
```diff
@@ -5,7 +5,7 @@
   --Sky-background-color-dark: color(var(--color-navy) h(+5) s(-12) l(+18));
   --Sky-color: var(--color-white);
 
-  --Sky-clouds-image: url("/images/sky-clouds.svg");
+  --Sky-clouds-image: svg-load("../../../../static/images/sky-clouds.svg");
```
tylersticka: You might consider specifying the `path` option to simplify this.

See [here](https://github.com/TrySound/postcss-inline-svg#options) and [here](https://github.com/postcss/postcss-use#standard-syntax) for info.

tylersticka:
One suggestion inline. Otherwise LGTM 👍

tylersticka:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;n206gSOy&#x2F;143-chore-inline-svg-icons-instead-of-using-a-sprite"> Chore: Inline SVG icons instead of using a sprite</a>

src/assets/toolkit/styles/components/sky.css
saralohr: @tylersticka changed this up per your suggestion.

PR #242, opened 2016-06-20
WorkImage component for case studies

tylersticka:
This PR implements a new `WorkImage` component for case studies using SVG and CSS.

## The Problem

The previous images were prepared using Illustrator and Photoshop, and exported to a couple different sizes. This came with a few drawbacks.

Most manual processes are inherently more error-prone. Although only five case study images were prepared (and with thoughtfulness and care), one of them already had part of an image's gray background extending past the bounds of the browser frame:

screen shot 2016-06-20 at 3 51 27 pm

A minor and understandable mistake, but one a computer wouldn't have made... not that computers are infallible.

Here, we see Photoshop has decided to give the bottom of the tilted iPhone a hard edge as it approaches the canvas, which makes sense for this size and angle:

hautelook-475

But once we apply a rotation to the image, the browser struggles to smooth out that edge in a way that feels natural:

screen shot 2016-06-20 at 3 56 09 pm

It's not uncommon for certain screenshots or other imagery to appear "fuzzy" if it's capturing a website or event at a specific moment in time. But the rasterized nature of the outer frames mean even though design elements we are in full control over start looking a little pixellated on high-res screens, like that of the iPad Air:

simulator screen shot jun 20 2016 4 01 02 pm copy

All those user-facing issues aside, let's face it... nobody likes spelunking through PSDs and manually exporting stuff. It'd sure be nicer if we could just provide a "small" and "large" image.

## The Solution

The `WorkImage` component solves these problems. It uses SVG to define the shapes and masks for the image frames, and CSS to control their visual presentation (tilt, composition, fill, etc.).

### Small, Tilted Comparison

Before | After
--- | ---
screen shot 2016-06-20 at 4 06 14 pm | screen shot 2016-06-20 at 4 06 01 pm

### iPad Comparison

Before | After
--- | ---
simulator screen shot jun 20 2016 4 01 02 pm | simulator screen shot jun 20 2016 4 06 48 pm

### Differences Comparison

Small versions of all five case study images overlayed with "difference" blend mode to highlight inconsistencies.

Before | After
--- | ---
differences-old | differences-new

## Performance

A single `WorkImage` instance will often beat the previous solution in both loading and rendering time. This is because the browser can display the basic SVG frame without loading anything, and can load the two screenshots asynchronously.

As the number of `WorkImage` instances grows, the rendering time will still be better, but the greater number of requests will cause the overall loading time to be a bit slower. This can be improved by resizing or tweaking optimization of images. I've avoided resizing imagery here because, assuming this PR is merged, we'll be able to control some of that optimization on the server.

## Other Notes

The previous implementation of these images introduced a `u-rotateBack3` utility that tilted an object by three degrees. I chose to replace this with a `WorkImage--tilt` modifier, because the amount seemed too idiosyncratic to be self-explanatory outside the context of the component.

I made a few tweaks to some of the screenshots prior to exporting new versions. I removed the blank advertising space from the WM screenshot, I restored the visibility of the top of the CO page, I removed part of some pasted-over text from Hautelook, etc.

---

@saralohr @mrgerardorodriguez @erikjung


tylersticka:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;WMOtuWp2&#x2F;187-component-make-case-study-images-dynamic"> Component: Make case study images dynamic</a>

mrgerardorodriguez:
LGTM! 👍

saralohr:
👍

PR #243, opened 2016-06-21
Add styles for WP generated classes

saralohr:
This PR adds styles for the WP generated classes, and accounts for some width and height inline styles and attributes applied by Wordpress that make for images that don't scale proportionally (or, at all).

For the `.Figure` component, I made it a mixin so that these styles could be applied to the WP `<figure>` class, oddly named `wp-caption` (I believe it's named this because images are only wrapped in a `<figure>` element if they have been provided a caption within the WP media dialogue).

cc @mrgerardorodriguez @tylersticka @erikjung @aileenjeffries

### DIFF
--- a/src/assets/toolkit/styles/vendor/wordpress.css
+++ b/src/assets/toolkit/styles/vendor/wordpress.css
@@ -1,0 +1,92 @@
+/* Wordpress generated classes */
+
+.aligncenter,
+div.aligncenter,
+a img.aligncenter {
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
+}
+
+.alignright,
+a img.alignright {
+  float:right;
+  margin-left: var(--space-sm);
+  margin-bottom: var(--space-sm);
+}
+
+.alignleft,
+a img.alignleft {
+  float: left;
+  margin-right: var(--space-sm);
+  margin-bottom: var(--space-sm);
+}
+
+/* Image scaling
+ * Make sure images with WordPress-added height and width
+ * attributes are scaled correctly
+ */
+
+img[class*="align"],
+img[class*="wp-image-"] {
+  height: auto;
+}
+
+/* `wp-caption` is the class Wordpress applies to a figure element.
+ *  1. Reuse `.Figure` component styles.
+ *  2. Override the `width` inline style  applied to `wp-caption` by Wordpress
+ */
+
+.wp-caption {
+  @mixin Figure; /* 1 */
+  width: 100% !important; /* 2 */
+}
+
+/**
+ * Override the inline fixed width and height styles applied by wordpress
+ */
+
+.wp-caption img {
+  width: auto !important;
+  max-width: 100% !important;
+  height: auto !important;
+}
+
+/* Reuse styles from the `.Figure-caption` pattern.
+ *  1. By default reduce text size.
+ */
+
+.wp-caption-text {
+  @mixin Figure-caption;
+  font-size: calc(1em * var(--ms-1)); /* 1 */
+}
+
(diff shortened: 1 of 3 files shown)


src/assets/toolkit/styles/vendor/wordpress.css
```diff
@@ -0,0 +1,92 @@
+/* Wordpress generated classes */
```
tylersticka: Can we add the URL to the codex CSS page to these comments for future generations?

src/assets/toolkit/styles/vendor/wordpress.css
```diff
@@ -0,0 +1,92 @@
...
+}
+
+.alignleft,
+a img.alignleft {
+  float: left;
+  margin-right: var(--space-sm);
+  margin-bottom: var(--space-sm);
+}
+
+/* Image scaling
```
tylersticka: Nit: Missing leading `/**` in block comments.

src/assets/toolkit/styles/vendor/wordpress.css
```diff
@@ -0,0 +1,92 @@
...
+ * Make sure images with WordPress-added height and width
+ * attributes are scaled correctly
+ */
+
+img[class*="align"],
+img[class*="wp-image-"] {
+  height: auto;
```
tylersticka: Instead of this, can we do [this](https://css-tricks.com/snippets/wordpress/remove-width-and-height-attributes-from-inserted-images/)?

src/assets/toolkit/styles/vendor/wordpress.css
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/**
+ * Override the inline fixed width and height styles applied by wordpress
+ */
+
+.wp-caption img {
+  width: auto !important;
+  max-width: 100% !important;
+  height: auto !important;
```
tylersticka: See [this comment](https://github.com/cloudfour/cloudfour.com-patterns/pull/243/files#r67788833).

src/assets/toolkit/styles/vendor/wordpress.css
```diff
@@ -0,0 +1,92 @@
...
+/* Reuse styles from the `.Figure-caption` pattern.
+ *  1. By default reduce text size.
+ */
+
+.wp-caption-text {
+  @mixin Figure-caption;
+  font-size: calc(1em * var(--ms-1)); /* 1 */
+}
+
+/* Text meant only for screen readers. */
```
tylersticka: It's a shame we can't re-use more of this utility: https://github.com/cloudfour/cloudfour.com-patterns/blob/master/src/assets/toolkit/styles/utils/display.css#L7-L15

I'm not really sure _how_ off-hand. I'm just saying it's a shame. 😅

src/assets/toolkit/styles/vendor/wordpress.css
saralohr: This was something that I had tried with Ceasefire, that seemed to be working at first, and then later not, for...reasons. Instead I resolved with CSS. I'm open to trying that path again, though.

tylersticka: Gotcha. That's fine. 👍

saralohr:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;nFGuIPP4&#x2F;196-how-to-handle-wordpress-image-classes-plus-our-own-classes-and-components-in-blog-entry-both-new-and-old"> How to handle Wordpress image classes, plus our own classes and components in blog entry (both new and old)</a>

PR #246, opened 2016-06-21
Syntax highlighting

tylersticka:
Adds syntax highlighting to the toolkit. Also removes syntax highlighting from Drizzle, since it can now inherit the feature from the toolkit.

I tried out and evaluated [Microlight](http://asvd.github.io/microlight/) and [Highlight.js](https://highlightjs.org/). The former lacked the features necessary to achieve something more attractive than the default, and the source was difficult to modify. The latter required a build script to customize, and was larger in file size than either of the others.

[Prism](http://prismjs.com/) had the lowest file size and was also easiest to theme. It was also a bonus that a version was already included in our library.

I chose to only customize classes I could anticipate us using. We can expand the theme in the future as our content needs expand.

## Before

screen shot 2016-06-21 at 4 26 52 pm

screen shot 2016-06-21 at 4 34 00 pm

## After

screen shot 2016-06-21 at 4 26 38 pm

screen shot 2016-06-21 at 4 33 49 pm

---

@mrgerardorodriguez @erikjung


tylersticka:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;j8DYNCJZ&#x2F;190-component-syntax-highlighting"> Component: Syntax highlighting</a>

src/assets/toolkit/styles/toolkit.css
```diff
@@ -1,4 +1,4 @@
 @import "./base/index.css";
 @import "./components/index.css";
 @import "./utils/index.css";
-@import "./vendor/wordpress.css";
+@import "./vendor/index.css";
```
mrgerardorodriguez: Good call.

mrgerardorodriguez:
Seems fine to me! :) 👍

PR #248, opened 2016-06-23
matching Our Work to other core pages

aileenjeffries:
re: https://trello.com/c/rFdXFU9e/201-chore-make-case-studies-core-header-more-consistent-in-terms-of-spacing-and-font-size-with-other-section-headers#

I was going to just tackle the header but it looks like this pattern was made before we switched a few things around in the templates. I used `talk-core.hbs` as an example. 
Old heading:
screen shot 2016-06-23 at 3 11 04 pm

New heading:
screen shot 2016-06-23 at 3 10 22 pm

The subheading is pretty long compared to other pages, but I think it looks fine.
cc @tylersticka @nicolemors

### DIFF
--- a/src/pages/pages/case-study-core.hbs
+++ b/src/pages/pages/case-study-core.hbs
@@ -1,79 +1,54 @@
 ---
 title: Case Study Core
-layout: toolkit.blank
-labels: latest
+layout: toolkit.default
+section:
+  heading: Our Work
+  subheading: We've built responsive web sites, JavaScript applications, hybrid apps, you name it—if it's mobile, we've built it.
+labels: 
+  - latest
 notes: |
   Work overview page.
-stylesheets:
-  - sandbox/sara-wip
 ---
 
-
-{{{embed "patterns.components.sky.base"}}}
-
-<main role="main" id="main">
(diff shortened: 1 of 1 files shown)


src/pages/pages/case-study-core.hbs
```diff
@@ -1,79 +1,54 @@
...
-    {{! Logos }}
-    <div class="Grid Grid--alignCenter Grid--alignMiddle u-padEnds3 u-md-padEnds6">
-      {{#each drizzle.data.case-studies.contents.logos}}
-        <div class="Grid-cell u-size1of2 u-sm-size1of3 u-md-size1of4 u-textCenter">
-          <img src="{{this}}" class="u-pad03">
-        </div>
-      {{/each}}
+{{! The Work }}
+{{#each drizzle.data.case-studies.contents.teasers}}
+  <div class="Grid Grid--alignMiddle Grid--withGutter u-padEnds6 Work">
```
tylersticka: @aileenjeffries This isn't your bug, but could you remove the `Work` class from this? It doesn't appear to be referenced anywhere.

tylersticka:
@aileenjeffries One comment inline, otherwise LGTM 👍

PR #250, opened 2016-06-24
Logos at bottom of Our Work are now svg

aileenjeffries:
re: https://trello.com/c/MXPBRAZ6/198-case-study-logos-should-totes-be-svg

Now, thanks to the magic of Googling and searching through Basecamp, they are!
screen shot 2016-06-24 at 9 51 09 am

@tylersticka @nicolemors @mrgerardorodriguez


src/data/case-studies.yml
```diff
@@ -1,11 +1,11 @@
...
-  - /images/sandbox/logo_entertainment.png
-  - /images/sandbox/logo_hautelook.png
-  - /images/sandbox/logo_ceasefire.png
+  - /images/case-studies/logos/logo_walmart.svg
+  - /images/case-studies/logos/logo_deschutes.svg
+  - /images/case-studies/logos/logo_obama-08.svg
+  - /images/case-studies/logos/logo_treasure-coast-hospice.svg
+  - /images/case-studies/logos/logo_entertainment.svg
+  - /images/case-studies/logos/logo_hautelook.svg
+  - /images/case-studies/logos/logo_ceasefire.svg
```
tylersticka: Nit: `logo_` prefixes are redundant if these files are in a `logos` directory.

src/static/images/case-studies/logos/logo_walmart.svg
```diff
@@ -0,0 +1,50 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 215 67">
+<g fill="#000000">
```
tylersticka: `fill` defaults to black, so this isn't necessary.

tylersticka:
A couple of inline comments. Some other, broader comments for the assets themselves:

- There are a lot of groups (`<g>`) that don't appear to be serving a purpose. We should remove them.
- At least one of the assets has many occurrences of `<path>` with `fill` attribute set to `none`. That probably means the `path` is invisible, perhaps some garbage data exported from Illustrator. This should be confirmed and the offending paths should be removed.

aileenjeffries:
@tylersticka I renamed the files to remove the `logo_` and cleaned up the svg code, getting rid of all groups and removing those empty filled shapes.

tylersticka:
@aileenjeffries Yay! LGTM 👍

PR #254, opened 2016-06-27
removing a Separator pattern we are not using

aileenjeffries:
re: https://trello.com/c/4PUKM2g3/135-chore-consistent-usage-of-hr-with-or-without-separator

I ran a search across `src` and determined that we are not using the `.Separator` class/pattern anywhere, although we do have plenty of `<hr>`s. I removed the css file and the pattern. 

cc @saralohr @mrgerardorodriguez @nicolemors @tylersticka


tylersticka:
LGTM :+1:

PR #255, opened 2016-06-28
What we do prototype

nicolemors:
Added new What We Do prototype for sharing, and also moved my icons to the correct location.

@tylersticka @aileenjeffries @saralohr @mrgerardorodriguez


tylersticka:
LGTM :+1:

PR #260, opened 2016-06-29
tweaking some icons for better vertical alignment

aileenjeffries:
re: https://trello.com/c/gQ0Lh4Ou/192-chore-icon-alignment-is-just-a-tad-off
since not all icons were off, I tweaked the svg's viewport for a few of them rather than changing the icon class. 

before:
icons-before

after:
icons-after

cc @tylersticka @nicolemors @saralohr @mrgerardorodriguez


tylersticka:
@aileenjeffries The envelope icon looks way better!

The social icons seem more awkward, specifically in the footer. Before, the icons had clean edges and aligned nicely in a row. Now, some of the icons have fuzzy edges and the non-square ones don't really line up cleanly. Details like the size of the Instagram icon (which was intentionally a tad larger to make up for the fact that it's no longer filled) have also been lost.

Before:

before

After:

after

Can we keep the new envelope but revert to the previous versions of the other icons?

aileenjeffries:
OK, Instagram,FB and LinkedIn are all back to where they were before.

tylersticka:
Thanks, @aileenjeffries. Sorry to be a buzzkill. LGTM 👍

PR #263, opened 2016-06-29
SVG icons: sprite → inline

tylersticka:
Well, _that_ was a pain.

This PR refactors our icons to be inline instead of using a sprite. This [follows GitHub's example](https://github.com/blog/2112-delivering-octicons-with-svg) and overcomes issues related to external `<use>` references and CDNs (which we'll likely encounter on WPEngine).

I originally wanted to make these updates to the base icon pattern so I wouldn't have to update every instance. Unfortunately, there were just too many differences between the idiosyncratic attribute names and structure of the pattern to easily "pass through" the old syntax to the new one, so I had to go replace them all by hand. There were also several instances of icons being reference directly without the helper, so those had to be tracked down as well.

I decided it wasn't worth the time right now to update old sandbox prototypes, except those that were dependent on patterns that are removed as a result of this PR. So some older sandbox prototypes may have broken icons for a while.

I'd love if someone could pull and try this locally before we merge because it touched so many individual files.

---

@saralohr @nicolemors @mrgerardorodriguez


tylersticka:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;n206gSOy&#x2F;143-chore-inline-svg-icons-instead-of-using-a-sprite"> Chore: Inline SVG icons instead of using a sprite</a>

saralohr:
What a chore. Checking out now @tylersticka!

saralohr:
I don't see anything amiss @tylersticka. Thanks for doing this. 👍

PR #274, opened 2016-07-01
Remove default hyphenation

### LINKED ISSUE #269: hyphens: auto in Firefox
I noticed that hyphens is set to auto in our body definition. Firefox take this as all the excuse it needs to put hyphens everywhere. I think this makes things harder to read and @megnotarte agrees with me. Can we just take this out or is there a reason it is there? 

screen shot 2016-06-30 at 12 48 37 pm

tylersticka:
Without smarter hyphenation in browsers and more predictable justified text, this just isn't as readable. I like that it maintains more consistently boxy shapes of text, but that's not worth the readability tradeoffs. It is text, after all!

This only affects Firefox and Safari. Screenshots below are from Firefox, which has the unique distinction of _not_ breaking "date-sensitive" at the hyphen for some reason. _Oh, Firefox, you card..._

## Before

screen shot 2016-06-30 at 5 55 41 pm

## After

screen shot 2016-06-30 at 5 56 02 pm

---

@saralohr @mrgerardorodriguez @aileenjeffries @megnotarte 

Fixes #269

### DIFF
--- a/src/assets/toolkit/styles/base/typography.css
+++ b/src/assets/toolkit/styles/base/typography.css
@@ -13,15 +13,13 @@
  *
  * 1. Base font style for everything
  * 2. Allow long words (links) to brake at arbitrary points
- * 3. Auto-hyphens by default (assuming removal for specific elements)
- * 4. Prevent automatic size adjustments on orientation change
+ * 3. Prevent automatic size adjustments on orientation change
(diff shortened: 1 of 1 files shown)


mrgerardorodriguez:
LGTM 👍🏽

PR #275, opened 2016-07-01
Don't wrap code blocks

### LINKED ISSUE #272: Consider not wrapping code samples
We have a lot of _verbose_ code samples in the blog, and I think the wrapping is ugly. [Example](http://c4site.staging.wpengine.com/wordpress-taking-the-hack-out-of-multiple-custom-loops/):

screen shot 2016-06-30 at 3 23 00 pm

We should consider making this a good ol' `pre` rather than `pre-wrap`, and maybe doing something with `overflow-x` on the container.

tylersticka:
In practice, this didn't work so great with our content.

Now, code blocks with very long lines will scroll horizontally.

---

@saralohr @mrgerardorodriguez 

Fixes #272

### DIFF
--- a/src/assets/toolkit/styles/base/base.css
+++ b/src/assets/toolkit/styles/base/base.css
@@ -66,11 +66,6 @@ pre {
   color: var(--color-fuchsia);
 }
 
-/**
- * TODO: Revise once we know how this will or won't interact with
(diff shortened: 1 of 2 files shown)


mrgerardorodriguez:
LGTM 👍🏽

Issue #270, opened 2016-06-30
Blog post: Image is overlapping author bio section on post detail


megnotarte:
@tylersticka says we need to add the `.u-cf` class to the container so this doesn't happen: 
![screen shot 2016-06-30 at 2 53 48 pm](https://cloud.githubusercontent.com/assets/66242/16505553/77bdd25e-3ed2-11e6-8889-9e3146b2249f.png)

tylersticka:
Fixed in the pattern library, still needs to be addressed in the theme (working on that now).

PR #276, opened 2016-07-01
Clear lingering floats in single post content

tylersticka:
This big scary diff is just just me wrapping the post content in a `<div>`, moving `u-spaceItems1` there and adding `u-cf` so floats will be contained.

Look at how well-contained this float is, for goodness' sake:

screen shot 2016-06-30 at 6 17 31 pm

Template will need updating in WP as well.

---

@saralohr @mrgerardorodriguez 

See: #270


saralohr:
👍

PR #281, opened 2016-07-05
Add workaround for `FlexEmbed-content` when using `wp_oembed_get()`

mrgerardorodriguez:
## Overview

There exists content on the WP site that takes advantage of the [oEmbed](http://oembed.com/) format. WordPress offers a handy [`wp_oembed_get()`](https://codex.wordpress.org/Function_Reference/wp_oembed_get) function that allows us to pass a URL to the content returning the proper embed code from the specific provider.

## The Challenge

The challenge lies in making sure the oEmbed content that is embedded uses the `FlexEmbed` classes properly. The [SUIT CSS `FlexEmbed`](https://github.com/suitcss/components-flex-embed) component expects a `FlexEmbed-content` CSS class to be added onto the `<iframe>` of the embedded content to work properly. This cannot be done using the `wp_oembed_get()` function. We are only allowed to set the `height` and/or `width` attributes for the embedded `<iframe>`.

## The Proposed Solution

The proposed solution is this PR. I will be the first to admit that his is a hack and is not DRY. If this solution feels _too_ hacky, I am open to ideas/suggestions/conversations on how to overcome his challenge in a much more elegant fashion. 😉 

Some thoughts...
- Should this be part of the patterns repo? Feels forced/unnecessary.
- Should the `>` combinator be used?

---

cc: @saralohr @tylersticka

### DIFF
--- a/src/assets/toolkit/styles/components/flex-embed.css
+++ b/src/assets/toolkit/styles/components/flex-embed.css
@@ -9,6 +9,7 @@
  *
  * @see https://codex.wordpress.org/Function_Reference/wp_oembed_get
  * @see https://github.com/suitcss/components-flex-embed/blob/master/lib/flex-embed.css#L66-L73
+ * @see https://github.com/cloudfour/cloudfour.com-wp/pull/91#discussion-diff-69822034
  */
 .FlexEmbed > iframe {
   bottom: 0;
--- a/src/assets/toolkit/styles/components/flex-embed.css
+++ b/src/assets/toolkit/styles/components/flex-embed.css
@@ -1,1 +1,20 @@
 @import "suitcss-components-flex-embed";
+
+/**
+ * The wp_oembed_get() does not allow us to add the `FlexEmbed-content` class
+ * to the embedded iframe. This is a workaround (hack) to get the flex embed
+ * to work in this situation. This means the `FlexEmbed-content` CSS rules
+ * are duplicated, unfortunately, but I failed to find a different solution.
+ *
+ *
+ * @see https://codex.wordpress.org/Function_Reference/wp_oembed_get
+ * @see https://github.com/suitcss/components-flex-embed/blob/master/lib/flex-embed.css#L66-L73
+ */
+.FlexEmbed > iframe {
+  bottom: 0;
+  height: 100%;
+  left: 0;
+  position: absolute;
+  top: 0;
+  width: 100%;
+}


tylersticka:
@mrgerardorodriguez And the following doesn't work?

```html
<div class="FlexEmbed">
  <div class="FlexEmbed-ratio FlexEmbed-ratio--16by9"></div>
  <div class="FlexEmbed-content">
    <!-- iframes or whatever -->
  </div>
</div>
```

mrgerardorodriguez:
Great question @tylersticka! I tried this and it works _better_. The `FlexEmbed-content` container is correctly positioned, but the `<iframe>` is not the correct size within that container. It appears I cannot pass along `100%` as the `width` or `height`, instead only pixels. Boo.

screen shot 2016-07-05 at 5 10 32 pm

I just found this and am wondering if I can use some sort of WP filter to modify the HTML???
- http://wordpress.stackexchange.com/questions/5362/wp-oembed-not-passing-through-the-autoplay-1-variable
- https://developer.wordpress.org/reference/hooks/oembed_result/

I'll take a closer look.

tylersticka:
Lazier idea: Setting `width` and `height` to something larger than we need it to be, like `1024` and `768`. Pretty sure the `<iframe>` will just max out the `.FlexEmbed-content` element.

But the HTML idea is probably cleaner. 😆

mrgerardorodriguez:
> Lazier idea: Setting `width` and `height` to something larger than we need it to be, like `1024` and `768`. Pretty sure the `<iframe>` will just max out the `.FlexEmbed-content` element.

Thanks for the suggestion @tylersticka! I was hoping the same, but the `<iframe>` doesn't resize down, it seems. 😞 You'll notice the play button is low, to the right, because the `<iframe>` is much larger than the `.FlexEmbed-content`. 

screen shot 2016-07-06 at 9 31 14 am

tylersticka:
@mrgerardorodriguez Ugh, that is _so_ lame.

mrgerardorodriguez:
Going to close this PR as this is not the best solution for this challenge.

mrgerardorodriguez:
Welp, guess we'll open this back up again. See https://github.com/cloudfour/cloudfour.com-wp/pull/91#discussion_r69824599

mrgerardorodriguez:
Can I get a re-review, please, @tylersticka @saralohr and also @erikjung?

saralohr:
This seems okay to me @mrgerardorodriguez.

> Should the` >` combinator be used?

I'd say yes to prevent styles from being added to an `iframe` child of `FlexEmbed-content` in other usages.

erikjung:
@mrgerardorodriguez This is bazaar. I think your solution is fine, but looking at the SUIT component test page, it looks like we shouldn't need to add the class directly to the `<iframe>`: http://suitcss.github.io/components-flex-embed/test/

Are there other CSS rules affecting the `<iframe>`?

erikjung:
@mrgerardorodriguez Also, can we try a simpler solution first:

```css
iframe {
  width: 100%;
  height: 100%;
}

mrgerardorodriguez:
> This is bazaar.

Yeah, it doesn't feel right.

> Are there other CSS rules affecting the <iframe>?

So, this is interesting. To answer your question, no. _But,_ I noticed this on that SUIT component test page. There is a `width: 100%` and `height: 100%` CSS rule being applied to `.Test-run iframe`:

screen shot 2016-07-06 at 5 03 16 pm

This is the difference maker and _totally_ makes sense. Seems like we should, instead of this hacky solution, add those same styles to our `<iframe>`.

mrgerardorodriguez:
>  Also, can we try a simpler solution first

LOL, yes. 😆

erikjung:
@mrgerardorodriguez Dang, my misspelling will live forever now. :)

erikjung:
@mrgerardorodriguez If the simple `iframe` selector version actually works, then I think that's a fine base style to apply to all iframes, assuming we'll be putting them within the `.FlexEmbed` structure every time.

mrgerardorodriguez:
Closing this PR again. Found a more elegant solution (see above).

mrgerardorodriguez:
@erikjung @tylersticka @saralohr See https://github.com/cloudfour/cloudfour.com-patterns/pull/285 for new PR

PR #287, opened 2016-07-07
Add utilities for flex shrink and width full

saralohr:
Add a couple utilities for a bug fix for IE 11. 

Will fix Issue #11 once theme template is updated.

### DIFF
--- a/src/assets/toolkit/styles/utils/size.css
+++ b/src/assets/toolkit/styles/utils/size.css
@@ -17,3 +17,7 @@
     max-height: 50vh !important;
   }
 }
+
+.u-widthFull {
+  width: 100% !important;
+}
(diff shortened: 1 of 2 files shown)


src/assets/toolkit/styles/utils/size.css
```diff
@@ -17,3 +17,7 @@
     max-height: 50vh !important;
   }
 }
+
+.u-widthFull {
```
tylersticka: Any way we can keep this a `u-size` syntax? `u-size1of1` or something?

saralohr: Ha, yeah. I just (finally) submitted an issue to SUITCSS for adding a utility for just `width: 100%` and that was the class name I suggested. I _didn't_ do it here because I thought it might be misleading if it didn't have all the responsive options like the other size utilities, but I guess that's silly. I'll change it up.

saralohr:
Is this good for a merge?

tylersticka:
@saralohr Indubitably!

Issue #283, opened 2016-07-06
IE 11: Case study images are smaller than expected


megnotarte:
http://c4site.staging.wpengine.com/made/

![bs_win10_ie_11 0](https://cloud.githubusercontent.com/assets/66242/16637005/a24925a6-4390-11e6-9450-d72a305838e8.jpg)

tylersticka:
Adorbs 👶

megnotarte:
I think this is a problem with the `WorkImage` class. I see it here as well: 
![bs_win10_ie_11 0 2](https://cloud.githubusercontent.com/assets/66242/16637129/b17e7926-4391-11e6-8c76-fa2a094d987d.jpg)

tylersticka:
It's just mobile-optimized. :trollface:

nicolemors:
The small screen is in the wrong location and neither is tilted as well.

PR #289, opened 2016-07-08
Chore: Add some a couple utilities in support of WorkImage

saralohr:
This PR adds a couple new utilities in support of fixing an issue with `WorkImage` in IE (Issue #283). Theme PR coming up next.

Also removes some classes that are no longer being used in `WorkImage`.

Regarding the removal of `WorkImage-small` properties, the fix for positioning this element in IE was to add the transforms directly to the SVG as attributes. 

### DIFF
--- a/src/assets/toolkit/styles/components/work-image.css
+++ b/src/assets/toolkit/styles/components/work-image.css
@@ -5,10 +5,15 @@
   --WorkImage-detail-alt-stroke: var(--WorkImage-detail-fill);
   --WorkImage-detail-alt-stroke-width: 2px;
   --WorkImage-screen-fill: var(--color-white);
+  --WorkImage-tilt-rotate: -3deg;
 }
 
 .WorkImage {
 
+}
+
+.WorkImage--tilt {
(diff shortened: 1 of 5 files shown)


tylersticka:
> Regarding the removal of WorkImage-small properties, the fix for positioning this element in IE was to add the transforms directly to the SVG as attributes.

So should this PR include those SVG updates? I'm confused since I don't see changes to anything but CSS..?

saralohr:
Oh...I forgot the SVG was included in the patterns library. Doh. I just updated it in the theme. I'll make that change here too.

saralohr:
@tylersticka Updated the `WorkImage` component, and in doing so realized that since the `WorkImage` base class doesn't actually have any properties associated with it, it could move to the new `FlexEmbed` parent element and retain `WorkImage--tilt` as a modifier rather than moving this bespoke style to a utility.

screen shot 2016-07-08 at 10 41 42 am

screen shot 2016-07-08 at 10 41 26 am

tylersticka:
LGTM 👍

PR #299, opened 2016-07-11
very minor tweak to 404 robot

aileenjeffries:
This is such a minor detail but I noticed it when I was working on the WP 404 page. Since the robot's body has depth, I think it looks better when his arms are coming out of his sides than his front. 

screen shot 2016-07-11 at 12 07 44 pm

cc @nicolemors @tylersticka @saralohr @mrgerardorodriguez


tylersticka:
Agreed! 👍

PR #303, opened 2016-07-11
icons for Flickr, Lanyrd and Speakerdeck

aileenjeffries:
re: https://trello.com/c/V9lx7Eis/297-add-missing-social-networking-icons

adds icons for the lesser social media platforms. 

screen shot 2016-07-11 at 2 35 46 pm

cc @tylersticka @saralohr @nicolemors @mrgerardorodriguez


tylersticka:
The Flickr one looks a little beefy. I'm wondering if we can avoid using the entire canvas, similar to LinkedIn or Facebook's icons?

I also wonder if the SpeakerDeck icon would benefit from a containing shape? https://dribbble.com/shots/240722-Social-Icons-v1-8

aileenjeffries:
@tylersticka for the flickr one, I started with thne FB box but forgot to edit the viewbox. It should match now. For the speakerdeck, I started off with a more complex shape, it just didn't read too well at a small size. Something like the attached. 

Maybe it would be better to have a solid rectangular box instead of a stroked box? 

screen shot 2016-07-11 at 2 44 51 pm

tylersticka:
@aileenjeffries Maybe! It's also not a very good icon to begin with, so if you try and it and are still like "bleh," we can either keep it as-is or say "we like our generic link icon better." ¯\\\_(ツ)_/¯

aileenjeffries:
screen shot 2016-07-11 at 2 56 25 pm
updated flickr icon, including fb icon for comparison

aileenjeffries:
I agree the speakerdeck icon is weak, but I don't want it to be the only one without an icon (for now, I'm sure we'll eventually add something else iconless), and I think it looks more balanced with the other icons like this: 

screen shot 2016-07-11 at 3 09 49 pm

tylersticka:
I like that better, too!

src/static/images/icons/speakerdeck.svg
```diff
@@ -0,0 +1,3 @@
+<svg  xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
```
tylersticka: Nit: Extra space

src/data/icons.yml
```diff
@@ -7,13 +7,16 @@ mono:
   "cloud-four": "icons/cloud-four"
   "envelope": "icons/envelope"
   "facebook": "icons/facebook"
+  "fickr": "icons/flickr"
```
tylersticka: `fickr` ? 😆

aileenjeffries:
good to meeerge, @tylersticka ?

tylersticka:
Yes! 👍

PR #307, opened 2016-07-11
adjusting speakerdeck icon viewbox

aileenjeffries:
Even though I have a 24x24 viewbox on the speakerdeck icon, it wasn't vertically aligning nicely with text. By tweaking the viewbox, it now lines up more nicely. 
screen shot 2016-07-11 at 4 07 36 pm

cc @tylersticka specifically because you'll be able to tell me if this is the wrong way to fix this.


tylersticka:
Looks a tad high to me, but no biggy.

This is a fine way of fixing this. 👍

PR #308, opened 2016-07-11
Device Lab Form - Class name fix

nicolemors:
Mistakenly used wrong JetPack class name. Fixed it, form should look legit now.

@tylersticka @saralohr @aileenjeffries @mrgerardorodriguez


tylersticka:
LGTM :+1:

PR #312, opened 2016-07-12
Add dimensions to SVG

saralohr:
This PR addresses issue [#119](https://github.com/cloudfour/cloudfour.com-wp/issues/119), adding width and height attributes to SVG icons so they are not huge if the CSS fails to load.

Tested this in a few browsers and did not see any issues, but this should probably be tested more thoroughly to ensure no regression.

screen shot 2016-07-12 at 8 37 12 am

cc @tylersticka @mrgerardorodriguez @aileenjeffries


tylersticka:
> Tested this in a few browsers and did not see any issues, but this should probably be tested more thoroughly to ensure no regression.

Sorry if this is a dumb question, but you also tested _with_ CSS enabled to make sure those attributes were being overridden by our CSS, right?

If yes, then 👍

saralohr:
> Sorry if this is a dumb question, but you also tested with CSS enabled to make sure those attributes were being overridden by our CSS, right?


Yep! The CSS overrides the SVG attributes.

Issue #310, opened 2016-07-12
Toolkit JS includes unused FactSlider component


tylersticka:
We should probably delay including `FactSlider` in our toolkit JS until we're ready to launch that feature?

(Assigning this to "Launch" milestone so we can discuss, not as a statement of priority.)

saralohr:
@tylersticka Sure. It will probably need modified anyways to support the slide animation. I can remove it for now and we can refer back to earlier commits when we revive it.

PR #313, opened 2016-07-12
Remove fact slider

saralohr:
Addresses Issue #310 to remove fact slider JS until we implement this component.


tylersticka:
LGTM :+1:

PR #318, opened 2016-07-13
Minimizing wrapping in the navigation

aileenjeffries:
I made some small adjustments to the navigation to reduce awkward wrapping. What I did:
- removed the left margin on the ul with the class name of `Sky-nav-menu-items`.
- reduced the breakpoint for the `--Sky-nav-full-viewport` variable from 65em to 45em (the point where all nav items fit in one line)
- At this same breakpoint, I reduced the left and right padding on the links from 1em to 0.5em. This also helps to get everything on one line in smaller screens. Because the list items space out at this breakpoint, the screen size range where the links are actually only 0.5em + 0.5em away from each other is very brief and they quickly expand outward. 
- In the lg breakpoint, I switched the font size from md to sm, because the font size increase was wrapping links again. I increased the font size back to md at the xl breakpoint because there's enough room. 

nav
 
cc @nicolemors @saralohr @grigs @erikjung @mrgerardorodriguez but who I am kidding, it's gonna be @tylersticka reviewing this. :)

### DIFF
--- a/src/assets/toolkit/styles/components/sky.css
+++ b/src/assets/toolkit/styles/components/sky.css
@@ -40,6 +40,8 @@
 @custom-media --Sky-nav-expanded-viewport (width >= 36em);
 @custom-media --Sky-nav-full-viewport (width >= 56em);
 @custom-media --Sky-scene-full-viewport (--md-viewport);
+@custom-media --Sky-nav-logo-sm-min (width >= 64em);
+@custom-media --Sky-nav-logo-sm-max (width < 68em);
 
 /**
  * 1. Keep absolute-positioned elements from leaving this container.
@@ -110,13 +112,14 @@
  * 1. Align items vertically.
  * 2. We let items wrap initially because we want the menu to be above the
  *    other elements on narrow screens.
+ * 3. So the navigation can align right when desirable.
  */
 
 .Sky-nav {
   display: flex;
   align-items: center; /* 1 */
-  flex-wrap: wrap;
-  justify-content: space-between;
+  flex-wrap: wrap; /* 2 */
+  justify-content: space-between; /* 3 */
 }
 
 /**
@@ -215,19 +218,19 @@
   fill: var(--color-relative); /* 4 */
 }
 
-@media (width >= 64em) {
+@media (--Sky-nav-logo-sm-min) and (--Sky-nav-logo-sm-max) {
+/**
+ * In this viewport range, reducing the size of the logo slightly prevents orphaned
+ * nav items.
+ */
+
   .Sky-nav-logo-object {
     width: var(--Sky-nav-logo-size-sm);
     width: var(--Sky-nav-logo-size-sm);
   }
 }
 
-@media (width >= 68em) {
-  .Sky-nav-logo-object {
-    width: var(--Sky-nav-logo-size);
-    width: var(--Sky-nav-logo-size);
-  }
-}
+
 
 
 /**
@@ -262,7 +265,7 @@
  * 2. Remove the bump (since it won't be above anything at this point).
  * 3. Reset the background.
  * 4. To align right, this now needs to be last.
- * 5. Groups the last three links on a second line to avoid orphans.
+ * 5. Constraining the width keeps the last three links on a second line to avoid orphans.
  */
 
 @media (--Sky-nav-expanded-viewport) {
@@ -276,6 +279,10 @@
 }
 
 @media (--Sky-nav-full-viewport) {
+/**
+ * No need for a max width now that there's enough room for all nav items to fit on a line.
+ */ 
+
   .Sky-nav-menu {
     max-width: 100%;
   }
@@ -302,16 +309,19 @@
   .Sky-nav-menu-items {
     display: flex; /* 1 */
     margin: var(--Sky-nav-menu-space); /* 2 */
-    margin-left: 0;
+    margin-left: 0; /* 3 */
     flex-wrap: wrap; /* 4 */
     justify-content: flex-end; /* 5 */
   }
 }
 
 @media (--xl-viewport) {
+  /**
+   * Adds balance between the logo and nav when there's enough room.
+   */
+
   .Sky-nav-menu-items {
     margin: var(--Sky-nav-menu-space);
-    max-width: 100%;
   }
 }
 
--- a/src/assets/toolkit/styles/components/sky.css
+++ b/src/assets/toolkit/styles/components/sky.css
@@ -43,8 +43,8 @@
 @custom-media --Sky-nav-expanded-viewport (width >= 36em);
 @custom-media --Sky-nav-full-viewport (width >= 48em);
 @custom-media --Sky-scene-full-viewport (--md-viewport);
-@custom-media --Sky-nav-font-sm (width >= 64em);
-@custom-media --Sky-nav-font-md (width < 80em);
+@custom-media --Sky-nav-font-size-sm-min (width >= 64em);
+@custom-media --Sky-nav-font-size-sm-max (width < 80em);
 
 /**
  * 1. Keep absolute-positioned elements from leaving this container.
@@ -353,7 +353,7 @@
   }
 }
 
-@media (--Sky-nav-font-sm)  and (--Sky-nav-font-md) {
+@media (--Sky-nav-font-size-sm-min)  and (--Sky-nav-font-size-sm-max) {
   /**
    * In this viewport range, the body font size increases enough to wrap the last
    * couple of nav items. This prevents the wrapping.
--- a/src/assets/toolkit/styles/components/sky.css
+++ b/src/assets/toolkit/styles/components/sky.css
@@ -25,7 +25,8 @@
   --Sky-nav-menu-item-line-height: var(--line-height-xs);
   --Sky-nav-menu-item-transition-duration: var(--motion-duration-md);
   --Sky-nav-menu-item-transition-timing-function: var(--motion-timing-function-default);
-  --Sky-nav-menu-space: var(--space-xs);
+  --Sky-nav-menu-space: var(--space-sm) var(--space-xs);
+  --Sky-nav-menu-max: 30em;
 
   --Sky-scene-object-x-offset: calc(2 / 3 * 100%);
   --Sky-scene-object-x-offset-full: 75%;
@@ -204,7 +205,6 @@
   }
 }
 
-
 /**
  * 1. Most embeds are `inline-block` by default, which results in an unsightly
  *    bottom space. Yuck.
@@ -278,7 +278,7 @@
     padding-bottom: 0; /* 2 */
     background: transparent; /* 3 */
     order: 2; /* 4 */
-    max-width: 30em; /* 5 */
+    max-width: var(--Sky-nav-menu-max); /* 5 */
   }
 }
 
@@ -288,7 +288,7 @@
  */ 
 
   .Sky-nav-menu {
-    max-width: 100%;
+    max-width: initial;
   }
 }
 
--- a/src/assets/toolkit/styles/components/sky.css
+++ b/src/assets/toolkit/styles/components/sky.css
@@ -13,8 +13,8 @@
   --Sky-nav-controls-space-x: var(--space-md);
   --Sky-nav-controls-space-y: var(--space-xs);
   --Sky-nav-logo-color: var(--Sky-color);
-  --Sky-nav-logo-height: 5em;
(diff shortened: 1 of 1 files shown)


tylersticka:
Gonna critique design stuff before spelunking through code...

At the widest sizes, the lack of margin on the left is making the space between the logo and the first nav item feel noticeably smaller than in the current iteration:

screen shot 2016-07-13 at 4 20 43 pm

The reduced padding means that items start very close to one another. `768px` is a very common breakpoint, and I personally find these the links with spaces in them (the first three) difficult to differentiate when they're this close together:

screen shot 2016-07-13 at 4 21 45 pm

My eyes actually get a little confused until they hit "Articles," where every space and capital letter represents a new nav item.

tylersticka:
(An argument could be made to shorten one of the titles, like "Our Work" to just "Work." I dunno if I have the energy left for a naming debate, though.)

aileenjeffries:
The margin is the easier one to fix - I added it back in at larger screen sizes. 

For the links appearing too close together, what about dropping down to font size sm as soon as they're all on one line and increasing it to back md when there's enough room? This is what that decreased looks like at the tightest point. 

screen shot 2016-07-13 at 4 38 18 pm

aileenjeffries:
nav-font-size-tweak

tylersticka:
I just pulled to try it myself, and I feel like it gets _too_ small, particularly on non-retina screens. This is especially apparent on section pages, where it feels a bit dwarfed by the rest of the header content:

screen shot 2016-07-13 at 4 42 57 pm

One potential alternative to adjusting the `font-size` would be to adjust the left margin or width of the element (or whatever property is controlling the width) to reduce the width of the nav so the last item doesn't wrap. As-is, I prefer the current site's medium breakpoint where two items wrap (since they're all easier to read).

aileenjeffries:
@tylersticka what do you think of the latest changes, the addition of the left margin on the ul at larger screen sizes, and the smaller font size when there's enough room for all links to be on one line so there's more space between each link?

tylersticka:
As per [my previous comment](https://github.com/cloudfour/cloudfour.com-patterns/pull/318#issuecomment-232518762), I really dislike the reduced nav size. When I look at the site on a tablet in portrait, I unquestionably prefer these wrapping but large, easy-to-tap links (current `master`):

screen shot 2016-07-15 at 9 49 23 am

To these smaller links (this branch):

screen shot 2016-07-15 at 9 49 31 am

I understand the appeal of keeping things on one line _visually_, I just don't think it's worth the trade-off in hierarchy and target size from a UX standpoint.

aileenjeffries:
Somehow, I didn't see your comment from 2 days ago until now. :(

When you say, "One potential alternative to adjusting the font-size would be to adjust the left margin or width of the element (or whatever property is controlling the width) to reduce the width of the nav so the last item doesn't wrap. " do you mean constraining the overall nav width at certain screen sizes so that more than one nav item is on the second line? 

I understand your concerns about the font size of the nav relative to other text on the page. If how it is now is the best possible alternative, I can live with it. But, only having one or two nav items on the second line really looks unintentional to me, especially since they get on one line and then get moved back down to the second line as the font size increases. And yes, I realize normal users don't spend their time resizing their browsers.

tylersticka:
> do you mean constraining the overall nav width at certain screen sizes so that more than one nav item is on the second line?

Yep, exactly. In a perfect world, we'd somehow force this sort of wrapping when the nav needs to be on two lines:

screen shot 2016-07-15 at 10 18 40 am

I tried to do this months ago, but ran into a lot of issues. Percentage widths were finicky, splitting the nav into two elements increased maintenance and made the single-line sizes a real pain. My _hope_ is that I was overlooking a sane solution.

> But, only having one or two nav items on the second line really looks unintentional to me, especially since they get on one line and then get moved back down to the second line as the font size increases.

I agree that one or two nav items look unintentional, but I disagree about the "especially" part. I understand what you're saying, but because the change happens along with the relative size of all other page elements, I don't find it especially jarring.

aileenjeffries:
OK, I'll try to come up with something that:
-doesn't have the font size any smaller than `font-size-md` at any point.
-doesn't have less than 2em of space between links (1em padding left and right).

aileenjeffries:
OK, after much banging my head against the wall, I figure out why I couldn't limit the width of the nav and align it right - the `Sky-nav-controls` were messing with my order. Once I fixed that, I got it working as I envisioned. 

1. This is just a comparison between my branch and production. As you can see, they look just the same!
narrow

2. The switch to the expanded (in page) menu also looks pretty darn close.
narrowest-expanded

3. As the screen size increases, the last 3 nav items stay grouped together on the second line, thanks to that max-width on the nav.
3-stay

4. This is the narrowest point at which all items are on one line. As you can see, the font size is the same and there's still a reasonable amount of room between each item.
narrowest-one-line

5. The links spread out as soon as they're on one line, which I think looks better than right aligned. This is something that can be easily changed if you disagree. 
earlier-spreading-out

6. Currently, the font size increase at some break point (md? lg?) bumps up the font size enough to orphan the last nav item. This makes me sad. By decreasing the size of the logo ever-so-slightly and decreasing margins, I'm able to avoid this. Font size wasn't tampered with. 
no-more-sad-orphan

7. At the largest screen size, they match too. 
widest

Animation of my branch. 
animated

I also pushed my changes so you can try it yourself, although the CSS isn't ready for review.

tylersticka:
That's looking very promising! I'd try this out locally if my branch weren't a _mess_ right now to commit, but based on your GIFs and screenshots I'd say you're on a positive track. 👍

mrgerardorodriguez:
@aileenjeffries Would you like me to checkout your branch to try it locally? :)

aileenjeffries:
@mrgerardorodriguez please do! Especially now that I've cleaned up my sloppy CSS. :)

mrgerardorodriguez:
Works great locally @aileenjeffries! I looked at `http://localhost:3000/pages/global-nav.html`. Was that what I needed to look at?

aileenjeffries:
@mrgerardorodriguez, any full-screen page will do, including global-nav.html!

mrgerardorodriguez:
Cool, it was looking great locally then, @aileenjeffries!

tylersticka:
Apologies for the delay, @aileenjeffries, I finally reached a stopping point where I could pull and take a look.

The wider layouts look really, really nice across all the modern browsers (even Edge!).

On my phone, the nav items are harder to tap because they're now a bit shorter. The animation is also choppier, though I'm not sure why the performance has suffered... I'll take a look at the code and see if anything jumps out.

src/assets/toolkit/styles/components/sky.css
```diff
@@ -238,15 +268,27 @@
...
+ * 5. Constraining the width keeps the last three links on a second line to avoid orphans.
  */
 
 @media (--Sky-nav-expanded-viewport) {
   .Sky-nav-menu {
     display: block; /* 1 */
     padding-bottom: 0; /* 2 */
     background: transparent; /* 3 */
-    order: 0; /* 4 */
+    order: 2; /* 4 */
```
tylersticka: I'm confused why this change is necessary. Did you run into a conflict with the "skip to main content" controls?

src/assets/toolkit/styles/components/sky.css
```diff
@@ -238,15 +268,27 @@
...
+  }
+}
+
+@media (--Sky-nav-full-viewport) {
+/**
+ * No need for a max width now that there's enough room for all nav items to fit on a line.
+ */ 
+
+  .Sky-nav-menu {
+    max-width: 100%;
```
tylersticka: If there's no need for a `max-width`, better to set this to `initial` rather than an explicit value.

src/assets/toolkit/styles/components/sky.css
```diff
@@ -190,6 +194,17 @@
...
+/**
+ * Reducing the left padding on the logo balances it with the right margin on the nav.
+ */
+
+  .Sky-nav-logo {
+    padding-left: var(--space-xs);
+  }
+}
+
+
```
tylersticka: Nit: Extra whitespace

src/assets/toolkit/styles/components/sky.css
```diff
@@ -238,15 +268,27 @@
...
  */
 
 @media (--Sky-nav-expanded-viewport) {
   .Sky-nav-menu {
     display: block; /* 1 */
     padding-bottom: 0; /* 2 */
     background: transparent; /* 3 */
-    order: 0; /* 4 */
+    order: 2; /* 4 */
+    max-width: 30em; /* 5 */
```
tylersticka: This should probably be a variable?

tylersticka:
Some comments inline, but I'm more concerned about the choppier performance on small screens. I don't see anything obvious jumping out. I can take a deeper look next week (I wound up in JS-land way longer than intended today and my circuits are pretty fried from it).

src/assets/toolkit/styles/components/sky.css
aileenjeffries: This was the crazy frustrating part. If, in the expanded view I had anything but the flex order of logo, `Sky-nav-controls` then nav, I couldn't get the nav to align right. This is because I'm using `justify-content: space-between` on `Sky-nav` to get the logo on the left and the nav on the right. Even with the third element is hidden, it still counts, so the nav wasn't aligning right. I've added comments to explain this, but probably not well enough!

tylersticka: What I might recommend is resetting the order of this _and_ of the controls. At expanded sizes, the only visible control is absolute-positioned, so in theory that would get you were you need to be?

aileenjeffries:
The reduced height of the nav items on small screens was totally unintentional. I switched it back. I doubt this is the cause of the choppiness, but maybe? 

The only other thing I did that might have effected the way it works on small screens is adding `justify-content: space-between` to `.Sky-nav`. If this turns out to be the culprit, I could always wait until the `--Sky-nav-expanded-viewport` to add it.

tylersticka:
D'oh, I'm an idiot. I was comparing a really short page to an "article" page (which is comparatively long) locally. Longer pages will always be a little bit choppier because we're animating the `<body>` element.

After comparing "like" pages, the performance is the same as it was.

I'm actually really relieved it was my own idiocy... much less mysterious than the alternative! 😅

aileenjeffries:
Whew! I'll take idiocy over mystery choppiness any day.

src/assets/toolkit/styles/components/sky.css
aileenjeffries: I'm not sure I'm with you yet, @tylersticka  
on narrow screens, the order is:
-1: nav
0: logo (this is just by default, it isn't actually defined)
1: sky-nav-controls

On expanded screens, the order is:
0: logo
1: sky-nav-controls
2: nav

I'm not sure what resetting the order of sky-nav-controls on expanded screens would get me?

tylersticka: In the markup, the source order is:

1. Controls
2. Logo
3. Nav

So why do we change their `order` at all? Because at small screens, the `nav` must be _above_ the other two elements, and the "Menu" toggle needs to be to the _right_ of the logo.

But at large screens, neither is the case. The "menu" toggle isn't even visible, and the "nav" is fine exactly where it would be naturally.

With this in mind, you _should_ be able to set the nav's `order` to `1` (same as the controls), or you could even reset the `order` of controls _and_ nav (both `0`) because their natural order is just fine after that breakpoint.

aileenjeffries: screen shot 2016-07-18 at 12 02 00 pm

If I reset everything to `order: 0`, this is what happens (since ` sky-nav-controls` is first in the source order. Doing:
- logo: 0
- ` sky-nav-controls`: 1
- nav: 1
works fine, because it falls back to source order if two elements get the same `order`.

Buuuut, I think it is more clear when I set the order of nav to 2. I have to switch it to something greater than or equal to 1, since it starts off at -1. To me, switching it to 2 makes it more clear what order you can expect the elements to show up in?

tylersticka: > To me, switching it to 2 makes it more clear what order you can expect the elements to show up in?

I disagree, but on a scale of 1 to 10 I'm at a 1 in terms of caring. This is the most minor feedback I've given on the whole PR. As long as it's commented, we're fine.

aileenjeffries: Cool. It is definitely commented. Am I good to merge, @tylersticka ?

tylersticka:
I think you're good to merge! Thanks for enduring my feedback, I'm happy with the changes! 😄

PR #320, opened 2016-07-14
Graceful degradation for FloatLabel events

### LINKED ISSUE #311: FloatLabel component should support autofill
Creating an issue based on something @grigs [pointed out in Slack](https://cloudfourredesign.slack.com/archives/wordpress/p1468276106000064). The current `FloatLabel` pattern was not developed with autofill in mind, which can be finicky to detect.

The _good_ news is that we are defaulting to _assuming_ a `value`, so we're in a good spot to try whatever timeout- or event-based approaches are available. But something to have on our radar.

screen shot 2016-07-11 at 3 24 24 pm

tylersticka:
When we decided to move away from Modernizr for solutions that didn't block the page load, we lost their `oninput` event check. Our solution at the time was to overtly specify the `keyup` event for float labels, which seemed a fine solution.

But as described in #311, that event will only fire when keys are literally pushed, at least in some browsers. This meant that alternative means of filling out the forms, such as autofill or pasting text from a right-click, would not trigger the active state consistently.

This PR removes the `eventName` option from our `FloatLabel` component, instead using [a very simple technique](https://mathiasbynens.be/notes/oninput) for detecting `oninput` support and falling back to `keyup` otherwise.

To appropriately test this, I added [autofill attributes](https://cloudfour.com/thinks/autofill-what-web-devs-should-know-but-dont/) for the `name` and `email` fields of our comment form. But for some reason I'm having trouble triggering autofill in Firefox and Edge... I followed instructions for both browsers, but neither wants to cooperate. I'm hoping that @grigs can help me out since he wrote that post.

## Steps to test

1. Check out this branch locally.
2. Run `npm start`
3. Navigate to [http://localhost:3000/pages/blog-single-article.html#commentform](http://localhost:3000/pages/blog-single-article.html#commentform) or [http://10.0.2.2:3000/pages/blog-single-article.html#commentform](http://10.0.2.2:3000/pages/blog-single-article.html#commentform) in VirtualBox
4. Try autofilling the form fields

---

@grigs @mrgerardorodriguez @erikjung @saralohr 

Fixes #311

### DIFF
--- a/src/assets/toolkit/scripts/lib/component/float-label.js
+++ b/src/assets/toolkit/scripts/lib/component/float-label.js
@@ -2,26 +2,32 @@
 
 export class FloatLabel {
   /**
+   * @see https://mathiasbynens.be/notes/oninput
    * @param {DOM} element
    * @param {Object} options
    * @param {String} options.inputElement - Reference to input within element.
-   * @param {String} options.eventName - Event that will trigger updates.
    * @param {String} options.className - Class applied to `element` when `inputElement` is empty.
    */
   constructor(element, {
     inputElement = element.querySelector('input, textarea'),
-    eventName = 'input',
     className = 'is-empty'
   } = {}) {
     // Assign properties
     Object.assign(this, {
       element,
       inputElement,
-      eventName,
       className
     });
-    // Listen for event and update
-    this.element.addEventListener(this.eventName, this.update.bind(this));
+    // Fallback event handler
+    var keyUpHandler = () => this.update();
+    // Ideal event handler, will remove the other one if supported
+    var inputHandler = () => {
+      this.element.removeEventListener('keyup', keyUpHandler);
+      this.update();
+    };
+    // Attach events
+    this.element.addEventListener('input', inputHandler);
+    this.element.addEventListener('keyup', keyUpHandler);
     // Run first update now
     this.update();
   }
--- a/src/assets/toolkit/scripts/toolkit.js
+++ b/src/assets/toolkit/scripts/toolkit.js
@@ -20,9 +20,7 @@ import 'prismjs/components/prism-scss';
 (function() {
 
   arrayFromSelector('.js-FloatLabel').map(element => {
-    new FloatLabel(element, {
-      eventName: 'keyup'
-    });
+    new FloatLabel(element);
   });
 
   /**
--- a/src/patterns/combos/blog/post-comment-form.hbs
+++ b/src/patterns/combos/blog/post-comment-form.hbs
@@ -10,18 +10,22 @@ order: 5
     type="textarea"
     label="Message"
     inputclass="js-ElasticTextarea"
-    inputid="commentform-message"}}{{/embed}}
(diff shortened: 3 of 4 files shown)


tylersticka:
Assigning @mrgerardorodriguez to review my JavaScript in case we decide to try this out in WP-land soon.

src/assets/toolkit/scripts/lib/component/float-label.js
```diff
@@ -2,26 +2,32 @@
...
-      eventName,
       className
     });
-    // Listen for event and update
-    this.element.addEventListener(this.eventName, this.update.bind(this));
+    // Fallback event handler
+    var keyUpHandler = () => this.update();
+    // Ideal event handler, will remove the other one if supported
+    var inputHandler = () => {
+      this.element.removeEventListener('keyup', keyUpHandler);
```
mrgerardorodriguez: I don't know the answer to this, it's more of a _wonder_ on my part. I wonder if there is a way to check if the listener still exists so the listener doesn't have to be removed every time the `inputHandler` is called? I'm not sure you get dinged by removing the listener if it doesn't exist anymore. Just curious. 😊 Not a blocker IMO, though.

mrgerardorodriguez:
Just a _wonder_ thought inline. JS LGTM! 👍

src/assets/toolkit/scripts/lib/component/float-label.js
tylersticka: I wondered the same thing. Potentially, I could do something like this...

```js
// Create the generic, bound event handler
var eventHandler = () => this.update();
// Create a bound event handler for the first input event
var inputCheck = () => {
  // Remove the keyup handler
  this.element.removeEventListener('keyup', eventHandler);
  // Remove this check handler
  this.element.removeEventListener('input', inputCheck);
  // Add the generic event handler to this event
  this.element.addEventListener('input', eventHandler);
  // Fire the update
  this.update();
};
// Attach events
this.element.addEventListener('input', inputCheck);
this.element.addEventListener('keyup', eventHandler);
```

But that seemed like extra complexity, and Mathias's stuff (who was [the source](https://mathiasbynens.be/notes/oninput) for this) tends to be pretty performant.

I'd be interested in @erikjung's opinion here as well.

mrgerardorodriguez: > But that seemed like extra complexity, and Mathias's stuff (who was the source for this) tends to be pretty performant.

Very interesting. I like the spirit of that, but you are right, seems more complex than it needs to be. 😁 

What you have now is _totally_ fine. Yeah, curious if @erikjung has any new insights. 😄

tylersticka:
Did some hacky local stuff to the local pattern and I'm now having better luck testing.

## Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [ ] Opera
- [ ] Edge

tylersticka:
Okay, I can confirm this works in all the browsers mentioned in @grigs' original post.

I'm going to merge in this update so I can start implementing it in our theme.

PR #321, opened 2016-07-14
Extend pull utilites one more bump down the modular scale

saralohr:
This adds a root modular scale class to `u-pullSides` to provide a slightly smaller pull.

This is used to support a case in which the text in the pulled element is minimal and `u-pullSides1` felt a tad too big.

screen shot 2016-07-14 at 10 16 48 am

### DIFF
--- a/src/assets/toolkit/styles/utils/space.css
+++ b/src/assets/toolkit/styles/utils/space.css
@@ -84,6 +84,7 @@
 @mixin makeSpaceItemsUtils calc(var(--ms5)  * 1rem),  5;
 @mixin makeSpaceItemsUtils calc(var(--ms6)  * 1rem),  6;
 
+@mixin makePullUtils calc(var(--ms0) * 1rem);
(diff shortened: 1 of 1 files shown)


tylersticka:
How does this look on the narrowest screens? Does the container fail to reach the viewport edges on top-level comments?

saralohr:
Yes, it doesn't quite reach the edges on the top level. It felt a little strange to me that the top-level spanned the full width, and the reply did not. But the reply still looks kind of awkward.

Now I'm thinking of maybe dropping this class and leaving them with `u-pullSides1` since the comments stop spanning the full width between the small and medium breakports. The reply would span the full width up to the small breakpoint, and then indent in to avoid a between-the-breakpoints moment of a really long left side overhang. This is what I'm thinking:

comment-messages-3

tylersticka:
Yeah, I think that GIF looks great. Not sure what I'd change.

PR #322, opened 2016-07-15
CommentReply JS component

tylersticka:
For background: cloudfour/cloudfour.com-wp#198

Making this so @mrgerardorodriguez and @erikjung have something to easily review.

### DIFF
--- a/src/assets/toolkit/scripts/lib/component/comment-reply.js
+++ b/src/assets/toolkit/scripts/lib/component/comment-reply.js
@@ -1,93 +1,90 @@
 'use strict';
 
-import { u } from 'umbrellajs';
+import {u} from 'umbrellajs';
+import {ElasticTextarea} from './elastic-textarea';
+import {FloatLabel} from './float-label';
 
 export class CommentReply {
   constructor (element, {
-    formIdInputSelector = 'input[name="comment_parent"]',
-    formTemplate = '<p>No template provided.</p>',
-    idAttributeName = 'data-id',
-    insertAfterSelector = '.js-comment-insertFormAfter',
-    toggleSelector = '.js-comment-formToggle',
-    toggleShownClass = 'u-hidden',
+    idAttribute = 'data-id',
+    template = '<p>No template provided.</p>',
+    hiddenClass = 'u-hidden',
+    toggleSelector = '.js-comment-replyToggle',
+    insertMethod = 'after',
+    insertSelector = '.js-comment-insertReply',
     focusSelector = 'textarea, input',
-    onShow = null,
-    onHide = null
+    elasticTextareaSelector = '.js-ElasticTextarea',
+    floatLabelSelector = '.js-FloatLabel'
   } = {}) {
-    Object.assign(this, {
-      formIdInputSelector,
-      formTemplate,
-      toggleSelector,
-      toggleShownClass,
-      focusSelector,
-      onShow,
-      onHide
-    });
-
     this.element = u(element);
-    this.id = parseInt(this.element.attr(idAttributeName), 10);
-    this.insertAfterElement = this.element.find(insertAfterSelector);
-    this.toggleElement = this.element.find(toggleSelector);
-    this.toggleElement.handle('click', () => this.toggleForm());
+    this.id = this.element.attr(this.idAttribute);
+    this.template = template.replace(/\{\{id\}\}/g, this.id);
+    this.isOpen = false;
+    this.insertElement = this.element.find(insertSelector);
+    this.toggleElements = this.element.find(toggleSelector);
+
+    this.options = {
+      hiddenClass,
+      toggleSelector,
+      insertMethod,
+      focusSelector,
+      floatLabelSelector,
+      elasticTextareaSelector
+    };
+
+    this.toggleElements.handle('click', () => this.toggle());
   }
 
-  toggleForm () {
-    if (this.formElement) {
-      this.hideForm();
-    } else {
-      this.showForm();
+  toggle () {
+    if (this.isOpen) {
+      return this.close();
     }
+
+    return this.open();
   }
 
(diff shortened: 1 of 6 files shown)


mrgerardorodriguez:
Started looking at this right now, @tylersticka, but my brain is feeling moosh right now. LOL Gonna have to look at it Monday with a fresh pair of eyes and a working brain. 🙊 🙈

src/assets/toolkit/scripts/toolkit.js
```diff
@@ -46,4 +47,20 @@ import 'prismjs/components/prism-scss';
...
 
+  var replyFormTemplate = document.getElementById('comment-reply-form-template');
+
+  if (replyFormTemplate) {
+    replyFormTemplate = replyFormTemplate.innerHTML;
+    arrayFromSelector('.js-comment').map(element => {
+      new CommentReply(element, {
+        formTemplate: replyFormTemplate,
+        onShow: CommentReply => {
+          Array.from(CommentReply.formElement.querySelectorAll('.js-FloatLabel')).map(element => {
```
erikjung: Would `arrayFromSelector()` work here too?

src/assets/toolkit/scripts/lib/component/comment-reply.js
```diff
@@ -0,0 +1,116 @@
+'use strict';
+
+// these two functions should probably be moved to other files
+
+function stringToNode (str) {
+  var div = document.createElement('div');
+  div.innerHTML = str.trim();
+  return div.firstChild;
+}
+
```
erikjung: I tried to find a cleaner way to do this via `document.createDocumentFragment()`, but I failed.

erikjung:
@tylersticka 

Overall, seeing the DOM manipulation makes me wish were using jQuery, but I can't recommend any substantial improvements over how you're doing it. 

I think this is on a fine path as long as it gets heavily commented before getting merged.

👍

src/assets/toolkit/scripts/toolkit.js
tylersticka: No, because I need to use `querySelectorAll` relative to `CommentReply.formElement`.

mrgerardorodriguez:
Seems fine overall @tylersticka. I agree with @erikjung's comment on comments. 😊 

> Overall, seeing the DOM manipulation makes me wish were using jQuery, but I can't recommend any substantial improvements over how you're doing it.

This is definitely the most DOM manipulation we've had so far. Should we consider leveraging a DOM manipulation library now that we might actually benefit from it? I don't know...just wondering out loud. ¯\\\_(ツ)_/¯

src/assets/toolkit/scripts/toolkit.js
erikjung: Oh, I thought it supported a context in the second argument.

tylersticka:
@mrgerardorodriguez @erikjung I'm open to including a DOM manipulation library. I'm hesitant to add jQuery, which is `32kb` compressed and gzipped... that doesn't sound like a lot, but it would make up _most_ of our JavaScript payload for this single use-case. If I went that route, I'd definitely be tempted to add this to a separate JS file instead of the main `toolkit.js`... though if I did that, the `FloatLabel` stuff might get more complicated.

I went looking for leaner DOM manipulation libraries that would give us what we needed and are still maintained. Here's what I found:

- [Umbrella](http://umbrellajs.com/) (`3kb`)
- [DOMtastic](https://domtastic.js.org/) (`5kb`)
- [Dominus](https://github.com/bevacqua/dominus) (`4kb`)

Umbrella stands out to me for being the most active, the smallest file size, and the nicest in terms of documentation. But I'm open to other ideas!

tylersticka:
The more I look at Umbrella, the more I like it. Maybe because it's written and structured the way we would write/structure this sort of thing? Example that directly pertains to my WIP JS: https://github.com/umbrellajs/umbrella/blob/master/src/plugins/after/after.js

erikjung:
@tylersticka 

Those all look great. I was going to say that I prefer the latter two simply because they don't include any AJAX, but I don't feel too strongly about it. They all looks nice and appear to be actively maintained. I'm all for any of them 👍

tylersticka:
> I was going to say that I prefer the latter two simply because they don't include any AJAX, but I don't feel too strongly about it.

I'd feel the same way, except even with AJAX included Umbrella is smaller. ¯\\\_(ツ)_/¯

mrgerardorodriguez:
Oh, great finds @tylersticka! Yeah, I'm cool with going with Umbrella since it's smaller. :)

tylersticka:
@mrgerardorodriguez @erikjung Just pushed some updates to this PR with Umbrella in use. It works really well, and would simplify some stuff in other components (one step at a time, though).

One thing I'm still struggling with a bit is @erikjung's suggestion to manipulate attributes of the template directly instead of a string replacement type of deal. I can't help but wonder if a simple `formTemplate.replace(/{{id}}/g, this.id)` wouldn't feel a bit nicer than the unique ID and input value logic I'm doing right now. Or maybe I just haven't executed on this in the intended way?

erikjung:
@tylersticka I think the simple solution is fine for this, especially If don't need nested data binding.

erikjung:
(Meaning, regex string replacement 👍)

erikjung:
@tylersticka For clarification, here is a sloppy example of how I thought it could work with the `data-bind`:

http://codepen.io/erikjung/pen/QEQrxG?editors=1010

tylersticka:
@erikjung @mrgerardorodriguez I just removed the `[WIP]` from the issue title. Ready for re-review. Changed a lot in the last few commits.

erikjung:
LGTM 👍

mrgerardorodriguez:
Sweet! 👍

PR #325, opened 2016-07-18
Update manifest.json based on latest from WP

tylersticka:
This was accidentally updated in the wrong repo, and I’m tired of seeing it updated whenever the pattern tasks run in our theme.

Also converted tabs to spaces because YOLO.

See: cloudfour/cloudfour.com-wp#170

---

@mrgerardorodriguez @erikjung


mrgerardorodriguez:
Ha ha, thanks! 👍

PR #326, opened 2016-07-18
Add vendor styles for Jetpack Markdown footnotes

tylersticka:
This begins to address cloudfour/cloudfour.com-wp#255 by customizing the markup [Jetpack Markdown](https://en.support.wordpress.com/markdown-quick-reference/) outputs for footnotes.

I went this route after confirming that there are no documented filters for the footnotes, and any that existed in the underlying library are not exposed in a way we could modify without rolling our own plugin.

The only thing it does not do is replace the unicode character with our SVG icon. This is possible, but it would represent the bulk of the styles, and filtering for the unicode character is one of the few things we can do relatively easily from our WP theme if we desire.

## Before

screen shot 2016-07-18 at 3 36 14 pm

## After

screen shot 2016-07-18 at 3 35 58 pm

---

@saralohr @mrgerardorodriguez @erikjung @grigs 

See: cloudfour/cloudfour.com-wp#255

### DIFF
--- a/src/assets/toolkit/styles/components/footnote-link.css
+++ b/src/assets/toolkit/styles/components/footnote-link.css
@@ -10,39 +10,45 @@
   --FootnoteLink-padding: 0 calc(var(--ms-6) * 1em);
 }
 
-/**
- * Override color defaults for all states
- */
+@define-mixin makeFootnoteLinkComponent {
+  /**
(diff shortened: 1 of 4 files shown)


saralohr:
Cool! Sooo much nicer. 👍

PR #327, opened 2016-07-18
Fix dumb typo because I’m a stupid 🐒

tylersticka:
Was tearing my hair out with this and it turns out I'm a dumb dumb.

---

@mrgerardorodriguez @erikjung


mrgerardorodriguez:
Ah! 🙈 

👍

PR #329, opened 2016-07-19
Specify text-decoration-skip property

### LINKED ISSUE #323: Safari text decoration
Recently, Safari began natively styling underlines in the "cutout" fashion that many have grown to love on sites like Medium. This appearance is typically achieved in a pseudo-hack-like way using a `text-shadow` that matches the background color.

I noticed that our site lacks this styling:

Actual:
screen shot 2016-07-18 at 10 54 14 am

Expected:
screen shot 2016-07-18 at 10 54 52 am

After some detective work, I found that Normalize.css (via **suitcss-base**) is introducing this reset:

https://github.com/necolas/normalize.css/blob/master/normalize.css#L99-L107

Should we override this reset? Also, should we assess whether or not we still need to include `suitcss-base` at all? Maybe we can hand-pick the things we need and bake them in?

tylersticka:
Allows descenders and other glyphs to disrupt underlines.

In @erikjung's original issue describing this problem (#323), he wonders whether or not it makes sense to omit or incorporate the necessary `suitcss-base`/`normalize.css` fixes directly. I commented out `suitcss-base` and noticed numerous regressions, which gave me cold feet. If we think a refactor like that is a good idea, I think it should be tackled as a separate issue.

### Before

screen shot 2016-07-19 at 9 15 52 am

### After

screen shot 2016-07-19 at 9 15 38 am

---

@erikjung @saralohr @mrgerardorodriguez @aileenjeffries 

Fixes #323

### DIFF
--- a/src/assets/toolkit/styles/base/base.css
+++ b/src/assets/toolkit/styles/base/base.css
@@ -40,6 +40,7 @@ body {
 a {
   text-decoration: underline;
   color: var(--link-color);
+  -webkit-text-decoration-skip: ink;
 }
 
 a:matches(:active, :focus, :hover) {


mrgerardorodriguez:
👍

PR #330, opened 2016-07-19
Specify msapplication-config

### LINKED ISSUE #319: Windows tile image doesn't work
Despite linking to [this image](https://cloudfour.com/mstile-150x150.png), which appears to resolve, on my home PC the tile image does not show up when pinning the site to start from Edge:

tile

The tile color is also incorrect, which makes me think the [the config](https://cloudfour.com/browserconfig.xml) isn't being picked up for some reason.

tylersticka:
This is necessary to get the tile image working in Edge. This will also need to be done in WP-land.

See: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/101636/

## Before

virtualbox_msedge - win10th2_19_07_2016_09_28_46

## After

(I don't know why the "after" added it to the top and not the bottom.)

virtualbox_msedge - win10th2_19_07_2016_09_35_04

---

@saralohr @mrgerardorodriguez @erikjung 

Fixes #319

### DIFF
--- a/src/templates/default.hbs
+++ b/src/templates/default.hbs
@@ -24,6 +24,7 @@
       <link rel="manifest" href="/manifest.json">
       <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#456BD9">
       <meta name="theme-color" content="#456BD9">
+      <meta name="msapplication-config" content="/browserconfig.xml">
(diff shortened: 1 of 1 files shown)


mrgerardorodriguez:
👍

PR #335, opened 2016-07-21
Feature: ToggleControl Component

tylersticka:
I'm in the process of restoring the "notify me of follow-up comments" feature, but my goal is for the form to look as nice and simple as it did before. Something that I thought would help: A friendlier style for checkboxes.

While I was at it, I figured I'd also style a similar treatment for `radio` inputs.

I took a slightly different tactic with these than I have in the past. I decided to require the `<label>` to eliminate an extra element and to make positioning the indicator and label text easier for the end user. I still tried to define the `[type]` styles in such a way that if we decide that was a bad idea later on, we can decouple stuff with relative ease.

screen shot 2016-07-21 at 2 06 14 pm

---

@saralohr @erikjung @mrgerardorodriguez @nicolemors @aileenjeffries


tylersticka:
Here's how it looks in the context of that form in my WIP subscriptions branch:

screen shot 2016-07-21 at 2 16 47 pm

saralohr:
These look nice! 👍

PR #337, opened 2016-07-22
Wip componentize trivia

saralohr:
This isn't quite ready for a PR, but I wanted to get opinions on a couple different options for the slideshow animation:

Currently, adding a `slide-forward` or `slide-back` class to the parent `ul` triggers an animation on the element. For each click forward or back, the above classes are removed, and then re-added depending on the direction. If the class ends up being the same, it is not read as a change, since the end result is the same, and the animation does not run again.

I found a few ways around this, and tried out a couple of these that seemed the least offensive.

One is to trigger a reflow by checking the element's width. (this is what is currently in place)

The other is to alternate the class names being added, with the alternate class name running an identical animation under a different name. (this is present but commented out currently)

So, the question is: is it better to reflow, or have some bloat in the CSS? (we could use mixins to make it maintainable, but still adds bloat to the final output) Or, something else altogether?

I got ideas from [this article](https://css-tricks.com/restart-css-animation/).

NOTE: You can ignore `slideshow1.js`. This is an alternate version I was working that didn't come to fruition. I'll remove it before final PR.

cc @tylersticka @erikjung @mrgerardorodriguez

### DIFF
--- a/src/assets/toolkit/scripts/lib/component/slideshow.js
+++ b/src/assets/toolkit/scripts/lib/component/slideshow.js
@@ -8,9 +8,12 @@ export class Slideshow {
     prevTrigger = element.querySelector('.js-Slideshow-prev'),
     currentCount = element.querySelector('.js-Slideshow-current'),
     totalCount = element.querySelector('.js-Slideshow-total'),
-    classNameVisible = 'is-visible',
-    classNameForward = 'slide-forward',
-    classNameBack = 'slide-back'
+    classIsVisible = 'is-visible',
+    classWasVisible = 'was-visible',
+    classForward = 'slide-forward',
+    classBack = 'slide-back',
+    classForwardAlt = 'slide-forward-alt',
+    classBackAlt = 'slide-back-alt'
   } = {}) {
 
     this.slides = Array.from(slides);
@@ -20,13 +23,17 @@ export class Slideshow {
 
     Object.assign(this, {
       element,
+      slideHolder,
       nextTrigger,
       prevTrigger,
       currentCount,
       totalCount,
-      classNameVisible,
-      classNameForward,
-      classNameBack
+      classIsVisible,
+      classWasVisible,
+      classForward,
+      classBack,
+      classForwardAlt,
+      classBackAlt
     });
 
     this.attachEvents();
@@ -43,20 +50,47 @@ export class Slideshow {
     // Update the navigation with the current slide number
     this.currentCount.innerHTML = slideToShow + 1;
 
+    // Remove previously applied directional class
+    this.slideHolder.classList.remove(this.classForward, this.classBack, this.classForwardAlt, this.classBackAlt);
+
+    // This triggers a reflow which gives a fresh start
+    // for adding back the same class name so the same animation
+    // can re-run
+    this.slideHolder.offsetWidth;
+
+    // Add new directional class to slide container
+    if (direction == 'forward') {
+      // Alternate class names to apply a different animation since you can't
+      // run the same animation twice in a row
+      // if (this.counter % 2) {
+      //   this.slideHolder.classList.add(this.classForward);
+      // } else {
+      //   this.slideHolder.classList.add(this.classForwardAlt);
+      // }
+      this.slideHolder.classList.add(this.classForward);
+    } else {
+      // if (this.counter % 2) {
+      //   this.slideHolder.classList.add(this.classBack);
+      // } else {
+      //   this.slideHolder.classList.add(this.classBackAlt);
+      // }
+      this.slideHolder.classList.add(this.classBack);
+    }
+
+    this.slides.forEach(slide => {
+      slide.classList.remove(this.classWasVisible);
+    });
+
     // Remove class names from all items
     this.slides.forEach(slide => {
-      slide.classList.remove(this.classNameVisible, this.classNameForward, this.classNameBack);
+      if (slide.classList.contains(this.classIsVisible)) {
+        slide.classList.remove(this.classIsVisible);
+        slide.classList.add(this.classWasVisible);
+      }
     });
 
     // Add current class to current item
-    this.slides[slideToShow].classList.add(this.classNameVisible);
-
-    // Add forward or back class depending on direction
-    if (direction == 'forward') {
-      this.slides[slideToShow].classList.add(this.classNameForward);
-    } else if (direction == 'back') {
-      this.slides[slideToShow].classList.add(this.classNameBack);
-    }
+    this.slides[slideToShow].classList.add(this.classIsVisible);
   }
 
 
--- a/src/assets/toolkit/scripts/lib/component/slideshow.js
+++ b/src/assets/toolkit/scripts/lib/component/slideshow.js
@@ -53,9 +53,8 @@ export class Slideshow {
     // Remove previously applied directional class
     this.slideHolder.classList.remove(this.classForward, this.classBack, this.classForwardAlt, this.classBackAlt);
 
-    // This triggers a reflow which gives a fresh start
-    // for adding back the same class name so the same animation
-    // can re-run
+    // Trigger a reflow which gives a fresh start to the element, so we can add
+    // back the same class that runs the same animation
     this.slideHolder.offsetWidth;
 
     // Add new directional class to slide container
@@ -77,11 +76,12 @@ export class Slideshow {
       this.slideHolder.classList.add(this.classBack);
     }
 
+    // Remove `was-visible` class
     this.slides.forEach(slide => {
       slide.classList.remove(this.classWasVisible);
     });
 
-    // Remove class names from all items
+    // Find slide that has class `is-visible`, and replace it with `was-visible`
     this.slides.forEach(slide => {
       if (slide.classList.contains(this.classIsVisible)) {
         slide.classList.remove(this.classIsVisible);
@@ -89,7 +89,7 @@ export class Slideshow {
       }
     });
 
-    // Add current class to current item
+    // Add `is-visible` class to current item
     this.slides[slideToShow].classList.add(this.classIsVisible);
   }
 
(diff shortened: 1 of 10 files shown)


src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,111 @@
...
+    // Update the navigation with the current slide number
+    this.currentCount.innerHTML = slideToShow + 1;
+
+    // Remove previously applied directional class
+    this.slideHolder.classList.remove(this.classForward, this.classBack, this.classForwardAlt, this.classBackAlt);
+
+    // Trigger a reflow which gives a fresh start to the element, so we can add
+    // back the same class that runs the same animation
+    this.slideHolder.offsetWidth;
+
```
saralohr: This is the reflow option.

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,111 @@
...
+    // Add new directional class to slide container
+    if (direction == 'forward') {
+      // Alternate class names to apply a different animation since you can't
+      // run the same animation twice in a row
+      // if (this.counter % 2) {
+      //   this.slideHolder.classList.add(this.classForward);
+      // } else {
+      //   this.slideHolder.classList.add(this.classForwardAlt);
+      // }
```
saralohr: This (and below) is the alternate class option.

src/assets/toolkit/styles/utils/position.css
```diff
@@ -1,5 +1,13 @@
 @import "suitcss-utils-position";
 
+.u-posAbsoluteMiddle {
```
tylersticka: Is this different from `u-posAbsoluteCenter`?

src/assets/toolkit/scripts/lib/component/slideshow1.js
```diff
@@ -0,0 +1,134 @@
...
+    });
+
+    this.attachEvents();
+  }
+
+  updateCurrent () {
+    if (this.counter > 0) {
+      var currentSlide = (this.counter % this.numSlides);
+    } else {
+      var currentSlide = (this.counter % this.numSlides) ? ((this.counter % this.numSlides) + this.numSlides) : 0;
```
tylersticka: I've tried for the last ten years and I still don't understand what the `%` operator does. 😓

src/assets/toolkit/scripts/lib/component/slideshow1.js
```diff
@@ -0,0 +1,134 @@
...
+    nextTrigger = element.querySelector('.js-Slideshow-next'),
+    prevTrigger = element.querySelector('.js-Slideshow-prev'),
+    currentCount = element.querySelector('.js-Slideshow-current'),
+    totalCount = element.querySelector('.js-Slideshow-total'),
+    classForward = 'slide-forward',
+    classForwardReset = 'slide-forwardReset',
+    classBack = 'slide-back'
+  } = {}) {
+
+    this.slides = Array.from(slides);
```
tylersticka: You might consider using [Umbrella](http://umbrellajs.com/) (already a dependency) which would render this sorta stuff unnecessary.

src/assets/toolkit/styles/utils/position.css
saralohr: Yes, it just vertically aligns the element, and doesn't do anything with left-to-right alignment. I having an issue in which I wanted an element to span the full width, and it was being compacted by the left positioning. I might be able to fix this by altering properties on the element though. I was planning to go back through these classes with fine-toothed comb, because some of them work might be WIP that aren't relevant anymore.

src/patterns/components/slideshow/base.hbs
```diff
@@ -0,0 +1,45 @@
...
+          </div>
+        </div>
+        <div class="Grid-cell u-sm-size3of4 u-lg-size1of2 u-spaceItems06 u-textCenter u-sm-textLeft u-spaceTop01 u-sm-spaceTopNone">
+          <h4 class="u-md-textGrow1">Did you know&hellip;?</h4>
+          <p>{{fact}}</p>
+        </div>
+      </li>
+    {{/each}}
+  </ul>
+
```
saralohr: Note that this is totally not accessible as it sits. That's next up on the TODO list.

src/assets/toolkit/scripts/lib/component/slideshow1.js
```diff
@@ -0,0 +1,134 @@
...
+    this.numSlides = this.slides.length;
```
tylersticka: You might consider defining [a "getter" method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) for this. So instead of defining it here, you'd have a method like...

```
get numSlides() {
  return this.slides.length;
}
```

Then you could just refer to `this.numSlides` and it'd always return that value. Magic!

src/assets/toolkit/scripts/lib/component/slideshow1.js
```diff
@@ -0,0 +1,134 @@
...
+    this.counter = 0;
+    totalCount.innerHTML = this.numSlides;
```
tylersticka: Consider making your element variable names clear, for example `totalElement`. It can be a little confusing to remember if `totalCount` or `numSlides` is the total or the element.

src/assets/toolkit/scripts/lib/component/slideshow1.js
```diff
@@ -0,0 +1,134 @@
...
+    } else {
+      var currentSlide = (this.counter % this.numSlides) ? ((this.counter % this.numSlides) + this.numSlides) : 0;
+    }
+
+    // Update the navigation with the current slide number
+    this.currentCount.innerHTML = currentSlide + 1;
+  }
+
+  getCurrentPosition (el) {
+    const transformMatrix = getComputedStyle(el).transform;
```
tylersticka: Does `el.style.transform` not work?

src/assets/toolkit/scripts/lib/component/slideshow1.js
```diff
@@ -0,0 +1,134 @@
...
+    const slideHolderWidth = el.offsetWidth;
+    var matrixValues = transformMatrix.split('(')[1],
+    matrixValues = matrixValues.split(')')[0],
+    matrixValues = matrixValues.split(',');
+    const translateX = matrixValues[4];
+    const currentPosition = this.convertToPercent(translateX, slideHolderWidth);
+    return currentPosition;
```
tylersticka: This method makes my head hurt!

src/assets/toolkit/scripts/lib/component/slideshow1.js
saralohr: I should prolly have deleted this file, but it is currently not being used at all. You can disregard its existence.

src/assets/toolkit/scripts/lib/component/slideshow1.js
```diff
@@ -0,0 +1,134 @@
...
+
+  findKeyframesRule(rule) {
+    var ss = document.styleSheets;
+    console.log(ss);
+    for (var j = 0; j < ss[1].cssRules.length; ++j) {
+      if (ss[1].cssRules[j].type == window.CSSRule.WEBKIT_KEYFRAMES_RULE &&
+      ss[1].cssRules[j].name == rule) {
+        return ss[1].cssRules[j]; }
+    }
+    return null;
```
tylersticka: What is this doing?

saralohr: This is a route I was considering taking to change the keyframe properties dynamically. It got really complex and I abandoned it.

src/assets/toolkit/scripts/lib/component/slideshow1.js
```diff
@@ -0,0 +1,134 @@
...
+      this.updateCurrent();
+      this.slide(-1);
+    });
+
+    this.prevTrigger.addEventListener('click', event => {
+      event.preventDefault();
+      this.counter--;
+      this.updateCurrent();
+      this.slide(1);
+    });
```
tylersticka: I would probably make methods for `next` and `previous`, then use Umbrella's [`handle()` method](http://umbrellajs.com/documentation#handle) to attach them in the constructor:

```js
this.nextTrigger.handle('click', () => this.next());
```

tylersticka:
Some nitpicky feedback from looking at the code, but I will likely have broader feedback soon. Still digging in a bit.

saralohr:
@tylersticka I kinda buried this comment in the opening PR comment, but `slideshow1.js` is the route I ended up not taking, and currently is unused. I appreciate the feedback on it though, for learning! and some of those comments apply to `slideshow.js` as well. Just wanted to make that clear so you didn't spend too much time figuring out what's what when you dig in.

src/assets/toolkit/scripts/lib/component/slideshow1.js
saralohr: Not sure... reading up on the difference now. It probably does work, this is the only thing I tried.

src/assets/toolkit/scripts/lib/component/slideshow1.js
saralohr: 😬  Yeah, it's kinda gnarly. I should probably put the matrix stuff in it's own method.

tylersticka:
Sorry for the delay, @saralohr.

Overall, your `slideshow.js` implementation is a big improvement over previous attempts. I like it a lot.

That said, I'd still like to challenge you to simplify a little more. And I'm going to leave my own exploration here: http://codepen.io/tylersticka/pen/0ba0c878cf741e7f4546711af5fa8b39

It lacks some of the more sophisticated aspects of yours, but it makes me think that there might be a way to simplify this sorta thing further.

But that might just be my Friday brain talking. ¯\_(ツ)_/¯

saralohr:
> That said, I'd still like to challenge you to simplify a little more. And I'm going to leave my own exploration here as possible inspiration.

Thanks @tylersticka! I'll check it out.

PR #285, opened 2016-07-07
Add width/height to base iframe

mrgerardorodriguez:
Re: https://github.com/cloudfour/cloudfour.com-patterns/pull/281

This adds a base `width` and `height` to the `<iframe>` element. This allows the `FlexEmbed` component to be happy.

---

cc: @erikjung @tylersticka @saralohr

PR #341, opened 2016-07-25
Move base iframe styles to FlexEmbed component

### LINKED ISSUE #338: CodePen (possibly more?) too short in blog posts due to height: 100%
We added a base `height: 100%;` rule to `iframe` as part of #281.

Unfortunately, this is overriding `height` attributes we want to keep around. Specifically, I can't get CodePen embeds to be taller than `150px` no matter what embed technique I use:

Example from [this post](https://cloudfour.com/thinks/breaking-out-with-viewport-units-and-calc/):

screen shot 2016-07-25 at 11 32 00 am

If this rule was added to make `FlexEmbed` happy, we should consider scoping it to that component instead of applying it globally.

---

@mrgerardorodriguez

tylersticka:
Allows other iframe embeds to continue to manage their own size and aspect ratio.

## Before

screen shot 2016-07-25 at 11 32 00 am

## After

screen shot 2016-07-25 at 12 32 33 pm

---

@mrgerardorodriguez @saralohr @erikjung 

Fixes #338 

Related: #285 #281

### DIFF
--- a/src/assets/toolkit/styles/base/base.css
+++ b/src/assets/toolkit/styles/base/base.css
@@ -79,8 +79,3 @@ pre {
 img {
   max-width: 100%;
 }
-
-iframe {
-  width: 100%;
-  height: 100%;
-}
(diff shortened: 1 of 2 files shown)


mrgerardorodriguez:
👍

PR #345, opened 2016-07-25
Remove eventName override for elastic textareas

### LINKED ISSUE #344: ElasticTextarea: Delay on expand
We currently default to `keyup` for `ElasticTextarea`. But compared to the `input` event, this has a noticeable delay.

We solved this problem for `FloatLabel` by moving the test to the event handling of the component itself (see #311). Maybe there's a way to move this test logic to a component that could be reused for either?

## `keyup`

elastictextarea-keyup

## `input`

elastictextarea-input

tylersticka:
The `input` event is extremely well-supported. Opera Mini is the only modern browser that doesn’t support it, but `keyup` won’t work as a fallback, so it’s unnecessary. Unlike the `FloatLabel` component, there is no perceivably UX degradation if the event fails to fire.

---

@mrgerardorodriguez @erikjung @saralohr 

Fixes #344

### DIFF
--- a/src/assets/toolkit/scripts/toolkit.js
+++ b/src/assets/toolkit/scripts/toolkit.js
@@ -42,9 +42,7 @@ import 'prismjs/components/prism-scss';
   });
 
   u('.js-ElasticTextarea').map(element => {
-    new ElasticTextarea(element, {
-      eventName: 'keyup'
-    });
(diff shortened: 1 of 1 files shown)


mrgerardorodriguez:
👍

saralohr:
👍

PR #350, opened 2016-07-26
Add trivia slider

saralohr:
I decided to open a new PR for this to keep things cleaner. You can reference PR #337 to see some of the feedback addressed here.

Requesting JS review @erikjung or @mrgerardorodriguez.

trivia-slider

cc @tylersticka

### DIFF
--- a/src/assets/toolkit/scripts/lib/component/slideshow.js
+++ b/src/assets/toolkit/scripts/lib/component/slideshow.js
@@ -70,8 +70,9 @@ export class Slideshow {
 
     // Find slide that has class `is-visible`, and replace it with `was-visible`
     this.slides.forEach(slide => {
-      if (u(slide).hasClass(this.classIsVisible)) {
-        u(slide).removeClass(this.classIsVisible).addClass(this.classWasVisible);
+      var slide = u(slide);
+      if (slide.hasClass(this.classIsVisible)) {
+        slide.removeClass(this.classIsVisible).addClass(this.classWasVisible);
       }
     });
 
--- a/src/assets/toolkit/scripts/lib/component/slideshow.js
+++ b/src/assets/toolkit/scripts/lib/component/slideshow.js
@@ -1,6 +1,15 @@
 'use strict';
+
 import {u} from 'umbrellajs';
+
+/**
+ * Class representing the Slideshow component
+ */
 export class Slideshow {
+  /**
+   * @parameter {DOM} element - The containing element for Slideshow
+   * @parameter {Object} options - Define the inner elements
+   */
   constructor (element, {
     slideHolderSelector = '.js-Slideshow-slides',
     slidesSelector = '.js-Slideshow-slide',
@@ -16,6 +25,7 @@ export class Slideshow {
     classIsBack = 'is-back'
   } = {}) {
 
+    // Store elements as Umbrella objects
     const slideHolder = u(slideHolderSelector, element);
     const slides = u(slidesSelector, element);
     const nextTrigger = u(nextTriggerSelector, element);
@@ -39,16 +49,28 @@ export class Slideshow {
     });
 
     this.counter = 0;
+
+    // Insert total number of slides into the DOM
     this.totalCountElement.text(this.numSlides);
+
+    // Add event listener for interaction with slide controls
     this.nextTrigger.handle('click', this.nextSlide.bind(this));
     this.prevTrigger.handle('click', this.prevSlide.bind(this));
-    this.slideHolder.on('animationend', this.onAnimationEnd.bind(this));
+
+    // Add event listener for end of animation
+    this.slideHolder.handle('animationend', this.onAnimationEnd.bind(this));
   }
 
+  /**
+   * Getter for the total number of slides
+   */
   get numSlides() {
     return this.slides.length;
   }
 
+  /**
+   * Getter for the index of the current slide
+   */
   get slideToShow() {
     var slideToShow;
     var slideRemainder = this.counter % this.numSlides;
@@ -63,6 +85,11 @@ export class Slideshow {
     return slideToShow;
   }
 
+  /**
+   * Slide the slider
+   *
+   * @parameter {string} direction - Accepts 'forward' or 'back'
+   */
   slide(direction) {
     // Update the navigation with the current slide number
     this.currentCountElement.text(this.slideToShow + 1);
@@ -92,15 +119,24 @@ export class Slideshow {
     }
   }
 
+  /**
+   * Remove the animation class
+   */
   onAnimationEnd() {
     this.slideHolder.removeClass(this.classIsSlidingForward, this.classIsSlidingBack);
   }
 
+  /**
+   * Slide forward
+   */
   nextSlide() {
     this.counter++;
     this.slide('forward');
   }
 
+  /**
+   * Slide back
+   */
   prevSlide() {
     this.counter--;
     this.slide('back');
--- a/src/assets/toolkit/scripts/lib/component/slideshow.js
+++ b/src/assets/toolkit/scripts/lib/component/slideshow.js
@@ -33,8 +33,8 @@ export class Slideshow {
 
     this.counter = 0;
     u(this.totalCountElement).text(this.numSlides);
-    u(this.nextTrigger).handle('click', () => this.next());
-    u(this.prevTrigger).handle('click', () => this.prev());
+    u(this.nextTrigger).handle('click', () => this.nextSlide());
+    u(this.prevTrigger).handle('click', () => this.prevSlide());
   }
 
   get numSlides() {
@@ -93,12 +93,12 @@ export class Slideshow {
     u(this.slideHolder).on("animationend", onAnimationEnd, false);
   }
 
-  next() {
+  nextSlide() {
     this.counter++;
     this.slide('forward');
   }
 
-  prev() {
+  prevSlide() {
     this.counter--;
     this.slide('back');
   }
--- a/src/assets/toolkit/scripts/lib/component/slideshow.js
+++ b/src/assets/toolkit/scripts/lib/component/slideshow.js
@@ -64,9 +64,7 @@ export class Slideshow {
     u(this.slideHolder).removeClass(this.classIsForward, this.classIsBack);
 
     // Remove previous 'was-visible' class
-    this.slides.forEach(slide => {
-      u(slide).removeClass(this.classWasVisible);
-    });
+    u(this.slides).removeClass(this.classWasVisible);
 
     // Find slide that has class `is-visible`, and replace it with `was-visible`
     this.slides.forEach(slide => {
--- a/src/assets/toolkit/scripts/lib/component/slideshow.js
+++ b/src/assets/toolkit/scripts/lib/component/slideshow.js
@@ -78,7 +78,7 @@ export class Slideshow {
     u(this.slides[this.slideToShow]).addClass(this.classIsVisible);
 
     // Add new directional classes to slide container
-    if (direction == 'forward') {
+    if (direction === 'forward') {
       u(this.slideHolder).addClass(this.classIsSlidingForward, this.classIsForward);
     } else {
       u(this.slideHolder).addClass(this.classIsSlidingBack, this.classIsBack);
@@ -89,7 +89,7 @@ export class Slideshow {
       u(this.slideHolder).removeClass(this.classIsSlidingForward, this.classIsSlidingBack);
     }
 
-    u(this.slideHolder).on("animationend", onAnimationEnd, false);
+    u(this.slideHolder).on('animationend', onAnimationEnd, false);
   }
 
   nextSlide() {
--- a/src/assets/toolkit/scripts/lib/component/slideshow.js
+++ b/src/assets/toolkit/scripts/lib/component/slideshow.js
@@ -32,7 +32,7 @@ export class Slideshow {
     });
 
     this.counter = 0;
-    u(this.totalCountElement).html(this.numSlides);
+    u(this.totalCountElement).text(this.numSlides);
     u(this.nextTrigger).handle('click', () => this.next());
     u(this.prevTrigger).handle('click', () => this.prev());
   }
@@ -43,15 +43,18 @@ export class Slideshow {
 
   slide(direction) {
 
+    var slideToShow;
+    var slideRemainder = this.counter % this.numSlides;
+
     // Get the index of the current item
     if (this.counter > 0) {
-      var slideToShow = (this.counter % this.numSlides);
+      var slideToShow = slideRemainder;
     } else {
-      var slideToShow = (this.counter % this.numSlides) ? ((this.counter % this.numSlides) + this.numSlides) : 0;
+      var slideToShow = slideRemainder ? (slideRemainder + this.numSlides) : 0;
     }
 
     // Update the navigation with the current slide number
-    u(this.currentCountElement).html(slideToShow + 1);
+    u(this.currentCountElement).text(slideToShow + 1);
 
     // Remove previous directional class
     u(this.slideHolder).removeClass(this.classIsForward, this.classIsBack);
--- a/src/assets/toolkit/scripts/lib/component/slideshow.js
+++ b/src/assets/toolkit/scripts/lib/component/slideshow.js
@@ -1,0 +1,99 @@
+'use strict';
+import {u} from 'umbrellajs';
+export class Slideshow {
+  constructor (element, {
+    slideHolder = element.querySelector('.js-Slideshow-slides'),
+    slides = element.querySelectorAll('.js-Slideshow-slide'),
+    nextTrigger = element.querySelector('.js-Slideshow-next'),
+    prevTrigger = element.querySelector('.js-Slideshow-prev'),
+    currentCountElement = element.querySelector('.js-Slideshow-current'),
+    totalCountElement = element.querySelector('.js-Slideshow-total'),
+    classIsVisible = 'is-visible',
+    classWasVisible = 'was-visible',
+    classIsSlidingForward = 'is-sliding-forward',
+    classIsSlidingBack = 'is-sliding-back',
+    classIsForward = 'is-forward',
+    classIsBack = 'is-back'
+  } = {}) {
+
+    Object.assign(this, {
+      slides,
+      slideHolder,
+      nextTrigger,
+      prevTrigger,
+      currentCountElement,
+      totalCountElement,
+      classIsVisible,
+      classWasVisible,
+      classIsForward,
+      classIsBack,
+      classIsSlidingForward,
+      classIsSlidingBack
+    });
+
+    this.counter = 0;
+    u(this.totalCountElement).html(this.numSlides);
+    u(this.nextTrigger).handle('click', () => this.next());
+    u(this.prevTrigger).handle('click', () => this.prev());
+  }
+
+  get numSlides() {
+    return this.slides.length;
+  }
+
+  slide(direction) {
+
+    // Get the index of the current item
+    if (this.counter > 0) {
+      var slideToShow = (this.counter % this.numSlides);
+    } else {
+      var slideToShow = (this.counter % this.numSlides) ? ((this.counter % this.numSlides) + this.numSlides) : 0;
+    }
+
+    // Update the navigation with the current slide number
+    u(this.currentCountElement).html(slideToShow + 1);
+
+    // Remove previous directional class
+    u(this.slideHolder).removeClass(this.classIsForward, this.classIsBack);
+
+    // Remove previous 'was-visible' class
+    this.slides.forEach(slide => {
+      u(slide).removeClass(this.classWasVisible);
+    });
+
+    // Find slide that has class `is-visible`, and replace it with `was-visible`
+    this.slides.forEach(slide => {
+      if (u(slide).hasClass(this.classIsVisible)) {
+        u(slide).removeClass(this.classIsVisible);
+        u(slide).addClass(this.classWasVisible);
+      }
+    });
+
+    // Add `is-visible` class to current item
+    u(this.slides[slideToShow]).addClass(this.classIsVisible);
+
+    // Add new directional classes to slide container
+    if (direction == 'forward') {
+      u(this.slideHolder).addClass(this.classIsSlidingForward, this.classIsForward);
+    } else {
+      u(this.slideHolder).addClass(this.classIsSlidingBack, this.classIsBack);
+    }
+
+    // Remove animation class so it can be re-applied as a new animation
+    var onAnimationEnd = () => {
+      u(this.slideHolder).removeClass(this.classIsSlidingForward, this.classIsSlidingBack);
+    }
+
+    u(this.slideHolder).on("animationend", onAnimationEnd, false);
+  }
+
+  next() {
+    this.counter++;
+    this.slide('forward');
+  }
+
+  prev() {
+    this.counter--;
+    this.slide('back');
+  }
+}
--- a/src/assets/toolkit/scripts/lib/component/slideshow.js
+++ b/src/assets/toolkit/scripts/lib/component/slideshow.js
@@ -2,12 +2,12 @@
 import {u} from 'umbrellajs';
 export class Slideshow {
   constructor (element, {
-    slideHolder = u('.js-Slideshow-slides'),
-    slides = u('.js-Slideshow-slide'),
-    nextTrigger = u('.js-Slideshow-next'),
-    prevTrigger = u('.js-Slideshow-prev'),
-    currentCountElement = u('.js-Slideshow-current'),
-    totalCountElement = u('.js-Slideshow-total'),
+    slideHolderSelector = '.js-Slideshow-slides',
+    slidesSelector = '.js-Slideshow-slide',
+    nextTriggerSelector = '.js-Slideshow-next',
+    prevTriggerSelector = '.js-Slideshow-prev',
+    currentCountElementSelector = '.js-Slideshow-current',
+    totalCountElementSelector = '.js-Slideshow-total',
     classIsVisible = 'is-visible',
     classWasVisible = 'was-visible',
     classIsSlidingForward = 'is-sliding-forward',
@@ -15,6 +15,13 @@ export class Slideshow {
     classIsForward = 'is-forward',
     classIsBack = 'is-back'
   } = {}) {
+
+    const slideHolder = u(slideHolderSelector, element);
+    const slides = u(slidesSelector, element);
+    const nextTrigger = u(nextTriggerSelector, element);
+    const prevTrigger = u(prevTriggerSelector, element);
+    const currentCountElement = u(currentCountElementSelector, element);
+    const totalCountElement = u(totalCountElementSelector, element);
 
     Object.assign(this, {
       slides,
@@ -32,14 +39,14 @@ export class Slideshow {
     });
 
     this.counter = 0;
-    u(this.totalCountElement).text(this.numSlides);
-    u(this.nextTrigger).handle('click', this.nextSlide.bind(this));
-    u(this.prevTrigger).handle('click', this.prevSlide.bind(this));
-    u(this.slideHolder).on('animationend', this.onAnimationEnd.bind(this));
+    this.totalCountElement.text(this.numSlides);
+    this.nextTrigger.handle('click', this.nextSlide.bind(this));
+    this.prevTrigger.handle('click', this.prevSlide.bind(this));
+    this.slideHolder.on('animationend', this.onAnimationEnd.bind(this));
   }
 
   get numSlides() {
-    return u(this.slides).length;
+    return this.slides.length;
   }
 
   get slideToShow() {
@@ -57,24 +64,20 @@ export class Slideshow {
   }
 
   slide(direction) {
-
-    const uSlideHolder = u(this.slideHolder);
-    const uSlides = u(this.slides);
-
     // Update the navigation with the current slide number
-    u(this.currentCountElement).text(this.slideToShow + 1);
+    this.currentCountElement.text(this.slideToShow + 1);
 
     // Remove previous directional class
-    uSlideHolder.removeClass(this.classIsForward, this.classIsBack);
+    this.slideHolder.removeClass(this.classIsForward, this.classIsBack);
 
     // Remove previous 'was-visible' class
-    uSlides.removeClass(this.classWasVisible);
+    this.slides.removeClass(this.classWasVisible);
 
     // Find slide that has class `is-visible`, and replace it with `was-visible`
-    uSlides.each((slide, i) => {
-      var uSlide = u(slide);
-      if (uSlide.hasClass(this.classIsVisible)) {
-        uSlide.removeClass(this.classIsVisible).addClass(this.classWasVisible);
+    this.slides.each((slide, i) => {
+      var slide = u(slide);
+      if (slide.hasClass(this.classIsVisible)) {
+        slide.removeClass(this.classIsVisible).addClass(this.classWasVisible);
       }
     });
 
@@ -83,14 +86,14 @@ export class Slideshow {
 
     // Add new directional classes to slide container
     if (direction === 'forward') {
-      uSlideHolder.addClass(this.classIsSlidingForward, this.classIsForward);
+      this.slideHolder.addClass(this.classIsSlidingForward, this.classIsForward);
     } else {
-      uSlideHolder.addClass(this.classIsSlidingBack, this.classIsBack);
+      this.slideHolder.addClass(this.classIsSlidingBack, this.classIsBack);
     }
   }
 
   onAnimationEnd() {
-    u(this.slideHolder).removeClass(this.classIsSlidingForward, this.classIsSlidingBack);
+    this.slideHolder.removeClass(this.classIsSlidingForward, this.classIsSlidingBack);
   }
 
   nextSlide() {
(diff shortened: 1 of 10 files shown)


src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,99 @@
...
+
+  get numSlides() {
+    return this.slides.length;
+  }
+
+  slide(direction) {
+
+    // Get the index of the current item
+    if (this.counter > 0) {
+      var slideToShow = (this.counter % this.numSlides);
```
mrgerardorodriguez: Could we move the `slideToShow` variable before the `if`/`else` statement? Something like:

```js
slide(direction) {
  var slideToShow;

  if (this.counter > 0) {
    slideToShow = (this.counter % this.numSlides);
  } else {
    slideToShow = // else code here...
  }
  // rest of code...
}
```

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,99 @@
...
+    } else {
+      var slideToShow = (this.counter % this.numSlides) ? ((this.counter % this.numSlides) + this.numSlides) : 0;
```
mrgerardorodriguez: Could we store the value of `this.counter % this.numSlides` in it's own variable so as not to duplicate it? 😉

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,99 @@
+'use strict';
+import {u} from 'umbrellajs';
+export class Slideshow {
+  constructor (element, {
+    slideHolder = element.querySelector('.js-Slideshow-slides'),
+    slides = element.querySelectorAll('.js-Slideshow-slide'),
+    nextTrigger = element.querySelector('.js-Slideshow-next'),
+    prevTrigger = element.querySelector('.js-Slideshow-prev'),
+    currentCountElement = element.querySelector('.js-Slideshow-current'),
+    totalCountElement = element.querySelector('.js-Slideshow-total'),
```
mrgerardorodriguez: I'm curious what @erikjung thinks about my thoughts below...

I'm wondering, should the querying of the elements happen at this point? Or should this just be the CSS selector?

For example, should it be something like this?
```js
  constructor (element, {
    slideHolderSelector = '.js-Slideshow-slides',
    // rest of selectors...
  } = {}) {
    
    const  slideHolder = u(slideHolderSelector);

    // We could even make sure the slideHolder exists, since we depend on it
    if (!slideHolder) {
      throw new Error(`No element was found using the "${slideHolderSelector}" CSS selector.`);
    }

    // Now that we are sure the slideHolder exists, let's assign it. :) 
    Object.assign(this, {
      slideHolder,
      // more assignments...
    }
```

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,99 @@
...
+      classIsVisible,
+      classWasVisible,
+      classIsForward,
+      classIsBack,
+      classIsSlidingForward,
+      classIsSlidingBack
+    });
+
+    this.counter = 0;
+    u(this.totalCountElement).html(this.numSlides);
```
mrgerardorodriguez: Could this use [`.text()`](http://umbrellajs.com/documentation#text) maybe?

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,102 @@
...
+      var slideToShow = slideRemainder;
+    } else {
+      var slideToShow = slideRemainder ? (slideRemainder + this.numSlides) : 0;
+    }
+
+    // Update the navigation with the current slide number
+    u(this.currentCountElement).text(slideToShow + 1);
+
+    // Remove previous directional class
+    u(this.slideHolder).removeClass(this.classIsForward, this.classIsBack);
```
mrgerardorodriguez: If we do decide to follow the suggestion I made [in this comment](https://github.com/cloudfour/cloudfour.com-patterns/pull/350/files?diff=unified#r72332915), then we can change `u(this.slideHolder).removeClass()` to `this.slideHolder.removeClass()` since `this.slideHolder` will hold an Umbrella JS instance that we can chain methods onto. 😁

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,102 @@
...
+  slide(direction) {
+
+    var slideToShow;
+    var slideRemainder = this.counter % this.numSlides;
+
+    // Get the index of the current item
+    if (this.counter > 0) {
+      var slideToShow = slideRemainder;
+    } else {
+      var slideToShow = slideRemainder ? (slideRemainder + this.numSlides) : 0;
```
mrgerardorodriguez: We can remove the `var` here. 😉

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  slide(direction) {
+
+    var slideToShow;
+    var slideRemainder = this.counter % this.numSlides;
+
+    // Get the index of the current item
+    if (this.counter > 0) {
+      var slideToShow = slideRemainder;
```
mrgerardorodriguez: No `var` needed. :)

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,102 @@
...
+    // Remove previous 'was-visible' class
+    this.slides.forEach(slide => {
+      u(slide).removeClass(this.classWasVisible);
+    });
+
+    // Find slide that has class `is-visible`, and replace it with `was-visible`
+    this.slides.forEach(slide => {
+      if (u(slide).hasClass(this.classIsVisible)) {
+        u(slide).removeClass(this.classIsVisible);
+        u(slide).addClass(this.classWasVisible);
```
mrgerardorodriguez: Could we store the value `u(slide)` in a variable so we don't duplicate/query it 3 times? :)

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,102 @@
...
+    u(this.totalCountElement).text(this.numSlides);
+    u(this.nextTrigger).handle('click', () => this.next());
+    u(this.prevTrigger).handle('click', () => this.prev());
+  }
+
+  get numSlides() {
+    return this.slides.length;
+  }
+
+  slide(direction) {
```
mrgerardorodriguez: This method seems to be tasked with doing many different things. Might we be able to break it into smaller chunks, perhaps?

Maybe you have one smaller method that returns the `slideToShow` as it's only purpose, for example?

src/assets/toolkit/scripts/lib/component/slideshow.js
mrgerardorodriguez: @erikjung Also see [this comment](https://github.com/cloudfour/cloudfour.com-patterns/pull/350/files#r72335530).

src/assets/toolkit/scripts/lib/component/slideshow.js
saralohr: Not sure... I could chain them to reduce a little?
`u(slide).removeClass(this.classIsVisible).addClass(this.classWasVisible);`

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
...
+    // Remove previous 'was-visible' class
+    this.slides.forEach(slide => {
+      u(slide).removeClass(this.classWasVisible);
+    });
+
+    // Find slide that has class `is-visible`, and replace it with `was-visible`
+    this.slides.forEach(slide => {
+      if (u(slide).hasClass(this.classIsVisible)) {
+        u(slide).removeClass(this.classIsVisible).addClass(this.classWasVisible);
+      }
```
mrgerardorodriguez: Sorry, I wasn't very clear. :) I was thinking something like this:
```js
this.slides.forEach(slide => {
  const uSlide = u(slide);
  if (uSlide.hasClass(this.classIsVisible)) {
    uSlide.removeClass(...).addClass(...);
  }
}
```
You don't have to use `uSlide` as the name, just an example. 😉

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
...
+
+    // Remove animation class so it can be re-applied as a new animation
+    var onAnimationEnd = () => {
+      u(this.slideHolder).removeClass(this.classIsSlidingForward, this.classIsSlidingBack);
+    }
+
+    u(this.slideHolder).on("animationend", onAnimationEnd, false);
+  }
+
+  next() {
```
mrgerardorodriguez: Nit: Could we rename this method to something a bit more descriptive? Perhaps `nextSlide()`? 😁

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
...
+    this.counter++;
+    this.slide('forward');
+  }
+
+  prev() {
```
mrgerardorodriguez: Nit: If [this is is done](https://github.com/cloudfour/cloudfour.com-patterns/pull/350/files#r72340627), maybe we could rename this to `prevSlide()`? Take it or leave it, I won't be offended. LOL

src/pages/pages/team-single.hbs
```diff
@@ -59,49 +59,8 @@ stylesheets:
...
-          <svg viewBox="0 0 24 24" class="Icon u-textGrow3" role="presentation">
-            <circle r="12" cx="12" cy="12" fill="#fff"/>
-            {{svg "icons/arrow-right" width="12" height="12" x="6" y="6"}}
-          </svg>
-        </a>
-      </li>
-    </ul>
-  </aside>
+  {{#embed "patterns.components.slideshow.base"}}
+  {{/embed}}
```
mrgerardorodriguez: I am not 100% sure about this, but I thought I saw @tylersticka using a different syntax that allowed us to avoid having to use the closing `{{/embed}}` part. Does that ring a bell at all @tylersticka? I _totally_ might be making things up. 😅

saralohr: Oh...since I am not passing anything, I could just do
```{{> "patterns.components.slideshow.base" }}```

File: src/assets/toolkit/scripts/lib/component/slideshow.js [hunk 14]
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    // Add `is-visible` class to current item
+    u(this.slides[this.slideToShow]).addClass(this.classIsVisible);
+
+    // Add new directional classes to slide container
+    if (direction == 'forward') {
+      u(this.slideHolder).addClass(this.classIsSlidingForward, this.classIsForward);
+    } else {
+      u(this.slideHolder).addClass(this.classIsSlidingBack, this.classIsBack);
```
mrgerardorodriguez: I see `u(this.slideHolder)` used at least 5 times in the `slide()` method. Perhaps we can store this in a variable within the `slide()` method so we don't have to query for it each time?

Something like:
```js
slide(direction) {
  const uSlideHolder = u(this.slideHolder);

  // code...
}
```

src/assets/toolkit/scripts/lib/component/slideshow.js
erikjung: @mrgerardorodriguez @saralohr 

In general, I'd say it's a good idea to avoid assuming the existence of elements, but I'm also aware that we might be guilty of this pattern in other modules.

Also, it does appear as though there are some excessive calls to `element.querySelectorAll()` in each case where we subsequently wrap the elements with Umbrella's `u()` function.

If we're going to use Umbrella in a module, then we probably don't need `element.querySelectorAll()`.

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
...
+  slide(direction) {
+
+    // Update the navigation with the current slide number
+    u(this.currentCountElement).text(this.slideToShow + 1);
+
+    // Remove previous directional class
+    u(this.slideHolder).removeClass(this.classIsForward, this.classIsBack);
+
+    // Remove previous 'was-visible' class
+    this.slides.forEach(slide => {
```
erikjung: Does Umbrella allow jQuery-style manipulation of en entire collection at once?

If so, could this be simplified to the following?

```js
this.slides.removeClass(this.classWasVisible);
```

src/assets/toolkit/scripts/lib/component/slideshow.js
mrgerardorodriguez: > Also, it does appear as though there are some excessive calls to element.querySelectorAll() in each case where we subsequently wrap the elements with Umbrella's u() function.

Ah, yes, great point @erikjung.

Related: I was wondering...when using jQuery, a pattern arised to use the `$` as a variable name prefix to denote if the element was a jQuery wrapped object.

Might we want to follow a similar pattern here? Perhaps adding `u` as a prefix?

For example:
```js
const uSlideHolder = u(slideHolderSelector);
```

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
...
+      if (u(slide).hasClass(this.classIsVisible)) {
+        u(slide).removeClass(this.classIsVisible).addClass(this.classWasVisible);
+      }
+    });
+
+    // Add `is-visible` class to current item
+    u(this.slides[this.slideToShow]).addClass(this.classIsVisible);
+
+    // Add new directional classes to slide container
+    if (direction == 'forward') {
```
erikjung: Nit: `===`

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+      u(this.slideHolder).addClass(this.classIsSlidingBack, this.classIsBack);
+    }
+
+    // Remove animation class so it can be re-applied as a new animation
+    var onAnimationEnd = () => {
+      u(this.slideHolder).removeClass(this.classIsSlidingForward, this.classIsSlidingBack);
+    }
+
+    u(this.slideHolder).on("animationend", onAnimationEnd, false);
```
erikjung: Nit: single quotes.

src/assets/toolkit/scripts/lib/component/slideshow.js
saralohr: Good call. It appears to... I just tried this and it works.

src/assets/toolkit/styles/utils/display.css
```diff
@@ -60,4 +68,16 @@
   .u-lg-block {
     display: block !important;
   }
+
+  .u-lg-inlineBlock {
+    display: inline-block !important;
+  }
+
+  .u-lg-unhideVisually {
```
erikjung: Is this only needed in order to override another responsive utility that applies to narrow breakpoints?

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,106 @@
...
+
+    // Add new directional classes to slide container
+    if (direction == 'forward') {
+      u(this.slideHolder).addClass(this.classIsSlidingForward, this.classIsForward);
+    } else {
+      u(this.slideHolder).addClass(this.classIsSlidingBack, this.classIsBack);
+    }
+
+    // Remove animation class so it can be re-applied as a new animation
+    var onAnimationEnd = () => {
```
mrgerardorodriguez: Should this be a method of the `SlideShow` class?

src/pages/pages/team-single.hbs
erikjung: We can use this when no partial block body is needed:

```hbs
{{{embed ""}}}
```

mrgerardorodriguez: Thanks @erikjung! That's what I had in mind, but couldn't remember.

Now I ask, should we use the `{{{embed ""}}}` or the Handlebars partial method that @saralohr noted? I...am not sure.

src/assets/toolkit/scripts/lib/component/slideshow.js
erikjung: @mrgerardorodriguez 

> Related: I was wondering...when using jQuery, a pattern arised to use the $ as a variable name prefix to denote if the element was a jQuery wrapped object.
>
> Might we want to follow a similar pattern here? Perhaps adding u as a prefix?

I don't have strong feelings about this. Honestly, I was never a fan of the `$` prefix, but adhered to it because it's such a widely accepted convention. As long as we're consistent. If we've already done this in a handful of other files, then I say do it. If this is the first file that uses the convention, then we should update all of the others using Umbrella to match it.

src/assets/toolkit/styles/utils/display.css
saralohr: Yep, exactly. The slide count is visually hidden at small screens, but I want to show it at large screens.

src/assets/toolkit/scripts/lib/component/slideshow.js
mrgerardorodriguez: Learning question: I don't understand what the `false` is for here:
```js
u(this.slideHolder).on('animationend', onAnimationEnd, false);
```
Can someone smarter than me explain? 😊

src/pages/pages/team-single.hbs
erikjung: @mrgerardorodriguez @saralohr 

> Now I ask, should we use the {{{embed ""}}} or the Handlebars partial method that @saralohr noted? I...am not sure.

I've asked myself that same question many times, and I always decided to use `embed` just to be consistent.

src/assets/toolkit/scripts/lib/component/slideshow.js
mrgerardorodriguez: Something like:
```js
export class Slideshow {
  slide() {
    u(this.slideHolder).on('animationend', this.onAnimationEnd.bind(this), false);
  }

  onAnimationEnd() {
   // code here...
  }
}
```

src/assets/toolkit/styles/utils/display.css
tylersticka: Any reason not to call it `showVisually` instead of `unhide`?

Also, you might dig [the `initial` keyword](http://cssnext.io/features/#initial-value) for this.

src/assets/toolkit/styles/utils/text.css
```diff
@@ -1,6 +1,16 @@
 @import "suitcss-utils-text";
 
 /**
+ * Make responsive SUIT CSS text-alignment utilities
+ */
+
+@media(--sm-viewport) {
+  .u-sm-textLeft {
```
tylersticka: This goes for all new utilities, but make sure they're added to the documentation prior to any merge.

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,104 @@
...
+      u(this.slideHolder).removeClass(this.classIsSlidingForward, this.classIsSlidingBack);
+    }
+
+    u(this.slideHolder).on('animationend', onAnimationEnd, false);
```
mrgerardorodriguez: Let's move this out of the `slide()` method and into the `constructor()` method. We don't need to bind to the `animationend` event every time a slide occurs, I'm pretty sure.

src/assets/toolkit/scripts/lib/component/slideshow.js
saralohr: Oh, I believe this is not necessary here. It is for whether `event.default()` should be called, I believe, from when I was using `addEventListener` at one point, I think.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

src/patterns/components/slideshow/base.hbs
```diff
@@ -0,0 +1,45 @@
+<div class="u-bgGray u-posRelative Slideshow js-Slideshow u-pullSides1">
+  <h2 class="u-hiddenVisually">Fun facts about Aileen</h2>
+  <ul class="u-flex u-listUnstyled Slideshow-slides js-Slideshow-slides u-padEnds2"  id="Slideshow-slides">
+    {{#each drizzle.data.fun-facts.contents.aileen}}
+      <li class="Grid Grid--withGutter Grid--alignMiddle Slideshow-slide js-Slideshow-slide u-flexExpand u-flexJustifyCenter u-sm-flexJustifyStart u-padSides5 {{#if @first}}is-visible{{/if}}">
```
tylersticka: Because this content is already structured with headings, the list seems possibly unnecessary? It's not _wrong_, but I'm also not sure it lends this area much semantic value?

src/patterns/components/slideshow/base.hbs
```diff
@@ -0,0 +1,45 @@
...
+        <div class="Grid-cell u-sm-size3of4 u-lg-size1of2 u-spaceItems06 u-textCenter u-sm-textLeft u-spaceTop01 u-sm-spaceTopNone">
+          <h4 class="u-md-textGrow1">Did you know&hellip;?</h4>
+          <p>{{fact}}</p>
+        </div>
+      </li>
+    {{/each}}
+  </ul>
+  <ul class="Slideshow-nav u-listInline u-spaceSidesNone u-flexJustifyBetween u-flexAlignContentCenter u-lg-flexJustifyEnd u-lg-padRight3 u-posAbsoluteCenter u-size1of1" aria-controls="#Slideshow-slides">
+    <li class="u-spaceLeft05">
+      <a class="u-linkClean js-Slideshow-prev" href="">
```
tylersticka: Is the idea that these links will resolve to the ID of the next and previous, or to some server-side implementation? If not, `<button type="button">` might be a better choice.

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,104 @@
...
+      classIsForward,
+      classIsBack,
+      classIsSlidingForward,
+      classIsSlidingBack
+    });
+
+    this.counter = 0;
+    u(this.totalCountElement).text(this.numSlides);
+    u(this.nextTrigger).handle('click', () => this.nextSlide());
+    u(this.prevTrigger).handle('click', () => this.prevSlide());
```
mrgerardorodriguez: Let's remove the anonymous function and use `.bind()`, like this:
```js
u(this.nextTrigger).handle('click', this.nextSlide.bind(this));
u(this.prevTrigger).handle('click', this.prevSlide.bind(this));
```

mrgerardorodriguez:
Great work @saralohr! I left some inline comments. Let me know when they've been addressed and I can take another look. 😉

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,104 @@
...
+    // Remove previous 'was-visible' class
+    u(this.slides).removeClass(this.classWasVisible);
+
+    // Find slide that has class `is-visible`, and replace it with `was-visible`
+    this.slides.forEach(slide => {
+      var slide = u(slide);
+      if (slide.hasClass(this.classIsVisible)) {
+        slide.removeClass(this.classIsVisible).addClass(this.classWasVisible);
+      }
+    });
```
saralohr: Regarding this comment:
> If we're going to use Umbrella in a module, then we probably don't need `element.querySelectorAll()`

I'm using `querySelectorAll()` for `this.slides`, and I wanted to switch this up and use `u` in the function above, so that I could use the Umbrella `each()` method, but for some reason I haven't had any luck making the [`each()`](http://umbrellajs.com/documentation#each) function work as expected. 

For instance, this gives a console error `Uncaught TypeError: Cannot read property 'toString' of undefined`:

```
u(this.slides).each(function(slide, i) {
  var slide = u(slide);
  if (slide.hasClass(this.classIsVisible)) {
    slide.removeClass(this.classIsVisible).addClass(this.classWasVisible);
  }
});
```

src/assets/toolkit/scripts/lib/component/slideshow.js
saralohr: Hrm, this didn't work. I don't know enough about `.bind()` to quite understand why, though.

src/assets/toolkit/scripts/lib/component/slideshow.js
mrgerardorodriguez: @saralohr I've yet to understand why it works, but can you give this ES2015 syntax a try?
```js
u(this.slides).each((slide, i) => {
  var slide = u(slide);
  if (slide.hasClass(this.classIsVisible)) {
    slide.removeClass(this.classIsVisible).addClass(this.classWasVisible);
  }
});
```
Did it work for you?

src/patterns/components/slideshow/base.hbs
saralohr: I think it could go either way. The one benefit it provides is the screenreader says how many items are in the section when you enter it. And, it _is_ a list of things, so I am inclined to keep it.

src/patterns/components/slideshow/base.hbs
saralohr: > Is the idea that these links will resolve to the ID of the next and previous, or to some server-side implementation?

Nope, I meant to swap this to a button, but forgot. I think I initially used a link since it has all the color/hover that I want, and then never went back to it. You're right that `button` is a better choice.

src/assets/toolkit/scripts/lib/component/slideshow.js
mrgerardorodriguez: I'm pretty sure it has something to do with scope and what `this` refers to. Still digging, though.

tylersticka: That's exactly why it works, @mrgerardorodriguez!

When you create an anonymous function the old-fashioned way (`function(){}`), the keyword `this` refers to _the function_.

In old-school JS, you had to do some strange stuff (or use jQuery's proxy method) to force `this` to stay the same.

But in ES6, arrow functions _retain_ that context. Which is awesome.

mrgerardorodriguez: @tylersticka LOL, that's what I figured. I was trying to get it to work with the old-school JS syntax, but couldn't figure out the "strange stuff". I was curious just cause I wanted to remind myself how to get that to work. 😁

src/assets/toolkit/scripts/lib/component/slideshow.js
mrgerardorodriguez: @saralohr Really? Darn! Let me know if you want me to take a closer (at your desk) look. 😉

saralohr: Hmm, it's working now. I must have had another breaking change in place when I added that.

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
+'use strict';
+import {u} from 'umbrellajs';
+export class Slideshow {
+  constructor (element, {
+    slideHolder = u('.js-Slideshow-slides'),
+    slides = u('.js-Slideshow-slide'),
+    nextTrigger = u('.js-Slideshow-next'),
+    prevTrigger = u('.js-Slideshow-prev'),
+    currentCountElement = u('.js-Slideshow-current'),
+    totalCountElement = u('.js-Slideshow-total'),
```
saralohr: Is there something additional that needs to be done to keep these within the `element` scope, or does that happen automagically? Like, previously they were `element.querySelector()` vs `document.querySelector()`.

tylersticka: I believe to do what you were doing before in Umbrella land, you'd need to first run `element` through Umbrella (`u(element)`) and then use the `find` modifier (similar to jQuery). So altogether: `u(element).find('.js-Slideshow-slides')`

mrgerardorodriguez: Yep! That sounds right @tylersticka. You could also pass a 2nd argument as an alternative, something like this, I believe:
```js
const slides = u('.js-Slideshow-slides', element);
```

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
+'use strict';
+import {u} from 'umbrellajs';
+export class Slideshow {
+  constructor (element, {
+    slideHolder = u('.js-Slideshow-slides'),
```
mrgerardorodriguez: I still would like to see this follow the suggestion [made here](https://github.com/cloudfour/cloudfour.com-patterns/pull/350#discussion_r72332915). We could also refer to [`comment-reply.js`](https://github.com/cloudfour/cloudfour.com-patterns/blob/master/src/assets/toolkit/scripts/lib/component/comment-reply.js) as it does something similar in that these are only string CSS selectors (and not Umbrella objects). They then get wrapped as Umbrella objects [within the `constructor()` method](https://github.com/cloudfour/cloudfour.com-patterns/blob/master/src/assets/toolkit/scripts/lib/component/comment-reply.js#L31).

Let me know if this isn't making sense, though. 😉

saralohr: Thanks, @mrgerardorodriguez, will do. I think that makes sense.

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
...
+      slideToShow = slideRemainder;
+    } else {
+      slideToShow = slideRemainder ? (slideRemainder + this.numSlides) : 0;
+    }
+
+    return slideToShow;
+  }
+
+  slide(direction) {
+
```
mrgerardorodriguez: Nit: Extra space.

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
...
+    // Add new directional classes to slide container
+    if (direction === 'forward') {
+      uSlideHolder.addClass(this.classIsSlidingForward, this.classIsForward);
+    } else {
+      uSlideHolder.addClass(this.classIsSlidingBack, this.classIsBack);
+    }
+  }
+
+  onAnimationEnd() {
+    u(this.slideHolder).removeClass(this.classIsSlidingForward, this.classIsSlidingBack);
```
mrgerardorodriguez: You don't need to re-wrap `this.slideHolder` in a second Umbrella object since you've already wrapped it in the `constructor()`. This should work just fine: 😉 
```js
this.slideHolder.removeClass(this.classIsSlidingForward, this.classIsSlidingBack);
```

saralohr: Thanks @mrgerardorodriguez. I'm currently updating all of the elements to be defined in the constructor with Umbrella, and removing the `u()` from other instances.

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,105 @@
...
+
+    this.counter = 0;
+    u(this.totalCountElement).text(this.numSlides);
+    u(this.nextTrigger).handle('click', this.nextSlide.bind(this));
+    u(this.prevTrigger).handle('click', this.prevSlide.bind(this));
+    u(this.slideHolder).on('animationend', this.onAnimationEnd.bind(this));
+  }
+
+  get numSlides() {
+    return u(this.slides).length;
```
mrgerardorodriguez: Same here, see [this comment](https://github.com/cloudfour/cloudfour.com-patterns/pull/350/files#r72480142), this should work: 😊 
```js
return this.slides.length;
```

src/assets/toolkit/scripts/lib/component/slideshow.js
mrgerardorodriguez: Cool, thanks @saralohr! 😄

saralohr:
@mrgerardorodriguez Alrighty, ready for re-review.

src/assets/toolkit/scripts/lib/component/slideshow.js
```diff
@@ -0,0 +1,108 @@
...
+    classIsVisible = 'is-visible',
+    classWasVisible = 'was-visible',
+    classIsSlidingForward = 'is-sliding-forward',
+    classIsSlidingBack = 'is-sliding-back',
+    classIsForward = 'is-forward',
+    classIsBack = 'is-back'
+  } = {}) {
+
+    const slideHolder = u(slideHolderSelector, element);
+    const slides = u(slidesSelector, element);
```
mrgerardorodriguez: _Super_ picky...could we order these in the same order as the `Object.assign()`? Feel free to ignore this comment if you'd like. 😅  🙈 

```js
const slides = // code...
const slideHolder = // code...
```

mrgerardorodriguez:
The JS looks awesome @saralohr!!! Great work! 👏  😄  

Might it be possible to add minimal documentation? Referring to [comment-reply.js](https://github.com/cloudfour/cloudfour.com-patterns/blob/master/src/assets/toolkit/scripts/lib/component/comment-reply.js) or [sky.js](https://github.com/cloudfour/cloudfour.com-patterns/blob/master/src/assets/toolkit/scripts/lib/component/sky.js) might be helpful.

Other than that, LGTM! 👍  😊

PR #356, opened 2016-07-29
Prevent animation from starting in Opera Mini

### LINKED ISSUE #351: Opera Mini not showing fallback state for home animation
The home animation SVG has three states...

1. **If SVG is supported**, a basic (un-animated) SVG displays.
2. **If JavaScript is also supported**, the basic SVG is altered a bit and the animation begins.
3. **If SVG is unsupported**, a PNG fallback is displayed instead.

In either the first or third case, the image should look like this:

home-animation-fallback

This isn't what's happening in Opera Mini. If I had to guess, I'd say the browser is _starting_ the animation, but only showing the first frame? But that might be totally wrong:

img_1159

## Reminder: How to test Opera Mini on iOS

1. [Install the app](https://itunes.apple.com/us/app/opera-mini-web-browser/id363729560?mt=8)
2. Open the app
3. Tap the red "Opera" icon at the bottom of the UI
4. Tap the area with the "savings" graph
5. Select "Opera Mini"
6. Tap "Done"

tylersticka:
This prevents the homepage animation from starting in Opera Mini, which shows the nicer fallback experience instead of a single frame of the animation.

I first attempted to pull this off by deferring the animation start time somehow, but to no avail. I resorted to a browser check [after reading Opera Mini's JS docs](https://dev.opera.com/articles/opera-mini-and-javascript/) and understanding a bit more about what it was doing and why.

The diff for the home animation JS is kinda bonkers because I moved everything into a conditional. Sorry about that. 😅 

## Before

img_1159

## After

img_1168

---

@saralohr @erikjung @mrgerardorodriguez 

Fixes #351


mrgerardorodriguez:
LGTM 👍

src/assets/toolkit/scripts/lib/tests/opera-mini.js
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * 99% of the time, you should absolutely use feature detection over browser
+ * detection. But Opera Mini is kind of a different beast, and for things like
+ * animation it can be helpful/necessary to detect it specifically.
+ *
+ * @see https://dev.opera.com/articles/opera-mini-and-javascript/
+ */
+
+export default function operaMini () {
+  return Object.prototype.toString.call(window.operamini) === "[object OperaMini]";
```
erikjung: Very curious why all of the examples suggest this verbose statement.

Looks like `window.operamini` is not a boolean, but a pointer to an `OperaMini` object?

mrgerardorodriguez: Oh, good catch @erikjung. What happens if you don't `toString` it?

erikjung:
👍

src/assets/toolkit/scripts/lib/tests/opera-mini.js
tylersticka: I copied and pasted this directly from their docs: https://dev.opera.com/articles/opera-mini-and-javascript/#detecting-opera-mini-using-the-windowoperamini-property

I do not know the answers to your questions.

tylersticka:
@erikjung @mrgerardorodriguez Is the discussion [here](https://github.com/cloudfour/cloudfour.com-patterns/pull/356#discussion_r72869730) simply a point of curiosity, or should I hold on merging until it's resolved? Happy to address feedback but not sure if I have action items currently or not.

mrgerardorodriguez:
Curiousity on my part.

Issue #346, opened 2016-07-25
Move Jetpack Contact Form styles to vendor stylesheet, make DRY


tylersticka:
Currently there are styles customizing the appearance of Jetpack contact forms, but these styles are descendents of `.CmsContent`. While this was my direction initially, I realize now this would be better suited for a vendor stylesheet a la Jetpack Markdown.

Why? So we can leverage mixins to keep things DRY between repetitive styles without worrying about source order (since `vendor` stuff gets included last).

PR #357, opened 2016-07-29
Fix issue 346

saralohr:
Use mixins for the Jetpack form styles to keep things DRY. Fixes issue #346.

screen shot 2016-07-29 at 4 47 56 pm


tylersticka:
I haven't tried locally, but the code's exactly what I envisioned! Excellent job! :+1:

PR #360, opened 2016-08-04
Add u-scrollX utility, simplify Table component

### LINKED ISSUE #359: Table cells do not fill width of table
To support horizontal scrolling our `Table` element has `display: block` applied to it. This is causing the table cells to not fill the full width of the table. Find an alternate solution for horizontal scroll, or perhaps apply these styles via a modifier so it can be added as necessary.

tylersticka:
Our technique for overflow scrolling of tables meant tables no longer honored `width: 100%`.

As @saralohr mentioned in her issue (#359), we really only had two options... a container with `overflow-x: auto;` or some sort of modifier on the element itself.

I started by creating a modifier, but the point at which overflow-scrolling should apply seemed entirely dependent on the content. The only good way to do it would be to create breakpoint-specific iterations of the modifier, which seemed like overkill when a single, wrapping utility would serve the same purpose.

u-scrollx

---

@saralohr @erikjung @mrgerardorodriguez 

Fixes #359

### DIFF
--- a/src/assets/toolkit/styles/components/table.css
+++ b/src/assets/toolkit/styles/components/table.css
@@ -10,8 +10,6 @@
 
 .Table {
   width: 100%;
-  display: block;
-  overflow-x: auto;
 }
 
 .Table :matches(th, td) {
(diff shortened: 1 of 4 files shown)


saralohr:
Nice! 👍

PR #361, opened 2016-09-01
Fix attachment icon

tylersticka:
This PR removes a duplicate `width` and `height` attribute from the attachment icon, which was making this SVG invalid.

It also adds `.svgsus`' icon set preference file to the `.gitignore`, because that's how I found this issue.

---

@saralohr @nicolemors @erikjung @gerardo-rodriguez

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 dist
 node_modules
 npm-debug.log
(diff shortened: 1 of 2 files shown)


saralohr:
👍

PR #365, opened 2016-10-04
Offline notification

aileenjeffries:
A little pattern for an offline notification. I developed a base "Message" pattern in case we ever want to add different types of messages (alerts, success, etc) and extended it for the offline version of this pattern. 

I need some JS help getting the close link and "try again" link to do anything. 

cc @erikjung and @tylersticka


src/assets/toolkit/styles/components/message.css
```diff
@@ -0,0 +1,38 @@
+/** @define Message */
+
+:root {
+  --Message-padding: var(--space-xs) 2.5em;
+  --Message-background: var(--color-gray);
+  --Message-offline-background: var(--color-orange);
+  --Message-offline-color: var(--color-white);
+  --Message-close-size: 1.4em;
```
tylersticka: Did you try [our modular sizes](https://github.com/cloudfour/cloudfour.com-patterns/blob/master/src/assets/toolkit/styles/base/_config.css#L109-L113) and find that none of them worked?

src/assets/toolkit/styles/components/message.css
```diff
@@ -0,0 +1,38 @@
+/** @define Message */
+
+:root {
+  --Message-padding: var(--space-xs) 2.5em;
```
tylersticka: None of our [modular sizes](https://github.com/cloudfour/cloudfour.com-patterns/blob/master/src/assets/toolkit/styles/base/_config.css#L98-L101) worked for this?

src/assets/toolkit/styles/components/message.css
```diff
@@ -0,0 +1,38 @@
...
+  --Message-background: var(--color-gray);
+  --Message-offline-background: var(--color-orange);
+  --Message-offline-color: var(--color-white);
```
tylersticka: Did you test the contrast of the white on orange using [tota11y](http://khan.github.io/tota11y/) or some other tool?

src/assets/toolkit/styles/components/message.css
```diff
@@ -0,0 +1,38 @@
...
+
+.Message-close {
+  position: absolute;
+  display: block;
+  top: var(--space-xs);
+  right: var(--space-xs);
+}
+
+.Message-close svg {
+  transform: rotate(45deg); /* turns a plus sign into an X */
```
tylersticka: It would be cleaner to create a new `x.svg` icon with the transform attribute I mentioned in Slack.

It's source should be this:

```html
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24">
  <g transform="rotate(45 12 12)" fill="none" stroke="currentColor" stroke-width="4" stroke-linecap="round">
    <line x1="12" y1="3" x2="12" y2="21"/>
    <line x1="3" y1="12" x2="21" y2="12"/>
  </g>
</svg>
```

Then you can just update the template.

src/assets/toolkit/styles/components/message.css
tylersticka: Also, there is an `.Icon--large` modifier that increases the size by `1.5em`. [More info](http://cloudfour-patterns.netlify.com/patterns/components/icon.html#size-modifiers)

src/patterns/components/offline-notification/example.hbs
```diff
@@ -0,0 +1,12 @@
+---
+name: Example
+sourceless: true
+hidden: true
+---
+
+<div class="Message Message--offline">
```
tylersticka: Please create two separate patterns, one to show the _default_ look and feel and one for the offline appearance.

src/patterns/components/offline-notification/example.hbs
```diff
@@ -0,0 +1,12 @@
...
+  You seem to be offline at the moment. <span class="u-hidden u-md-inlineBlock">We'll keep trying to connect.</span> <a href=".">Try again</a>
```
tylersticka: Nit: Use smart apostrophes for nicer typography (`We&rsquo;ll`)

src/patterns/components/offline-notification/example.hbs
```diff
@@ -0,0 +1,12 @@
...
+  <a href="#" class="Message-close">
```
tylersticka: This should be a `<button>` element since it does not actually resolve to a destination or trigger a reload of the current destination.

src/patterns/components/offline-notification/overview.hbs
```diff
@@ -0,0 +1,7 @@
+---
+order: 10
+notes: |
+  In page notification about lack of internet connectivity.
+---
+
+{{{embed "patterns.components.offline-notification.example"}}}
```
tylersticka: Not sure what the purpose is of this pattern that only includes the other pattern?

src/assets/toolkit/styles/components/message.css
```diff
@@ -0,0 +1,38 @@
+/** @define Message */
+
+:root {
+  --Message-padding: var(--space-xs) 2.5em;
+  --Message-background: var(--color-gray);
+  --Message-offline-background: var(--color-orange);
```
erikjung: Can we use something more general for this state name? Like "warn"? 

Unless I'm misunderstanding, this _could_ be a component used for something else, right?

src/assets/toolkit/styles/components/message.css
aileenjeffries: No. The largest still didn't provide enough room for the close X. I actually find our modular sizes to not be a wide enough range - xs is too big for a lot of things and ls is too small. Or maybe I'm not using them correctly.

src/assets/toolkit/styles/components/message.css
aileenjeffries: I just tested it and it fails for smaller text and passes for larger text. At this point, I'd like to point out I didn't pick this color (although of course I love it) and white text provides the greatest possible amount of contrast. :) Orange says "warning" to me, so it seems like a natural fit for this pattern.

src/assets/toolkit/styles/components/message.css
aileenjeffries: `.Icon--large` works just fine.

src/assets/toolkit/styles/components/message.css
aileenjeffries: OK!

src/patterns/components/offline-notification/overview.hbs
aileenjeffries: I just copied how pagination is set up. I'm guessing it is done this way so the example can be embedded in other pages, which we'll probably want to do for this one eventually.

tylersticka: Pagination is set up in that way because we wanted to show it with a "current page" of `2` so both arrows were visible, but it would be kind of a pain to include a template that always defaulted to `2` instead of `1` as the default page.

Any pattern is embeddable, so `{{{embed "patterns.components.offline-notification.overview"}}}` would work just fine.

src/assets/toolkit/styles/components/message.css
tylersticka: > I just tested it and it fails for smaller text and passes for larger text.

Gotcha. When you say smaller text, do you mean when this pattern is shown at smaller sizes?

If it's just _barely_ below passing, then I'm okay proceeding. If we're failing pretty hardcore, though, then we're doing a disservice to folks with visual impairments.

src/assets/toolkit/styles/components/message.css
aileenjeffries: I did this a little differently, and used 2 * `space-md` to get the space I needed. Hopefully by using a multiple of one of our standard sizes, I'm being more modular.

src/assets/toolkit/styles/components/message.css
aileenjeffries: It is just barely failing. We also use this color combo on buttons, so if it is a big concern, we'll have to reconsider a few things, maybe even making `--color-orange` darker.

src/assets/toolkit/styles/components/message.css
tylersticka: What about `calc(1em * --ms4)`? That's ~ 2.4em and uses our modular scale.

I agree that the `--space-*` variables aren't a wide enough range. We should switch to the same numeric system as our scale for consistency. It goes from `--ms-6` to `--ms6`.

src/patterns/components/offline-notification/overview.hbs
aileenjeffries: Makes sense. I reorganized things. I also renamed this whole pattern to message, with the offline message as an example of a warning message, since that's really more accurate.

src/assets/toolkit/styles/components/message.css
tylersticka: If it's just barely failing, I'm not concerned. Human beings don't conform to a concrete numeric scale... I think it's fine to give ourselves a little wiggle room.

(Though if you ever have suggested color adjustments to the palette that will make them more attractive _and_ readable, I encourage those proposed edits site-wide!)

aileenjeffries:
I believe I've addressed all points of feedback. I also renamed the pattern to Message since that's really what it is. The offline notification won't be a pattern, it will just use the warning message pattern. 

I need some JS help to get the "close" button working, and then may need to add some css to get the close transition looking good. But, that's for another PR probs. 

possible to get a re-review, @tylersticka ?

src/assets/toolkit/styles/components/message.css
```diff
@@ -0,0 +1,34 @@
...
+  --Message-padding-right: calc(var(--space-md) * 2);
+  --Message-background: var(--color-gray);
+  --Message-warning-background: var(--color-orange);
+  --Message-warning-color: var(--color-white);
+}
+
+.Message {
+  position: relative;
+  padding: var(--Message-padding);
+  padding-right: var(--Message-padding-right); /* room for close button if present */
```
tylersticka: Please use our [standard comment structure](https://github.com/cloudfour/guides/tree/master/css#comments).

src/assets/toolkit/styles/components/message.css
```diff
@@ -0,0 +1,34 @@
+/** @define Message */
+
+:root {
+  --Message-padding: var(--space-xs);
+  --Message-padding-right: calc(var(--space-md) * 2);
```
tylersticka: I would recommend `calc(1rem * var(--ms4))` instead.

src/assets/toolkit/styles/components/message.css
tylersticka: Also, you may not need this value at all [if you use Flexbox](http://codepen.io/tylersticka/pen/0d624f543bf0653c3e9cab12d525d7d5).

src/assets/toolkit/styles/components/message.css
```diff
@@ -0,0 +1,34 @@
...
+  padding: var(--Message-padding);
+  padding-right: var(--Message-padding-right); /* room for close button if present */
+  background: var(--Message-background);
+}
+
+.Message-close {
+  position: absolute;
+  display: block;
+  top: var(--Message-padding);
+  right: var(--Message-padding);
```
tylersticka: Per my other comment, positioning for this may be simpler if [you use Flexbox](http://codepen.io/tylersticka/pen/0d624f543bf0653c3e9cab12d525d7d5).

src/assets/toolkit/styles/components/message.css
```diff
@@ -0,0 +1,34 @@
...
+  cursor: pointer;
+}
+
+.Message--warning {
+  background: var(--Message-warning-background);
+  color: var(--Message-warning-color);
+}
+
+.Message--warning a {
+  color: var(--Message-warning-color);
```
tylersticka: Please use `inherit` instead.

src/data/icons.yml
```diff
@@ -6,6 +6,7 @@ mono:
   "attachment": "icons/attachment"
   "check": "icons/check"
   "cloud-four": "icons/cloud-four"
+  "close": "icons/x"
```
tylersticka: Please use the same key here as the key you would need to reference to use the icon (in this case, `x`). Or change the filename to match. That way we can look at the docs to know which key to use.

src/patterns/components/message/overview.hbs
```diff
@@ -0,0 +1,10 @@
+---
+order: 1
+notes: |
+  Message pattern
+---
+
+<div class="Message">
+ Hi, I'm a message!
```
tylersticka: Should this be a paragraph semantically?

Also, I'd recommend using our typical pattern structure to extend this for `warning`. For example:

```hbs
<div class="Message {{messageClass}}">
  {{#block "content"}}
    <p>Hi, I'm a message!</p>
  {{/block}}
  {{#if includeClose}}
    <button class="Message-close">
      {{#svg "icon/x" aria-labelledby=(defaultTo closeIconTitle "Message-close-title")}}
        <title id="Message-close-title">Close</title>
      {{/svg}}
    </button>
  {{/if}}
</div>
```

See my comment on the warning template for how this would make future maintenance a bit easier.

src/patterns/components/message/warning.hbs
```diff
@@ -0,0 +1,12 @@
...
+name: Warning Message
+sourceless: true
+---
+
+<div class="Message Message--warning">
+  You seem to be offline at the moment. <span class="u-hidden u-md-inlineBlock">We&rsquo;ll keep trying to connect.</span> <a href=".">Try again</a>
+  <button class="Message-close">
+    {{svg "icons/x" class="Icon Icon--large"}}
+  </button>
+</div>
```
tylersticka: If we edit our base message template per my earlier comment, this could be rewritten as:

```hbs
{{#embed "patterns.components.message.overview" 
  messageClass="Message--warning"
  includeClose=true}}
  {{#content "content"}}
    <p>You seem to be offline at the moment. <span class="u-hidden u-md-inlineBlock">We&rsquo;ll keep trying to connect.</span> <a href=".">Try again</a></p>
  {{/content}}
{{/embed}}
```

Now if we need to modify the markup structure, the close button semantics, or anything like that, it'll trickle down to every instance.

src/patterns/components/message/with-close.hbs
```diff
@@ -0,0 +1,12 @@
...
+name: Message with Close Button
+sourceless: true
+---
+
+<div class="Message">
+  I'm a message with a close button!
+  <button class="Message-close">
+    {{svg "icons/x" class="Icon Icon--large"}}
+  </button>
+</div>
```
tylersticka: See earlier comments about not duplicating markup.

src/assets/toolkit/styles/components/message.css
aileenjeffries: You flexbox solution is way better.

src/assets/toolkit/styles/components/message.css
aileenjeffries: Flexbox FTW

src/data/icons.yml
aileenjeffries: Makes sense! Switched to `close`

src/patterns/components/message/warning.hbs
aileenjeffries: Much nicer.

aileenjeffries:
Once more, with feeling, @tylersticka!

src/patterns/components/message/overview.hbs
```diff
@@ -0,0 +1,18 @@
...
+---
+
+<div class="Message {{messageClass}}">
+  {{#block "content"}}
+    <p>Hi, I'm a message!</p>
+  {{/block}}
+  {{#if includeClose}}
+    <button class="Message-close">
+      {{#svg "icons/close" aria-labelledby=(defaultTo closeIconTitle "Message-close-title")}}
+        <title id="Message-close-title">Close</title>
```
tylersticka: Gah, sorry, my mistake: the `id` value should be `{{defaultTo closeIconTitle "Message-close-title"}}`. No use supporting a custom title if I don't actually use it. (facepalm)

PR #367, opened 2016-10-13
[WIP] Breakup font-family styles

erikjung:
Related to: https://github.com/cloudfour/cloudfour.com-wp/pull/332

This changes the way `font-family` is applied to the body. Instead of setting our webfont (`Source Sans Pro`) on the `body` by default, we now require a `.fonts-loaded` class to be applied on a parent (`<html>`).

The reasoning behind this can be read here: https://www.filamentgroup.com/lab/font-events.html

It's related to the FOUT + FOIT that occurs when the browser is applying CSS and encounters an element trying to use a font defined by `@font-face`. By default, the font `url(...)` will be downloaded only when first needed: as soon as an element with `font-family: Source Sans Pro` is found. This can result in the following sequence:

0. FOUT when page is loading, prior to the `@font-face` definitions
0. FOIT after the `@font-face` definitions, when a font is needed to render but has yet to load
0. Expected font presentation

By making it so that a class must be added in order to apply the `font-family`, the sequence can instead be simplified to:

0. FOUT when page is loading, prior to the `@font-face` definitions
0. Expected font presentation, after dynamically loading the fonts with JS and adding the `fonts-loaded` class.

### Pros
- We can make the initial paint time a bit faster by deferring the loading of our `@font-face` CSS from Google.
- We can eliminate the secondary FOIT by manually loading the font and applying the `font-family` rule only after the associated font is loaded (and there should be no initial FOUT either after the fonts are cached).

### Cons 
- We need JS to apply our webfonts (similar to with TypeKit)
- More complexity in both JS and CSS

Thoughts?

### DIFF
--- a/src/assets/toolkit/styles/base/_config.css
+++ b/src/assets/toolkit/styles/base/_config.css
@@ -112,8 +112,10 @@
   --font-size-lg: calc(var(--ms2) * 1em);
   --font-size-xl: calc(var(--ms3) * 1em);
 
-  --font-family-default: "Source Sans Pro", Tahoma, "Helvetica Neue", Helvetica, Arial, sans-serif;
-  --font-family-mono: "Source Code Pro", Monaco, Consolas, "Lucida Console", "Andale Mono", monospace;
+  --font-family-default-fallbacks: Tahoma, "Helvetica Neue", Helvetica, Arial, sans-serif;
+  --font-family-default: "Source Sans Pro", var(--font-family-default-fallbacks);
+  --font-family-mono-fallbacks: Monaco, Consolas, "Lucida Console", "Andale Mono", monospace;
+  --font-family-mono: "Source Code Pro", var(--font-family-mono-fallbacks);
 
   --font-weight-normal: 400;
   --font-weight-semibold: 600;
@@ -148,8 +150,14 @@
 :root {
   --base-background: var(--color-gray);
   --base-color: var(--color-black);
-  --base-font: var(--font-size-sm)/var(--line-height-md) var(--font-family-default);
+  --base-font-size: var(--font-size-sm);
+  --base-line-height: var(--line-height-md);
+  --base-font-family: var(--font-family-default-fallbacks);
   --base-margin: var(--space-md);
+}
+
+:root {
+  --body-font-family: var(--font-family-default);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


tylersticka:
How does this technique compare to [embedding critical characters directly in the markup](https://www.zachleat.com/web/comprehensive-webfonts/) and then only deferring the full font with JS?

([Subset font I made back in May](https://trello-attachments.s3.amazonaws.com/56c6489dbd145e7ab71db5d8/5728d997267a58024da32b58/b633d3c47d4cebf2ec444ac978c04af7/SourceSansPro-regular-subset.woff), in case that's helpful.)

erikjung:
@tylersticka 

> How does this technique compare to embedding critical characters directly in the markup and then only deferring the full font with JS?

I assume you mean a scenario where:
- We have a `<style>` containing the subset font in a data-URI
- We defer/create the Google Fonts `<link>` with JS
- We don't alter the way `font-family` is applied within `toolkit.css` (and thus don't need to add the `fonts-loaded` class or leverage the Font Loading API)

Here is a branch with that approach: https://github.com/cloudfour/cloudfour.com-wp/compare/feat-pwa_style_loading_2

Compared to using no subset, deferring all the fonts, and using the `fonts-loaded` class to trigger the application of `font-family`, the only notable difference I saw when using the subset font was that we still had a FOIT. Typically, the subset font would show very quickly (sometimes immediately with no fallback seen), but we would still have a FOIT once the full font loaded and redefined the `@font-face`.

Running Lighthouse against each branch didn't provide any conclusive information about the initial rendering performance:

### With all fonts deferred and `fonts-loaded` class:

lighthouse_perf_wout_inline_1
lighthouse_perf_wout_inline_2
lighthouse_perf_wout_inline_3

### With inline subset and full fonts deferred:

lighthouse_perf_with_inline_1
lighthouse_perf_with_inline_2
lighthouse_perf_with_inline_3

One thing to note:

When using `wp_add_inline_style()` the resulting `<style>` gets rendered _after_ the `toolkit.css` stylesheet. Not sure at the moment how to make it render before. The function depends on an existing queued stylesheet handle. Perhaps the results would be different if the `<style>` preceded the `<link>`.

tylersticka:
> When using wp_add_inline_style() the resulting <style> gets rendered after the toolkit.css stylesheet. Not sure at the moment how to make it render before. The function depends on an existing queued stylesheet handle. Perhaps the results would be different if the <style> preceded the <link>.

You could also try messing with WP actions to append content of your choosing to the `<head>`, for example using [the `wp_head` hook](https://codex.wordpress.org/Plugin_API/Action_Reference/wp_head). You can specify a priority integer as a third argument to `add_action` if you want it to happen earlier or later.

erikjung:
Closing this since we won't need to make this change after all. See https://github.com/cloudfour/cloudfour.com-wp/issues/336

PR #372, opened 2016-11-15
Experiment: Replace webpack and Babel with Rollup and Bublé

erikjung:
This updates the `js` build task to use [Rollup](http://rollupjs.org) and [Bublé](https://buble.surge.sh) instead of webpack and Babel.

This effort is aimed at addressing the size of our JS bundles. Based on claims from [this article](https://nolanlawson.com/2016/08/15/the-cost-of-small-modules/) about Rollup producing smaller bundles by _tree-shaking_, I thought it would be worth it to experiment here and see how it handles our pattern library scripts.

Bublé is not coupled to Rollup, but seemed like a good fit for us. The transpiling strategy it uses aligns with the Babel preset we typically need (basic ES2015 syntax). So while using this tool instead of Babel isn't entirely necessary, it does simplify things a bit by shedding the need for configs/presets. It also claims to be faster (and a much smaller module), though I did not test this.

### The output of (webpack + Babel) vs. (Rollup + Bublé)

These are the output size differences between this branch vs. master, using local HAR files for the comparison. (On production with GZIP enabled, all of the numbers will be smaller.)

**Un-minified** (`gulp build --dev`)

```diff
toolkit.js
-size: 86.0K (86020 bytes)
+size: 63.7K (63766 bytes)
```

```diff
drizzle.js
-size: 4.2K (4287 bytes)
+size: 2.0K (2039 bytes)
```

bundle | webpack | Rollup | difference
---       | ---           | ---       | ---
toolkit.js | 86020 | 63766 | 26% smaller
drizzle.js | 4287 | 2039 | 52% smaller

**Minified** (`gulp build`)

```diff
toolkit.js
-size: 41.5K (41520 bytes)
+size: 35.6K (35627 bytes)
```

```diff
drizzle.js
-size: 1.7K (1789 bytes)
+size: 1.2K (1283 bytes)
```

bundle | webpack | Rollup | difference
---       | ---           | ---       | ---
toolkit.js | 41520 | 35627 | 14% smaller
drizzle.js | 1789 | 1283 | 28% smaller

So, there's a difference. Rollup does produce smaller bundles for us. Should we make this change?

### Pros
- Cleaner, smaller output
- Less configuration

### Cons
- Less community/plugins/etc.
- Less flexibility (mainly for the transpiling)

/CC @tylersticka


tylersticka:
Neat. Were the changes to `toolkit.js` required or an unrelated change?

erikjung:
@tylersticka 

> Neat. Were the changes to toolkit.js required or an unrelated change?

Good question about something I forgot to mention. That file was originally wrapped in an IIFE: 

```js
(function () { 
  // ...
}())
```

This is something that Rollup does for you via the `format: 'iife'` option.

The other changes (formatting, pushing to the `components` array) were not needed for Rollup, but were done to conform to our eslint config (which was complaining about the stray `new Thing()` calls).

Issue #379, opened 2016-11-18
Add `:active` styles for "Basic Button"


gerardo-rodriguez:
The "[Basic Button](http://cloudfour-patterns.netlify.com/patterns/components/button.html#basic-button)" should have an `:active` style. See conversation [in Slack](https://cloudfour.slack.com/archives/cloudfour-pwa/p1479511782000121).

This should be similar to the "[positive](https://github.com/cloudfour/cloudfour.com-patterns/blob/master/src/assets/toolkit/styles/components/button.css#L211-L213)" `:active` state, as an example.

gerardo-rodriguez:
The basic `Button` class uses `currentColor` for the `border-color`. This makes it a challenge. See https://github.com/postcss/postcss-color-function#can-you-support-currentcolor-so-we-can-do-colorcurrentcolor-adjuster

tylersticka:
`<thoughts>`

- If possible, I'd prefer the basic button _not_ have very specific pseudo state rules, unless those can safely cascade to every modifier. Otherwise it's going to be a specificity issue.
- With that in mind, the current `:hover` style is inherited from the link (`a`) style. So it would make more sense to add an `:active` color to default links, rather than graft an `:active` style onto the basic button.
- Alternatively, you could create a button modifier like `.Button--link` that you could explicitly style in the desired way.
- Also... should these be anchors? I notice they all do `href="#"`, which is a telltale sign that these are actually `<button>` elements in disguise. If they don't have a destination, they're probably not links. And if they should really be buttons, then you'll probably _need_ a new modifier to achieve what you're going for.

`</thoughts>`

gerardo-rodriguez:
> If possible, I'd prefer the basic button not have very specific pseudo state rules, unless those can safely cascade to every modifier. Otherwise we'll end up overriding it for every modifier, and colliding with it anytime we want the button to inherit its parent style.

Agree.

> Alternatively, you could create a button modifier like .Button--link that you could explicitly style in the desired way.

Yes, I was thinking this should probably be a modifier instead on my drive.

> Also... should these be anchors? I notice they all do href="#", which is a telltale sign that these are actually <button> elements in disguise. If they don't have a destination, they're probably not links. And if they should really be buttons, then you'll probably need a new modifier to achieve what you're going for.

Yep, good call. This also points to the idea that it should be a new button modifier.

Thanks for sharing your thoughts @tylersticka!

PR #380, opened 2016-11-28
Feature: Plain button

gerardo-rodriguez:
Re: #379 

## Overview

This PR adds a "plain" button (`Button--plain`) modifier. This modifier enhances the "basic" button by adding color changes on each of `:hover`, `:focus` & `:active` for `<a>`, `<button>` and `<input>` elements.

(The animated GIF makes it look like the `background-color` changes to a grayish color. Don't fall for the lies!!)

plain-button

---

cc: @tylersticka @erikjung

### DIFF
--- a/src/assets/toolkit/styles/components/button.css
+++ b/src/assets/toolkit/styles/components/button.css
@@ -235,3 +235,25 @@
 .Button--negative:active {
   background-color: color(var(--Button-negative-background-color) shade(10%));
 }
+
+/**
+ * Modifier: plain button
+ */
+
+:root {
+  --Button-plain-color: var(--color-blue);
+}
+
+.Button--plain {
+  color: var(--Button-plain-color);
+}
+
+.Button--plain:--enter {
+  color: color(var(--Button-plain-color) l(+10%));
+  border-color: color(var(--Button-plain-color) l(+10%));
+}
+
+.Button--plain:active {
+  color: color(var(--Button-plain-color) shade(10%));
+  border-color: color(var(--Button-plain-color) shade(10%));
+}
--- a/src/patterns/components/button/colors-dark.hbs
+++ b/src/patterns/components/button/colors-dark.hbs
@@ -1,6 +1,6 @@
 ---
 name: Dark Color Modifiers
-order: 3
+order: 4
 previewClass: Sky
 notes: |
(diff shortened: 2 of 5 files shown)


erikjung:
I'm a little confused by this. These changes LGTM, but it seems like we could address this by "fixing" the base `.Button` class instead of adding a new modifier:

0. Make `.Button` have a defined `color` and `border-color` of blue instead of `currentColor`
0. Apply the color changes needed for the pseudo classes*

* I get that there is a concern about needing to override the pseudo classes for each modifier, but it looks like we're already doing that, unless I'm misunderstanding.

tylersticka:
I guess I can see what @erikjung means. Perhaps the _idea_ of the base `.Button` class being neutral might contradict our actual usage.

erikjung:
@gerardo-rodriguez @tylersticka Lately I've been looking for low-impact ways to start leveraging custom properties lately.

It'll be nice when we can just do something like this: http://codepen.io/erikjung/pen/pNWeRe?editors=1100

gerardo-rodriguez:
> It'll be nice when we can just do something like this

Ah, yes, that makes sense @erikjung. I like the example you shared. Perhaps re-thinking how the button modifiers are set up may be a good idea.

erikjung:
@tylersticka @gerardo-rodriguez What if we _could_ use CSS vars to simplify this?

http://caniuse.com/#feat=css-variables
http://caniuse.com/#feat=css-supports-api

```css
@supports (--test: foo) {
  .Button:hover {
    background-color: --background-hover-color;
  }
}

@supports not (--test: foo) {
  .Button:hover {
    background-image: linear-gradient(
      rgb(0 0 0 / 15%),
      rgb(0 0 0 / 15%)
    );
  }
}
```

Paul-Hebert:
@gerardo-rodriguez , is this PR still relevant or can we close it out? I see you referenced this from a different PR that was merged in, but it doesn't looked like there's been any action here in a couple years. Are we waiting for better custom prop support to revisit this?

tylersticka:
I think we should probably close this out. I would approach this problem very differently today, so while the original intent is valid, I'm not sure this is the best path forward anymore.

Paul-Hebert:
Cool, sounds good @tylersticka , I'll close this out.

PR #393, opened 2016-12-13
Tweak: small CSS changes

erikjung:
The purpose of this was to just unprefix `text-decoration-skip`, but a few other tweaks made it in.

### DIFF
--- a/src/assets/toolkit/styles/components/thumbnail.css
+++ b/src/assets/toolkit/styles/components/thumbnail.css
@@ -51,7 +51,7 @@
   transition: border 300ms ease; /* 6 */
 }
 
-.Thumbnail:matches(:focus, :hover)::after {
+.Thumbnail:matches(:--enter)::after {
   border-color: color(var(--Thumbnail-border-color) a(25%));
 }
 
--- a/src/assets/toolkit/styles/sandbox/nm-nav-a.css
+++ b/src/assets/toolkit/styles/sandbox/nm-nav-a.css
@@ -109,11 +109,6 @@ body {
   transition: all 300ms ease;
 }
 
-/*.Masthead-control:matches(:--enter) {
-  color: var(--color-blue);
-  background-color: var(--color-white);
-}*/
-
 .Hero {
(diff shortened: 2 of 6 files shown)


src/assets/toolkit/styles/sandbox/nm-nav-a.css
```diff
@@ -109,7 +109,7 @@ body {
   transition: all 300ms ease;
 }
 
-/*.Masthead-control:matches(:focus, :hover) {
+/*.Masthead-control:matches(:--enter) {
```
tylersticka: This commented out CSS is probably an oversight. It would be fine to remove it.

src/assets/toolkit/styles/components/thumbnail.css
```diff
@@ -51,7 +51,7 @@
   transition: border 300ms ease; /* 6 */
 }
 
-.Thumbnail:matches(:focus, :hover)::after {
+.Thumbnail:matches(:--enter)::after {
```
tylersticka: Is `:matches` necessary in every case? I notice [it's missing from cssnext's example](http://cssnext.io/features/#custom-selectors).

erikjung: @tylersticka That is an interesting observation.

The output from that example is:

```css
button,
.button {
  /* styles for your buttons */
}
button:hover,
.button:hover,
button:focus,
.button:focus {
  /*
    hover/focus styles for your button

    Read more about :enter proposal
    http://discourse.specifiction.org/t/a-common-pseudo-class-for-hover-and-focus/877
   */
}
```

So it _does_ appear that when you concatenate a custom selector that itself is a multiple selector, it behaves like `:matches()`.

screen shot 2016-12-13 at 1 00 58 pm

What's unfortunate is how `:not()` behaves a bit differently:

screen shot 2016-12-13 at 1 04 55 pm

tylersticka: Yeah, that's... weird.

Issue #394, opened 2017-01-04
text-decoration-skip: ink only output with webkit prefix


tylersticka:
[The `text-decoration-skip` property should be supported in Chrome Canary](http://thenewcode.com/1168/Improve-the-Look-of-Links-with-the-CSS-Text-Decoration-Module), but in practice it isn't because the property is only output with a `-webkit` prefix:

![screen shot 2017-01-04 at 8 42 50 am](https://cloud.githubusercontent.com/assets/69633/21650500/f611633a-d259-11e6-954e-cb1fe27fdacb.png)

Removing the prefix in the inspector causes the property to work:

![screen shot 2017-01-04 at 8 43 03 am](https://cloud.githubusercontent.com/assets/69633/21650514/002516fa-d25a-11e6-9675-4dac13243321.png)

This _should_ be a matter of simply recompiling with Autoprefixer, but it's unclear whether or not [the caniuse.com entry for this feature](http://caniuse.com/#feat=text-decoration) is up-to-date. It may be [in need of a pull request](https://github.com/Fyrd/caniuse/blob/master/features-json/text-decoration.json).

erikjung:
Looks like we need to update cssnext. The latest version outputs it with and without the prefix.

<img width="1276" alt="screen shot 2017-02-03 at 9 52 26 pm" src="https://cloud.githubusercontent.com/assets/61204/22616115/3393f5e2-ea5b-11e6-9264-7cb708e43c09.png">

PR #395, opened 2017-02-07
Update cssnext

tylersticka:
Resolves issue with `text-decoration-skip: ink`

See: #394


tylersticka:
@erikjung @saralohr @aileenjeffries Bump. 🙂

Issue #369, opened 2016-11-02
Not enough contrast for accessibility


grigs:
I had noticed that Lighthouse had complained about some accessibility issues on our site. I used tota11y to find the the following:

<img width="1259" alt="screen shot 2016-11-02 at 1 22 27 pm" src="https://cloud.githubusercontent.com/assets/804237/19946002/ce3e935e-a0ff-11e6-8482-1c2914b37bef.png">

I'm a bit surprised by this because they all look they have sufficient contrast, but I trust the math more than my eyes.

tylersticka:
An instance of this came up recently: https://github.com/cloudfour/cloudfour.com-patterns/pull/365#discussion_r81851388

tylersticka:
I feel I've addressed this as best as I can in #402, but see that issue for an overview of a couple other potential issues.

I'm going to close this, we can open individual issues for specific problem areas in the future.

PR #402, opened 2017-03-22
Add a11y contrast variations of blue, fuchsia

tylersticka:
Resolves an issue where `<a>` and `<code>` sometimes failed to meet WCAG 2 AA contrast ratio thresholds.

I chose to create variations of our base color palette variables because changing the root colors caused a pretty dramatic change in appearance, most noticeably to elements that _didn't_ have contrast problems. My hope is by limiting the color changes to dark text only, it'll make the difference imperceptible to the average viewer.

It does not address contrast issues in the following scenarios:

- Orange buttons and message containers with white text. These only fail contrast ratio checks at the smallest sizes, and the only solution appears to be to adopt more of a _brown_ color... no small change to the color made a dent in the warnings. I find this _really_ peculiar... why would `18px` text pass without changes, but `16px` fails unless a _dramatic_ color shift occurs?
- Label elements in our `FloatLabel` pattern. Any darker and they could be easily mistaken for actual text, which defeats the purpose entirely. We'd be better off [abandoning the pattern](https://www.nngroup.com/articles/form-design-placeholders/).

I went and compared our contrast warnings to a few competitors I admire (using [aXe](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US), since tota11y appears to be abandoned), and with this change we'll show far fewer warnings per page at small viewport sizes than any of those I inspected. I feel confident we've gone above and beyond most of our peers in this space.

## Before

screen shot 2017-03-22 at 1 40 36 pm

screen shot 2017-03-22 at 1 42 38 pm

## After

screen shot 2017-03-22 at 1 40 27 pm

screen shot 2017-03-22 at 1 42 46 pm

---

@grigs @erikjung @gerardo-rodriguez @aileenjeffries @saralohr 

See #369

### DIFF
--- a/src/assets/toolkit/styles/base/_config.css
+++ b/src/assets/toolkit/styles/base/_config.css
@@ -39,6 +39,10 @@
 
   /* Other colors */
   --color-red: #bf0000;
+
+  /* a11y variations for light-on-dark text contrast */
+  --color-blue-a11y: color(var(--color-blue) l(-4%));
+  --color-fuchsia-a11y: color(var(--color-fuchsia) l(-14%) s(-5%));
 }
 
 /**
@@ -160,7 +164,7 @@
  */
 
 :root {
-  --link-color: var(--color-blue);
+  --link-color: var(--color-blue-a11y);
   --link-hover-color: color(var(--link-color) l(+10%));
 }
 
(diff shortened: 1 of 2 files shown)


erikjung:
Also FWIW, this could be addressed using `contrast()`: 

http://codepen.io/erikjung/pen/yMjVdN?editors=1100

tylersticka:
> Also FWIW, this could be addressed using contrast():

Well, you tell me that _after_ you approve the merge. 😜

Issue #385, opened 2016-12-07
Message alignment is off


tylersticka:
Noticing in Chrome on desktop (haven't checked elsewhere) that the alignment of `.Message` could use some finagling. It always looks like there's more space on the bottom than the top.

![screen shot 2016-12-07 at 1 54 28 pm](https://cloud.githubusercontent.com/assets/69633/20988328/2118e4ee-bc85-11e6-9d11-e04ea6bbefae.png)

![screen shot 2016-12-07 at 1 51 17 pm](https://cloud.githubusercontent.com/assets/69633/20988329/23e33c74-bc85-11e6-99ee-565b235ef1f3.png)

PR #403, opened 2017-03-22
Fix message alignment

tylersticka:
This tweaks the `Message` pattern so that its alignment is consistent as the base `font-size` changes.

There were two reasons this was happening. First, there was whitespace between the `Message-close` element and the SVG within. And second, this problem was worse at some sizes more than others because the icon was being sized as `24px` regardless, but everything around it was changing, which is why some resolutions were more symptomatic than others.

To change this, I borrowed an `Icon--block` modifier technique from our other recent projects so we can piggy-back on the `Icon` class and its built-in sizing logic. Then, I just made `Message-close` use `display: flex;` so its content wouldn't flow as text.

I also fixed some issues I saw with `line-height` and space between the "close" button and the text.

**Note:** To get the new icon sizing on the server, the markup will need to be updated to include the `Icon Icon--block` class on the close button. That said, alignment will still be improved even without that change, and nothing will appear overtly broken with _only_ styles applied, so everything should be A-OK.

## Before

screen shot 2017-03-22 at 2 56 43 pm

screen shot 2017-03-22 at 2 49 59 pm

## After

screen shot 2017-03-22 at 2 56 58 pm

screen shot 2017-03-22 at 2 49 44 pm

---

@aileenjeffries @saralohr @erikjung @gerardo-rodriguez 

See: #385

### DIFF
--- a/src/assets/toolkit/styles/components/icon.css
+++ b/src/assets/toolkit/styles/components/icon.css
@@ -12,11 +12,8 @@
 }
 
 /**
- * 1. Because vertical alignment is funny, icons always appear about a pixel too
- *    low when they're the same height as the current `font-size`. Using `top`
- *    frees up `margin-top` for adjustments we'll be applying to modifiers.
- * 2. Base dimensions on the current `font-size` for leaner adjustments.
- * 3. Fill with the current text color by default. May be overridden by icon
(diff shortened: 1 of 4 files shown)


aileenjeffries:
ahhhhh... much better. :)

PR #407, opened 2017-03-24
PWA Stats prototypes

saralohr:
Adds prototypes for PWA Stats site, with some prototype CSS. I'll do another PR to bring these styles into the fold.

index-sm
index-lg
single-sm
single-lg


src/assets/toolkit/styles/sandbox/sil-pwa.css
```diff
@@ -0,0 +1,74 @@
...
+  color: color(var(--color-gray) l(-45%) s(+5%)) !important;
+}
+
+/* Setting columns with min-width
+ * Probably break gap size into another utility
+ */
+.u-columns {
+  columns: 14.5rem 5;
+  column-gap: calc(1em * var(--ms1));
+}
```
tylersticka: FYI, for the final styles I might opt to use `display: grid` and a simplified fallback.

PR #408, opened 2017-03-24
Componentize PWA Stats

saralohr:


saralohr:
After some conversation with @tylersticka we are thinking we might want to do something different to handle the PWA Stats, like including them in a separate file, or something.

tylersticka:
[Convert prototypes to components and/or site-specific stylesheet](https://trello.com/c/T1UHzwcx/12-convert-prototypes-to-components-and-or-site-specific-stylesheet)

PR #409, opened 2017-03-24
Add hack for making comment footnotes smaller

tylersticka:
Part one of a two-part fix to make this edge case prettier.

## Before

screen shot 2017-03-24 at 4 14 56 pm

## After

screen shot 2017-03-24 at 4 14 46 pm

---

@erikjung @gerardo-rodriguez @aileenjeffries


tylersticka:
@erikjung Yes, it's handled by Jetpack's Markdown plugin and is not configurable. That whole file is nothing but hacks to get footnotes looking presentable.

PR #411, opened 2017-03-27
Address Safari flexbox bug in Pagination

### LINKED ISSUE #410: Pagination arrow & numbers should not overlap in smaller viewports
This happened on my iPhone 5s. Notice the arrow and the "7". 

img_9354

tylersticka:
See code comments for detailed explanation.

---

Fixes #410

### DIFF
--- a/src/assets/toolkit/styles/components/pagination.css
+++ b/src/assets/toolkit/styles/components/pagination.css
@@ -39,12 +39,30 @@
 }
 
 /**
- * Child elements should not grow, should shrink as necessary, and it should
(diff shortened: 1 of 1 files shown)


tylersticka:
@erikjung 

> (I am not familiar with all of the requirements of this component, so this may be oversimplified.)

The only things I see off-hand that may be over-simplified:

- We want to avoid the additional "next" and "previous" links because when navigating via screen reader it makes things confusing and repetitive. That's why the "next" and "previous" controls in the current version are actually just children of those page links.
- For usability, it's better for the padding and stuff to be on the links themselves rather than the parent items.

PR #414, opened 2017-03-28
PWA Stats R2

tylersticka:
This PR makes changes to three existing components, and refactors the PWA Stats prototypes to use CSS Grid Layout.

I'm doing a PR for this now because I'm still somewhat skittish about adding the PWA Stats specific components to the pattern library proper. They have some idiosyncratic styles that depend on being a child of `.Sky` (or at least a dark background, which is uncommon for the rest of cloudfour.com).


tylersticka:
[Convert prototypes to components and/or site-specific stylesheet](https://trello.com/c/T1UHzwcx/12-convert-prototypes-to-components-and-or-site-specific-stylesheet)

PR #433, opened 2018-07-12
Simplify pagination

### LINKED ISSUE #431: Simplify pagination items cap width
I think it would be beneficial to replace this CSS:

https://github.com/cloudfour/cloudfour.com-patterns/blob/master/src/assets/toolkit/styles/components/pagination.css#L64
```css
.Pagination > *:nth-last-child(-n+5):first-child,
.Pagination > *:nth-last-child(-n+5):first-child ~ * {
  flex: 0 1 calc(100% / 8);
}
```

With a max-width on `.Pagination > *`

@tylersticka I'd like to hear more details about this:
> I'm not comfortable making that change in the pattern library because there's a lot of history and edge-cases to consider.

calebeby:
Fixes #431 

90px is arbitrary, maybe we should use a different number (do we want to use viewport units?)


calebeby:
Please note: master is messed up (postcss plugins updating), and so so is this PR (until #432 is merged)

tylersticka:
@calebeby `90px` does feel arbitrary. Can we use a percentage?

tylersticka:
> If you think they should keep spreading out as the screen keeps getting wider, we could use a combination of px + % (i.e. calc(3% + 50px)

It'd be nice if they could continue spreading out as the screen got wider!

But my minimum requirement is that we not use pixels, because:

1. They will not scale as we change the root `font-size` at larger breakpoints.
1. They can make it harder for [users to change `font-size`](https://medium.com/@vamptvo/pixels-vs-ems-users-do-change-font-size-5cfb20831773) reliably.

I'll approve just about any solution that uses `rem`, `em`, `%`, `vw` or whatever as long as it looks good and doesn't introduce functional regressions.

calebeby:
It does not work in Safari 10 (only one item is shown), but Safari 10 does not match our [Browserslist](http://browserl.ist/?q=%3E+1%25%2C+Last+1+version). Do we care about supporting Safari 10 (should I make a PR to change the browserslist), or are we fine with not supporting it?

calebeby:
I found a workaround for Safari 10 (It's in one of the pages you linked in the comment 😂). Should I include the workaround and send a PR to update the browserslist?

(Safari 10 has about .3% of browser usage, from https://caniuse.com/usage-table)

tylersticka:
@calebeby How does it break in Safari 10?

If the issue is merely cosmetic, it's probably not worth addressing.

If the issue is more than cosmetic (like, you can't navigate back and forth at all), then it depends on whether or not the fix is horrendous. If it's a simple fix and it'll help an occasional user, then it seems worth including.

calebeby:
It is the same issue that led you to using the number of children fix.

All the flex items were arranged vertically, and only the first one was being shown (because the other ones were cut off)

tylersticka:
@calebeby Were the "next" and "back" arrows still visible? (It's been a while since that bug.)

calebeby:
Yep, looks like there is a top-level service worker that imports both ours and theirs

calebeby:
Yes

It looked like this:

```
<-            1            ->
```
and then below that (was not visible)

```
              2
              3
              4
             etc.
```

tylersticka:
Gotcha. So is the fix horrendous?

calebeby:
Nope, not horrendous at all. The workaround is literally changing

```css
  flex: 1 1 auto;
```

to

```css
  flex: 1 1 calc(3.5em + 6%);
```

tylersticka:
Oh! Then let's go for it.

src/assets/toolkit/styles/components/pagination.css
```diff
@@ -39,31 +39,18 @@
...
 
 .Pagination > * {
-  flex: 1 1 auto;
-}
-
-.Pagination > *:nth-last-child(-n+5):first-child,
-.Pagination > *:nth-last-child(-n+5):first-child ~ * {
-  flex: 0 1 calc(100% / 8);
+  flex: 1 1 calc(3.5em + 6%); /* specified flex-basis to fix Safari 10 bug */
+  max-width: calc(3.5em + 6%);
```
tylersticka: It would be nice if that size were a custom property since it's repeated twice.

src/assets/toolkit/styles/components/pagination.css
```diff
@@ -39,31 +39,18 @@
...
+  width: 100%; /* Fixes centering issue in ie11 */
```
tylersticka: Nit: Please follow [our comment conventions](https://github.com/cloudfour/guides/tree/master/css#comments).

PR #435, opened 2018-07-19
Add Safari 10 to browserslist

calebeby:
If we are going to have workarounds to support Safari 10, we should also ensure that autoprefixer generates prefixes compatible with Safari 10 and that postcss-preset-env emits syntax compatible with Safari 10

### DIFF
--- a/browserslist
+++ b/browserslist
@@ -4,3 +4,4 @@
 
 > 4%
 Last 1 version
+Safari 10
--- a/browserslist
+++ b/browserslist
@@ -4,4 +4,4 @@
 
 > 4%
 Last 1 version
-Safari 10
+Last 3 Safari version


browserslist
```diff
@@ -4,3 +4,4 @@
 
 > 4%
 Last 1 version
+Safari 10
```
tylersticka: Is there any benefit to making this `Safari > 9` or similar, to prevent any future weird "gaps" where fixes are included for Safari 10 but not 11?

calebeby: Would `Last 3 Safari versions` do what we want it to?

```
Safari 11.1
Safari 11
Safari 10.1
```

Then it would require less maintenance, because it would automatically update when new versions are released.

Another thing that might be worth looking into is [`browserslist-ga`](https://github.com/browserslist/browserslist-ga). What do you think?

tylersticka: I do not have a strong opinion on how it's done as long as the version gap problem is solved.

spaceninja:
@tylersticka I updated to `Last 3 Safari versions` — any objects to merging this so I can clear the decks?

PR #444, opened 2019-09-25
Update Colors and Button Styles

dromo77:
## Summary

The existing "primary" and "positive" buttons don't meet WCAG standards for accessibility. To make the “primary” and “positive” buttons accessible, this PR updates several colors including:

- `--color-green`
- `--color-fuchsia`
- `--color-red`

It also adds a new page to the sandbox to test how the new button styles look in use.

### Before
before

### After
after

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/*
+lts/carbon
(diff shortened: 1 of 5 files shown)


tylersticka:
Tested locally, works great, passes contrast for buttons 👍

PR #451, opened 2019-09-27
Add New Spacing Utilities

dromo77:
## Summary 

While we have `u-staggerItems` utilities, sometimes they don't provide enough contrast between  elements. This PR add two new spacing utilities that help to adjust the space between headings and their preceding content:

- `u-staggerHeadings`
- `u-staggerNestedHeadings`

### `u-staggerItems`

stagger-items

### `u-staggerHeadings`

stagger-headings

### `u-staggerNestedHeadings`

stagger-nested-headings

### Space Utilities

utilities

### DIFF
--- a/src/assets/toolkit/styles/utils/space.css
+++ b/src/assets/toolkit/styles/utils/space.css
@@ -11,11 +11,8 @@
     margin-top: $(value) !important;
   }
 
-  .u-staggerHeadings$(suffix) > * + :--headings {
-    margin-top: $(value) !important;
-  }
-
-  .u-staggerNestedHeadings$(suffix) > * > :--headings {
+  .u-staggerHeadings$(suffix) > * + :--headings,
+  .u-staggerHeadings$(suffix) > * + .HashHeading {
     margin-top: $(value) !important;
   }
 }
--- a/src/assets/toolkit/styles/utils/space.css
+++ b/src/assets/toolkit/styles/utils/space.css
@@ -10,6 +10,10 @@
   .u-staggerItems$(suffix) > * + * {
     margin-top: $(value) !important;
   }
+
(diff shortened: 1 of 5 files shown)


src/assets/toolkit/styles/utils/space.css
```diff
@@ -10,6 +10,14 @@
...
     margin-top: $(value) !important;
   }
+
+  .u-staggerHeadings$(suffix) > * + :--headings {
+    margin-top: $(value) !important;
+  }
+
+  .u-staggerNestedHeadings$(suffix) > * > :--headings {
+    margin-top: $(value) !important;
+  }
```
tylersticka: Instead of having separate utilities, could we add `.HashHeading` to the main utility?

```css
.u-staggerHeadings$(suffix) > * + :--headings,
.u-staggerHeadings$(suffix) > * + .HashHeading {
  margin-top: $(value) !important;
}
```

My concern with the technique used here is that we're already adding space above the hash heading, so the nested headings will be additive (which likely isn't what we want).

dromo77: This solution accomplishes what we want, but out of curiosity would we normally use a utility in combination with a specific class in this way?

tylersticka: No, but we normally wouldn't target specific elements either. This utility is kind of an outlier anyway.

Since it's already a little bit of a hack, combining the utilities has two benefits:
1. It now requires one utility instead of two.
1. The behavior of the utility is more predictable. Instead of upper margin being added to the container via `u-staggerItems` _plus_ additional margin being added to a nested heading (which may or may not add up to a predictable value), we are applying the margin to a single element.

PR #453, opened 2019-10-01
Differentiate fuchsia variations

tylersticka:
## Overview

As part of #444, we updated the shade of `--color-fuchsia` for accessibility. Unfortunately, this introduced regressions to the contrast accessibility of that color on dark backgrounds, for example in code blocks.

This PR restores the previous value of `--color-fuchsia`, moves the new value to `--color-orchid` (based on the most similar color keyword) and applies either value selectively.

### Before

Screen Shot 2019-10-01 at 12 22 15 PM

### After

Screen Shot 2019-10-01 at 12 22 02 PM

### DIFF
--- a/src/assets/toolkit/styles/base/_config.css
+++ b/src/assets/toolkit/styles/base/_config.css
@@ -45,7 +45,8 @@
   /* Secondary colors */
   --color-orange: #f26941;
   --color-green: #068362;
-  --color-fuchsia: #dd117d;
+  --color-fuchsia: #F14CA3;
+  --color-orchid: #dd117d;
 
   /* Other colors */
   --color-red: #de1d0b;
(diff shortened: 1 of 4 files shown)


tylersticka:
Deploy preview: https://5d93a827607f37000a4a6a39--cloudfour-patterns.netlify.com/

Paul-Hebert:
Code looks good — viewing the deploy

tylersticka:
> I wonder if if would be better to have two drastically different accent colors rather than two similar but different colors?
> 
> Or maybe we should explain that one is for contrasting against dark and the other is for contrasting against light?

I think both of these are good suggestions. I'll create a separate issue, though, since this is primarily an attempt to fix a regression.

Paul-Hebert:
Sounds good, thanks!

PR #460, opened 2019-10-15
Syntax highlighting tweaks

tylersticka:
## Overview

I noticed on some recent blog post drafts that functions had no syntax highlighting in either CSS or JavaScript. This adds that, plus tweaks the green just a tad (since that hue was changed for buttons recently).

## Screenshot

Screenshot_2019-10-15 Cms Content - Drizzle

### DIFF
--- a/src/assets/toolkit/styles/vendor/prism.css
+++ b/src/assets/toolkit/styles/vendor/prism.css
@@ -31,7 +31,8 @@
 .token.tag,
 .token.regex,
 .token.number,
-.token.boolean {
+.token.boolean,
+.token.function {
   color: color-mod(var(--color-fuchsia) l(+10%));
 }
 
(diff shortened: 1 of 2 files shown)


src/assets/toolkit/styles/vendor/prism.css
```diff
@@ -24,14 +24,15 @@
 .token.attr-value,
 .token.selector,
 .token.class-name {
-  color: color-mod(var(--color-green) l(+15%));
+  color: color-mod(var(--color-green) l(+17%));
```
Paul-Hebert: If this was changed to match the updated button color I wonder if it would storing this value in a variable?

Note: I'm missing a _ton_ of context here so I don't know if this suggestion makes sense given our compilation setup, buttons styles, colors, etc.

tylersticka: Because the code styles are one of the few areas where we're inverting more than white/gray on dark backgrounds, it's sort of an outlier for the library. You'll in the CSS file that we're tweaking a few colors for this context.

Paul-Hebert: Cool, no worries! Like I said I was missing a ton of context, but thought I might be remiss if I didn't mention it at all

PR #495, opened 2020-02-24
Theo design tokens

### LINKED ISSUE #471: Design token integration
To keep things DRY and avoid the need to duplicate little updates, we should use [Theo](https://github.com/salesforce-ux/theo) to store our design tokens in a single place we can then reference from Sass or our Storybook documentation (to start with).

## Tasks

- [x] Install Theo as a dependency
- [x] Add a test YAML file for Theo to interpret (for example, a dummy color palette)
- [x] Consume those values via Sass
- [x] Display those files in an `.mdx` story

It is acceptable for this integration to be tested via dummy patterns and docs we will remove later.

tylersticka:
## Overview

Resolves #471 by integrating Theo support into Sass stylesheets and Storybook stories. Both are accomplished via a small amount of custom code.

The result is that you can define properties in YAML files conforming to [the Theo spec](https://github.com/salesforce-ux/theo#spec), which you may then consume via Sass or JS. Changes made to these design tokens automatically refresh in Storybook.

Theo was chosen because its API was more flexible than Amazon's Style Dictionary, allowing direct integration with Sass and Webpack/Storybook. Other solutions would have required an intermediary step where files were generated and subsequently parsed.

The Sass portion is based on [Basalt's popular article on the subject](https://basalt.io/blog/theo-design-tokens-using-node-sass-importer-for-any-build-method), though simplified in one or two small areas.

I'd originally intended on using [Autodesk's theo-loader](https://github.com/Autodesk/theo-loader), but could not get it to work properly even in a new project. I'm guessing it may be out of date. It also had limitations as to the file types supported.

Luckily, writing a custom loader proved simpler than I anticipated.

## Testing

1. Check out branch locally.
1. `npm run storybook`
1. Edit the value of `blue` in `src/demo/design-tokens/colors.yml`. Observe changes applying to demo button and color page.
1. Edit the value of `brand_primary` directly in `src/demo/design-tokens/brand.yml`. Observe changes applying to demo button and color page.

## Screenshots

storybook-theo


.storybook/theo-importer.js
```diff
@@ -0,0 +1,42 @@
...
+ * Import Theo YAML files into SCSS files.
+ *
+ * @param {string} url - File path relative to SCSS file.
+ * @param {string} prev - File path of SCSS file that made import request.
+ * @param {function({ file: string, contents: string })} done - Callback to
+ * pass results to.
+ * @link <https://basalt.io/blog/theo-design-tokens-using-node-sass-importer-for-any-build-method>
+ * @link <https://github.com/sass/node-sass#importer--v200---experimental>
+ */
+function theoImporter(url, prev, done) {
```
spaceninja: `url` and `prev` are somewhat confusing variable names. I'd suggest `yamlPath` and `scssPath` to make things a bit more clear.

.storybook/theo-loader.js
```diff
@@ -0,0 +1,36 @@
+const { resolve } = require('path');
+const theo = require('theo');
+const yaml = require('js-yaml');
+
+function loader(source) {
```
spaceninja: It'd be nice to add a JSDoc block here, like you did in `theo-importer.js`

.storybook/theo-importer.js
tylersticka: @spaceninja I thought the same, but these are actually the variable names defined in Sass's importer documentation. Keeping these the same made it a bit easier to troubleshoot by referring to the Sass API docs.

.storybook/theo-loader.js
tylersticka: Good idea!

.storybook/theo-importer.js
spaceninja: 👍  that makes sense

tylersticka:
@spaceninja I made two changes:

1. Improved the Webpack loader documentation.
1. Output the loader result as a JS module. This makes tree-shaking possible in the future, but more importantly, it makes the results camel-case instead of underscore-separated! 😅 

Re-review appreciated!

.storybook/theo-importer.js
gerardo-rodriguez: > I thought the same, but these are actually the variable names defined in Sass's importer documentation. Keeping these the same made it a bit easier to troubleshoot by referring to the Sass API docs.

Can a comment be added explaining this so a future dev doesn't have the same question/thought? 🙂

tylersticka: @gerardo-rodriguez Done!

.storybook/theo-loader.js
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Webpack loader for Theo YAML/YML.
+ *
+ * @param {string} source - The contents of the requested file.
+ */
+function theoLoader(source) {
+  // The absolute path to the requested design token file.
+  const tokenPath = this.resourcePath;
+  // Lets Webpack now this is an async loader and provides us a callback.
```
spaceninja: typo: "Lets Webpack know this is…"

PR #499, opened 2020-02-25
Add initial base styles

### LINKED ISSUE #476: Base styles
It's pretty incredible [how small our CSS resets can be now](https://meowni.ca/posts/monica-dot-css/). But I do think there's value in establishing some very basic foundational styles. 

Off the top of my head…

```css
/**
 * Sensible box model
 */

*,
::before,
::after {
  box-sizing: border-box;
}

/**
 * Promote consistent line spacing.
 *
 * @see https://drafts.csswg.org/css-inline-3/#line-sizing-property
 */

html {
  line-sizing: normal;
}

/**
 * 1. Remove white space surrounding page.
 * 2. Prevent size adjustments on orientation change.
 * 3. Allow long words (like URLs) to break at arbitrary points.
 */

body {
  margin: 0; /* 1 */
  text-size-adjust: none; /* 2 */
  word-wrap: break-word; /* 3 */
}

/**
 * 1. Block display is usually what we want
 * 2. Remove strange space-below when inline
 * 3. Responsive by default
 * 
 * @see https://github.com/mozdevs/cssremedy
 */

audio,
canvas,
embed,
iframe,
img, 
object,
svg, 
video {
  display: block; /* 1 */
  vertical-align: middle; /* 2 */
  max-width: 100%; /* 3 */
}

canvas,
img,
svg, 
video {
  height: auto;
}

audio {
  width: 100%;
}

/**
 * Don't allow code blocks to overflow willy-nilly.
 */

pre {
  overflow: auto;
}

/**
 * Old browser fixes
 */

img {
  border-style: none;
}

[hidden] {
  display: none !important;
}

article,
aside,
figcaption,
figure,
footer,
header,
hgroup,
main,
nav,
section {
  display: block;
}
```

tylersticka:
## Overview

This PR adds some initial base/reset/normalize styles to the project. It then includes these in every Storybook preview via [a preview configuration file](https://medium.com/storybookjs/declarative-storybook-configuration-49912f77b78). This file also allowed me to simplify configuration of global decorators, specifically the a11y add-on.

After implementing these base styles, I noticed that all of the examples were running up against the edge of their container. I decided to install the extremely popular [centered add-on](https://github.com/storybookjs/storybook/tree/master/addons/centered), but if we dislike it we can use the `preview.scss` file to further customize however we wish.

## Testing

Either locally or at the deploy preview URL:

1. Inspect any of the demo elements and confirm that its `box-sizing` has been set to `border-box`.
1. Observe that story canvas contents are now centered.

## Screenshot

Screen Shot 2020-02-24 at 4 09 42 PM

---

Closes #476

### DIFF
--- a/src/base/_defaults.scss
+++ b/src/base/_defaults.scss
@@ -1,0 +1,78 @@
+/**
+ * Sensible box model
+ */
+*,
+::before,
+::after {
+  box-sizing: border-box;
+}
+
+/**
+ * 1. Remove white space surrounding page.
+ */
+body {
+  margin: 0; /* 1 */
+}
+
+/**
+ * 1. Block display is usually what we want
+ * 2. Remove strange space-below when inline
+ * 3. Responsive by default
+ *
+ * @see https://github.com/mozdevs/cssremedy
+ */
+audio,
+canvas,
+embed,
+iframe,
+img,
+object,
+svg,
+video {
+  display: block; /* 1 */
+  vertical-align: middle; /* 2 */
+  max-width: 100%; /* 3 */
+}
+
+canvas,
+img,
+svg,
+video {
+  height: auto;
+}
+
+audio {
+  width: 100%;
+}
+
+/**
+ * Don't allow code blocks to overflow willy-nilly.
+ */
+pre {
+  overflow: auto;
+}
+
+/**
+ * Old browser fixes
+ */
+
+img {
+  border-style: none;
+}
+
+[hidden] {
+  display: none !important;
+}
+
+article,
+aside,
+figcaption,
+figure,
+footer,
+header,
+hgroup,
+main,
+nav,
+section {
+  display: block;
+}
--- a/src/base/_defaults.scss
+++ b/src/base/_defaults.scss
@@ -8,10 +8,10 @@
 }
 
 /**
- * 1. Remove white space surrounding page.
+ * Remove white space surrounding page.
  */
 body {
-  margin: 0; /* 1 */
+  margin: 0;
 }
 
 /**
@@ -34,6 +34,10 @@ video {
   max-width: 100%; /* 3 */
 }
 
+/**
+ * Embedded elements with consistent aspect ratios require this to maintain
+ * that ratio when resized.
+ */
 canvas,
 img,
 svg,
@@ -41,6 +45,9 @@ video {
   height: auto;
 }
 
+/**
+ * It's pretty arbitrary that audio elements are 300px wide by default.
+ */
 audio {
   width: 100%;
 }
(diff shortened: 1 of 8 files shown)


tylersticka:
I'm actually noticing a problem with this technique. It is putting the base styles _after_ the component styles. That will introduce problems.

I actually understand _why_ they would do this: Most of the time, you'd want preview-specific styles _after_ component styles. But I think I need to re-think my approach.

tylersticka:
False alarm! I was mis-reading the inspector. Inspected the source and it is outputting in the expected order. Hooray!

src/base/_defaults.scss
```diff
@@ -0,0 +1,78 @@
+/**
+ * Sensible box model
+ */
+*,
+::before,
+::after {
```
spaceninja: You're only applying the sensible box model to generated content? I've typically seen this rule written to include a `*` selector, so everything gets it by default.

src/base/_defaults.scss
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Don't allow code blocks to overflow willy-nilly.
+ */
+pre {
+  overflow: auto;
+}
+
+/**
+ * Old browser fixes
```
spaceninja: Could you document what these fixes are for?

src/base/_defaults.scss
spaceninja: OMG I totally missed the first `*` selector because my brain lumped it in with the comment.

Sorry! My bad. I'll approve now.

src/base/_defaults.scss
tylersticka: @spaceninja Done!

src/demo/button/button.stories.mdx
```diff
@@ -1,12 +1,11 @@
 import { Story, Preview, Meta } from '@storybook/addon-docs/blocks';
 import { withKnobs, text, boolean, select } from '@storybook/addon-knobs';
-import { withA11y } from '@storybook/addon-a11y';
 import template from './button.twig';
 import './button.scss';
 
 <Meta
   title="Demo/Button"
-  decorators={[withKnobs, withA11y]}/>
+  decorators={[withKnobs]}/>
```
calebeby: What did this change do?

calebeby: Still trying to understand how this is working. Is it no longer necessary to specify withA11y because it is in the preview.js file?

tylersticka: The way the accessibility add-on works, it applies to _all_ stories once it's been added anywhere. Now that I've added a `preview.js` config file, I can define it _there_ once, which felt like a more intuitive place to initialize an add-on that applies everywhere.

Issue #498, opened 2020-02-24
Icon assets + tokens + pattern


tylersticka:
- [x] Bring in icon assets
- [x] Determine Twig integration strategy
- [x] Display as story
- [x] Basic CSS pattern
  - [x] Document component

tylersticka:
The Twig integration strategy portion of this has proven a bit more complicated than I was expected.

Below are all the options I've explored. I am currently leaning toward **Option 4**, but wanted to confirm with someone in @cloudfour/dev before proceeding in that direction.

## The Problem

SVGs benefit from some manipulation before they are embedded in a document:

- XML attributes (`xmlns`) and other cruft that isn't necessary or helpful in an HTML5 document may be safely stripped.
- Attributes like `class`, `role`, etc. may need to be manipulated so an SVG element has appropriate presentation and semantics for different situations.

To enable those sorts of manipulations, we need to coordinate between the SVG files and the template engine. In our case, the engine is [Twig](https://twig.symfony.com/doc/2.x/).

There are two template engine wrinkles to consider:

- Storybook refreshes components on the client side, and neither Twig.js or Twing (the two most mature Twig implementations for JavaScript) expose included files that weren't explicitly requested to begin with. This means that `{% include 'path/to/file.twig' %}` works fine, but `{% include filename_variable %}` will _not_ work because by the time `filename_variable` is evaluated in the client, Twig will have already finished compiling all the templates it _thought_ it needed.
- By default, Twig doesn't like including files that don't end in `.twig` or `.html`.

## Option 1: Inline SVG extensions for JavaScript and PHP

We could author or install a `inlineSvg` helper in both JavaScript and PHP that requests the source SVG from the file system, optimizes it and extends its attributes.

```twig
{{inlineSvg('path/to/svg', 'o-icon', {
  role: 'presentation'
})}}
```

### Pros
- Familiarity: We do something similar currently.
- SVGs are inlined on demand, so no need for intermediary compilation or transformation steps.

### Cons
- Requires a non-standard Twig extension.
- Even if we use [an external resource](https://github.com/clubstudioltd/craft-inline-svg) for this extension, we will need to duplicate its functionality across two languages. This increases maintenance time and fragility between the two repositories.

## Option 2: Custom Twig loader

Twig and the Twing library for JS both use a "loader" philosophy for resolving files. This means we can define custom logic for resolving things like includes.

[...]

PR #504, opened 2020-02-27
Basic icon CSS object

tylersticka:
## Overview

Introduces a very simple `o-icon` CSS class and example. I broke this piece off of #498 while I waited for confirmation that my Twig strategy (explained in that issue) was worth pursuing.

### Known Issues

- The example icon is hard-coded for now.
- I can probably clean up the inline example by putting it in its own `examples/inline.twig` template, but I didn't want to do too much more before revising this pattern once the rest of #498 is worked out.

## Screenshot

Screen Shot 2020-02-26 at 4 35 42 PM


src/objects/icon/icon.scss
```diff
@@ -0,0 +1,16 @@
+@use "sass:math";
+@use '../../mixins/svg';
+
+$inline-adjustment: 1rem / 16 * -2;
```
Paul-Hebert: It might be nice to include comments explaining this line and line 14.

tylersticka: @Paul-Hebert I ended up re-writing this file to eliminate the need for embedding this math. I also added more comments to hopefully make it easier to understand. Let me know if this helped!

Paul-Hebert: Yep, much clearer and I think those reusable utilities will come in handy going forwards! :+1:

PR #505, opened 2020-02-27
Refine color palette

### LINKED ISSUE #468: Refined color palette
Screen Shot 2020-02-19 at 4 54 31 PM

## Goal

Revise and expand our existing color palette to improve visual fidelity across devices, clarify usage and promote accessible contrast. This may be a good fit for a design tool like Sketch, Figma or Adobe XD?

This revised palette should be iterated on over time, so we should not let it paralyze our progress.

## Tasks

- [x] Define new alpha-transparent dark/black and light/white values for text on light or dark containers to differentiate from links and emphasized elements
- [x] Revise core blue
  - [x] Should have adequate text contrast on white background _and_ as background for light/white text (see above)
  - [x] Should appear less "dusty purple" on cheap monitors (could use a pinch more green?)
- [x] Revise accent colors
  - [x] Identify those suitable for text
  - [x] Identify those suitable for accents/illustrations
- [x] Include dark and light value variations for all colors in palette
- [x] Test all colors on devices in our lab
- [x] Organize our palette as a YAML file which may displayed in Storybook and consumed from Sass, etc. via [Theo](https://github.com/salesforce-ux/theo). (This may make more sense as a separate task.)

---

@cloudfour/design

dromo77:
This PR adds documentation for our refined color palette to storybook.

I chose to use numbers to indicate value changes for each color. As we continue to refine the palette I think this will be an easy way to make changes rather then using terms like dark-blue and light-blue. 

Something we may want to consider is using more descriptive terms to define the colors like `brand-primary` instead of `blue-3`. However, because we’re still experimenting with how many of these colors will be used yet, I kept it simple.

colors-storybook

- Fixes #468


src/design-tokens/design-tokens.stories.mdx
```diff
@@ -0,0 +1,221 @@
...
+</ColorPalette>
+
+## Extended Color Palette
+The extended palette is intended to be ever evolving, it offers an additional range of colors work with. Each color has 5 values, with 1 being the lightest and 5 being the darkest. The number of different values was determined by cataloging the current use of color throughout our site to create a sufficient number of variations. Usage of these colors will vary, but they should come in handy for illustrations and components.
+
+<ColorPalette>
+  <ColorItem
+    title="$blue-1"
+    subtitle="Accessible on the darkest colors in our palette."
+    colors={[brandTokens.blue1]}
```
tylersticka: I'm curious if you tried organizing these in an array from darkest to lightest? `{[brandTokens.blue1, brandtokens.blue2, ....]}`

I found this arrangement helpful from a design standpoint. We could highlight accessible color combinations in its own section?

dromo77: I did try that approach and I like that it condenses the information, however I didn't see a way to indicate the token for each color. Only the rgb values are visible beneath each swatch.

src/design-tokens/brand.yml
```diff
@@ -0,0 +1,139 @@
+imports:
+  - ./colors.yml
+props:
+  blue_1:
```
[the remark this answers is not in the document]
dromo77: A lot of the design systems I looked at use numbers (1 -10, 10 - 90, 100 - 900) to indicate steps in each color range. I'm open to other naming conventions if you have suggestions.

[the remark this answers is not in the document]
tylersticka: Some alternative ideas...

- `blue`, `blue-dark`, `blue-darker`, `blue-light`, `blue-lighter`
- Medium values are `0`, darker values are negative steps, positive values are lighter

I'd be open to other ideas from @cloudfour/design folks!

src/design-tokens/design-tokens.stories.mdx
tylersticka: You could potentially set `title` to `blue-*` and subtitle to `Steps 1 through 5` (or whatever convention we go with). But this isn't a blocker.

dromo77:
@tylersticka I made some changes based on your feedback, would you mind taking another look?

src/design-tokens/brand.yml
```diff
@@ -0,0 +1,139 @@
...
+    type: color
+    category: text-color
+  text_light_emphasis:
+    value: "{!white}"
+    type: color
+    category: text-color
+  text_action:
+    value: "{!blue}"
+    type: color
+    category: text-color
```
tylersticka: One other thought: Should all these tokens be in the same file? Is there value in having them separate. (I did that just to test that it worked!)

src/design-tokens/colors.yml
```diff
@@ -0,0 +1,171 @@
...
+  orange: "#f27041"
+  orange-dark: "#aa3f24"
+  orange-darker: "#622018"
+  gray-lighter: "#dee7ec"
+  gray-light: "#a5b9c6"
+  gray: "#738da0"
+  gray-dark: "#485968"
+  gray-darker: "#252b32"
+  black-transparent-85: "rgba(0, 0, 0, 0.85)"
+  white-transparent-85: "rgba(255, 255, 255, 0.85)"
```
tylersticka: I think we should be using underscores in this section for consistency.

src/design-tokens/colors.yml
```diff
@@ -0,0 +1,171 @@
...
+  primary_brand_dark:
+    value: "{!blue-dark}"
+    type: color
+    category: brand
+  primary_brand_darker:
+    value: "{!blue-darker}"
+    type: color
+    category: brand
+  green_lighter:
+    value: "{!green-lighter}"
```
tylersticka: Do you know if you can use token values as aliases? It feels a little funny having these defined twice, didn't know if that's a limitation of Theo or not.

dromo77: I'm not sure, maybe we don't need the aliases? We could replace the value directly with hex or rgba for each color instead of duplicating code.

tylersticka: I guess what I'm curious about is if this still works:

```
props:
  blue:
    value: "#XXXXXX"
    type: color
    category: example
  primary:
    value: "{!blue}"
    type: color
    category: example
```

If it does _not_, then the duplication is probably a good idea. If it _does_, then it probably isn't necessary.

dromo77: I tested this out and it looks like the aliases are necessary.

dromo77:
@tylersticka I updated the colors to use underscores, would you mind taking a look one more time?

PR #508, opened 2020-03-02
Templatize SVG assets

### LINKED ISSUE #498: Icon assets + tokens + pattern

tylersticka:
## Overview

This PR resolves #498 by creating `.twig` files from all SVG assets in the `src/assets` directory.

### Features

- SVG files may be included as you would any other Twig file, via `include`, `embed`, `extend`, etc.
- Eleven root attributes of each SVG is now dynamic, allowing you to override things like `class` or `aria-*`.
- Each SVG has three blocks added: `before`, `after` and `content`. This lets you easily extend and embed SVGs with prepended or appended content. You can even print _just_ the contents of any SVG via `{{block('content', 'path/to/asset.svg.twig')}}`.
- SVGO settings are shared between this task and Storybook's icon loader.

### Known Issues

- The task only runs when you start the npm script (no watching).
- To expose the templates to Storybook in an intuitive way, they need to share the same directories as other `src` files. It felt odd to version these files, though, so I added a nested `.gitignore` file that ignores `.twig` files in `src/assets`.
- It still isn't possible to pass an `include` path dynamically via Storybook, preventing a truly cross-platform `icon.twig` component.


gulpfile.js
```diff
@@ -0,0 +1,109 @@
...
+  // Identify props already in use in the SVG versus those yet to be used
+  const usedProps = dynamicSvgProps.filter(prop => Boolean(svg.attrs[prop]));
+  const unusedProps = dynamicSvgProps.filter(prop => !svg.attrs[prop]);
+
+  // Properties already in use should have their value set to a conditional.
+  // The `default` filter would be less code, but things get tricky when it
+  // comes to managing quotation marks in XML.
+  usedProps.forEach(prop => {
+    const current = svg.attrs[prop];
+    const twigProp = prop.replace(/-/g, '_');
```
Paul-Hebert: Might be nice to add a comment explaining why we're performing this transform. Is this just to match our twig conventions?

gulpfile.js
```diff
@@ -0,0 +1,109 @@
...
+    svg.attrs[
+      prop
+    ] = `{% if ${twigProp} %}{{${twigProp}}}{% else %}${current}{% endif %}`;
+  });
```
Paul-Hebert: This may be overkill, but I wonder if some props like class should be set up to append content that's passed in instead of replacing old prop content.

tylersticka:
> I don't 100% understand the syntax in the gulpfile, but I checked out the branch and confirm it builds and runs as described.

@spaceninja Is there additional commentary I could provide that would make it more clear?

gulpfile.js
Paul-Hebert: (I'm fine punting on this)

gulpfile.js
```diff
@@ -0,0 +1,109 @@
...
+            file.contents = Buffer.from(template);
+          }
+
+          cb(null, file);
+        })
+      )
+      // Append `.twig` to filenames
+      .pipe(rename({ extname: '.svg.twig' }))
+      // Output to same directory to expose to Storybook
+      .pipe(dest('src/assets'))
```
Paul-Hebert: I'm not super familiar with storybook so I don't know if this suggestion is feasible.

Could we move these to `src/assets/compiled` or something, so we have one directory housing all of our gitignored twig files instead of having them intermingled with non-ignored files?

gulpfile.js
tylersticka: @Paul-Hebert We had this ability in another project, but the more I thought about it the more I wondered if it was really necessary. Eliminating this feature greatly reduced the code this takes, so my instinct was to proceed without it for now.

Paul-Hebert:
>The task only runs when you start the npm script (no watching).

Do we plan to tackle this as a follow-up issue or is this a technical limitation we'll have to live with?

gulpfile.js
Paul-Hebert: Cool, that works for me :+1:

Paul-Hebert:
This is looking great @tylersticka ! I left a few questions and comments inline but besides those I think this is ready to 🚢 !

gulpfile.js
tylersticka: @Paul-Hebert It's feasible, but since we need to use relative paths (to keep the templates usable both in the pattern library in our theme), I'm a little sensitive to adding another path segment to all of these that we'll have to type _and_ remember. I'm open to feedback on that, though.

Paul-Hebert: Cool, that's fine by me :+1:

tylersticka:
> Do we plan to tackle this as a follow-up issue or is this a technical limitation we'll have to live with?

@Paul-Hebert Thankfully not a technical limitation! It just felt a little premature to start figuring out how to add file-watchers _in addition to_ Storybook. Once we understand how this will integrate with the back-end, I'm hoping that will give us an idea of what `dist` and the `postinstall` process looks like, which will make the full array of dependencies a bit more clear.

Paul-Hebert:
Sweet, sounds good @tylersticka

gulpfile.js
tylersticka: @Paul-Hebert Comment added!

PR #506, opened 2020-02-27
Add variable fonts

calebeby:
Closes #469 

https://deploy-preview-506--cloudfour-patterns.netlify.com/?path=/story/demo-button--button-element

## Testing

- Apply different font weights and styles to the button using devtools to verify that the Source Sans Pro variable font is working correctly
- Change the button to use "Source Code Pro" and verify that it works (this is not a variable font)

Test in IE11 to see if the Source Sans Pro non-variable font fallbacks are loaded

PR #515, opened 2020-03-04
Fix lint errors

tylersticka:
## Overview

Fixes lint errors introduced between #506 and #507 being merged.

## Testing

Verify status checks pass.

### DIFF
--- a/src/demo/button/button.scss
+++ b/src/demo/button/button.scss
@@ -3,11 +3,11 @@
 
 .demo-button {
   color: colors.$primary-brand;
-  transition: transform motion.$speed-quick motion.$ease-out;
 
   // TODO: These font styles are temporary, for demo purposes
   font-family: inherit;
   font-size: 18px;
+  transition: transform motion.$speed-quick motion.$ease-out;
 }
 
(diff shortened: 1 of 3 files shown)


src/demo/button/button.scss
```diff
@@ -3,11 +3,11 @@
 
 .demo-button {
   color: colors.$primary-brand;
-  transition: transform motion.$speed-quick motion.$ease-out;
 
   // TODO: These font styles are temporary, for demo purposes
   font-family: inherit;
   font-size: 18px;
+  transition: transform motion.$speed-quick motion.$ease-out;
```
spaceninja: Is transition temporary? If not, I'd recommend changing the temporary comment to a footnote style to make it clearer which lines it applies to.

tylersticka: Everything in `demo` is temporary.

spaceninja: If it's temporary, then you're good to go. :)

spaceninja: I just wanted to make sure that rule moving below the comment wasn't inadvertently confusing.

Issue #475, opened 2020-02-20
Button Pattern


dromo77:
Create a button pattern. We currently use several different buttons that vary in color, width, and use of iconography.

<img width="642" alt="buttons" src="https://user-images.githubusercontent.com/42841342/74987773-db04c200-53f0-11ea-8fc0-9679e287cf10.png">

tylersticka:
This pattern should coordinate with other form patterns to make sure inputs and buttons look good together.

Other considerations:

- Let's really showcase our ability to [design excellent, accessible button states](https://cloudfour.com/thinks/designing-button-states/)!
- If we are worried our focus states are too insistent, we should [consider a `focus-visible` polyfill](https://github.com/WICG/focus-visible).
- Can we use `filter` to simplify hover/active states and `transform` to give them a bit of visual delight?
- Let's consider re-balancing our border widths relative to our font weight.

Issue #470, opened 2020-02-20
Refined typography


tylersticka:
<img width="494" alt="Screen Shot 2020-02-19 at 5 23 26 PM" src="https://user-images.githubusercontent.com/69633/74891728-8ef84580-533c-11ea-856a-ec77053fa51f.png">

## Goal

In a world where variable fonts (#469) afford us more weights than we could have had before, can we _push_ sizes, weights, etc. of headings to feel more interesting and dynamic, particularly in page "heros" and article/page content?

It is _encouraged_ to refine these ideas as we continue!

## Tasks

- [x] Do design explorations of heading revisions using existing site content as a starting point
  - [x] Consider mobile-first and larger sizes
  - [x] Consider heading levels 1 through 6
  - [x] Consider ideal reading `font-size` responsively and associated line height, max content width
  - [x] Consider blockquotes, asides, introductory copy, meta copy (author info, publication date, etc.)
- [x] Determine what logic ([modular scale](https://www.modularscale.com/) or otherwise) might define type sizes relative to one another
- [x] Translate these ideas into a structure that could be interpreted by [Theo](https://github.com/salesforce-ux/theo) and used in Sass, etc.

tylersticka:
Now that #545 is merged and #547 is drafted, I feel safe calling this issue closed.

PR #516, opened 2020-03-04
Basic size and scale tokens

tylersticka:
## Overview

Introduces design tokens for modular scale and misc. size values. Also introduces a modular scale function for Sass. I found cases where these would be useful while working on #475 and #470.

There are a handful of existing modular scale libraries for Sass, but none of them seem to have been written since Sass added built-in support for exponential math. I found I was able to recreate 99% of their functionality in a single line.

The documentation pages are a work in progress but at least make the existence of the tokens known.

## Screenshots

Screen Shot 2020-03-04 at 9 18 10 AM

## Testing

1. View deploy preview
1. Observe tweaks to demo button using values from new tokens
1. Observe existence of "Design Tokens/Size and Scale" page


src/design-tokens/modular-scale.yml
```diff
@@ -0,0 +1,23 @@
...
+  octave: 2
+  major_tenth: 2.5
+  major_eleventh: 2.667
+  major_twelfth: 3
+  double_octave: 4
+props:
+  ratio:
+    value: "{!major_third}"
+    type: float
+    category: modular_scale
```
Paul-Hebert: I was confused by this file at first. I thought it was housing our actual modular scale values as opposed to the ratios. I wonder if renaming this to something like `modular-scale-ratios` would be clearer? Unless we plan to add more MS stuff here?

src/mixins/_ms.scss
```diff
@@ -0,0 +1,7 @@
+@use "sass:math";
+@use "../design-tokens/modular-scale.yml";
+
+@function step($step, $base: 1em, $ratio: modular-scale.$ratio) {
```
Paul-Hebert: Nit: Should the base value be a design token as well?

src/design-tokens/modular-scale.yml
tylersticka: There are infinite possible modular scale values based on the ratio, so I'm unsure what else we would need to store? 🤔

src/mixins/_ms.scss
tylersticka: Why?

src/design-tokens/sizes.stories.mdx
```diff
@@ -0,0 +1,44 @@
+import { Meta } from '@storybook/addon-docs/blocks';
+import * as msTokens from './modular-scale.yml';
+import * as sizeTokens from './sizes.yml';
+let msStepRows = [];
+for (let i = -4; i <= 10; i++) {
```
Paul-Hebert: We could also potentially make these -4 and 10 design tokens, though I don't know that we need to

tylersticka: Why?

I'm not trying to be argumentative, I'm just trying to understand the use case.

Paul-Hebert: Thinking about this more I don't think there's a very compelling use case.

I was thinking it could help to document the range of modular scale values devs should use. However, it already displays these values in the docs, and these numbers don't affect the SCSS, so I don't see a great reason to make these design tokens.

src/mixins/_ms.scss
tylersticka: I see this as more the base of the current value output. Most of the time you will want to output `em`, but sometimes you may want to output `rem`, etc.

src/design-tokens/modular-scale.yml
Paul-Hebert: I was just confused reading through this the first time since I misunderstood the purpose of this file. I thought it was going to house the scale values as opposed to the scale ratios. I thought tweaking the file name might make this clearer, but I don't think it's a big deal either way.

src/mixins/_ms.scss
Paul-Hebert: Oh yeah, that's clever :+1:

src/design-tokens/modular-scale.yml
tylersticka: Thanks for the clarification.

I think I'll keep it as-is to avoid having a `modular-scale-ratios.ratio` variable, which feels redundant to me.

PR #519, opened 2020-03-04
Basic Button pattern

### LINKED ISSUE #475: Button Pattern

tylersticka:
## Overview

Adds an `o-button` CSS/Twig object, with hover, active, focus and disabled states. Also includes `--secondary` and `--tertiary` modifiers.

I added [the focus-visible polyfill](https://github.com/WICG/focus-visible) because I enjoy the freedom it gives me to design _big_ focus states without fear of distracting mouse/touch users.

I decided to rename `examples` directories to `demo` because I got tired of typing "examples" over and over.

This might feel a little premature to tackle, but I wanted to start writing fundamental patterns soon so we can review and scare out any issues with our approach sooner rather than later. So please feel free to nitpick this PR!

### Known Issues

- This does not include buttons optimized for dark backgrounds. I started adding these but hit a bit of a rabbit hole getting a documentation page to show some previews with dark backgrounds, which made me think I should stop and make sure what I have so far is decent.
- It also does not include color variations like purple, pink, green, red, etc. It felt a little premature to do that right now.
- I think there will be an opportunity to share certain sizes or preferences with other form patterns, but I felt I would be _guessing_ at what form that should take until we started making those.

## Screenshots

Screenshot_2020-03-04 Objects Button - Elements ⋅ Storybook

## Testing

1. View deploy preview
1. Confirm focus state displays via keyboard but not via mouse/touch
1. Play with knobs in "canvas" view, confirm they work as expected
1. Confirm accessibility tab in "canvas" view passes

---

- Fixes #475


src/objects/button/button.scss
```diff
@@ -0,0 +1,81 @@
...
+  background: colors.$primary-brand;
+  border-color: colors.$primary-brand-dark;
+  border-radius: sizes.$radius-medium;
+  border-style: solid;
+  border-width: sizes.$edge-medium;
+  color: colors.$text-light-emphasis;
+  cursor: pointer; /* 1 */
+  display: inline-block;
+  font: inherit;
+  font-weight: 600;
```
spaceninja: Feel free to ignore me if it's just too early, but in the long run, I would expect font-weight to also consume a design token.

tylersticka: Agreed!

tylersticka:
> The scaling animation looks really bad on my non-retina monitor. The effect is that the button blurs on hover while growing, and then snaps back into focus when the animation is complete.

@spaceninja Hmm, I can take a look. What browser are you using, and does the problem persist when you are _only_ outputting to a 1x monitor or is it symptomatic when you have one 2x display (for example, your MacBook display) and a second 50%-scaled display?

Paul-Hebert:
> The disabled state doesn't do anything to link buttons. I'm not sure what we could or should do about that. Perhaps just a documentation note? Or would we want to add a .disabled class as well as the attribute?

A couple thoughts on this: 

1. Does it ever make sense to disable a link? I guess for a wizardy flow we could prevent them from accessing the next page or something? Still seems a little odd to disable a link. 
2. We could do something with `pointer-events: none;` though I wonder about the a11y of this. Would screen readers still be able to view and click the link? Would they signal it was disabled?

Paul-Hebert:
Would it make sense for the button modifier classes to show up as knobs in the canvas?

src/objects/button/button.scss
```diff
@@ -0,0 +1,81 @@
...
+.o-button {
+  background: colors.$primary-brand;
+  border-color: colors.$primary-brand-dark;
+  border-radius: sizes.$radius-medium;
+  border-style: solid;
+  border-width: sizes.$edge-medium;
+  color: colors.$text-light-emphasis;
+  cursor: pointer; /* 1 */
+  display: inline-block;
+  font: inherit;
```
Paul-Hebert: If we're inheriting `font` should we explicitly set `font-style` to `normal`? As is, if a button was in an italic block for some reason it would have italic text

Paul-Hebert:
I added a couple nits, but I agree with @spaceninja : this is great!

The buttons are beautiful. I like the interaction states a lot. Also storybook seems like a really nice dev environment looking at this PR. Excited to get in there and try it out!

tylersticka:
I haven't actually been able to reproduce _blurry_ button appearances, but _regardless_ of monitor I have discovered that the scale animations look pretty terrible in Chrome.

I've tried a bunch of fruitless suggestions on Stack Overflow and elsewhere, but the only thing that's worked to alleviate so far is setting various values to hard pixels.

What's more perplexing: It's not _just_ a matter of `em` values computing to round pixel values. Even layout values specified in `em` units that calculate to hard pixel values appear to disrupt the smooth scaling.

I need to do more research and try some stuff… 🤔

tylersticka:
Oh weird, the animations in Chrome look a _lot_ better when the Storybook examples are no longer centered… 🤯

Here's how those animations look like when centered (notice the edge shifts, especially the bottom edge):

chrome-hover-centered

Here's how those animations look when centering is disabled:

chrome-hover-topleft

That said, this may _not_ be related to @spaceninja's blurry issue, which I still haven't reproduced. But if someone _can_ reproduce it and wants to test the effect of centering on it, here are steps:

1. Check out this branch locally
1. Comment out line 9 (`addDecorator(centered)`) of `.storybook/preview.js`

spaceninja:
non-retina-hover

This was recorded on my Asus non-retina monitor, connected to my retina macbook, using Chrome v80.

tylersticka:
@spaceninja I honestly think this might look _better_ than in Chrome on a retina display! 😅

Paul-Hebert:
>I've tried a bunch of fruitless suggestions on Stack Overflow and elsewhere, but the only thing that's worked to alleviate so far is setting various values to hard pixels.

I'm guessing you've already tried this, but these are the hacks I usually try when facing this kind of thing. I think these all aim to get the GPU to help out with the transition, but I honestly don't understand them all that well:

1. `will-change`
2. `backface-visibility`
3. `translateZ(0)`
4. `translate3d(0, 0, 0)`

I think `will-change` is the modern, "good" one, but I've had to use the others in the past.

tylersticka:
@Paul-Hebert I tried `will-change`, `backface-visibility` and various font-smoothing type properties. I didn't try the 3D transform solutions because I don't think this effect is worth engaging the GPU and draining batteries faster... I'd sooner remove it, honestly.

After seeing @spaceninja's GIF and the lessened animation weirdness when the container isn't centered, I'm leaning toward moving on for now. I think we need to see more of the button working in context to know the best path forward. Heck, we might not even _like_ the scale animation in the context of a full-on interface!

Paul-Hebert:
Cool, sounds good to me @tylersticka

spaceninja:
That makes sense to me, and that's why I approved earlier — this is definitely an edge case. Worth noting now and being aware of if we stick with scaling buttons moving forward. But the _real_ point of this PR is to test the Storybook system, and in that regards, it's a wild success. Thanks @tylersticka!

PR #524, opened 2020-03-05
Dark theme + dark-optimized buttons

tylersticka:
## Overview

This PR started because I wanted to bring buttons optimized for dark backgrounds into the library. Originally, I planned to use a similar technique as what we do [in our current pattern](https://cloudfour-patterns.netlify.com/patterns/components/button.html#dark-color-modifiers), where buttons would need a different modifier depending on the background they were set against.

After [a brief conversation on Twitter about naming conventions](https://twitter.com/tylersticka/status/1235630911313477632), I realized that it was more intuitive to define a _theme_ which button styles would then respond to. One of the benefits of [BEMIT](https://csswizardry.com/2015/08/bemit-taking-the-bem-naming-convention-a-step-further/) is that it includes a few more [namespaces](https://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/), including themes.

From there, I saw two different ways of approaching this, both with pros and cons.

The first method would involve CSS custom properties. It might allow me to inherit certain property values from the theme itself, and could also present some chances to decrease specificity by overriding values rather than entire properties. But in practice, the inheritance potential seemed limited (the buttons are far more complex than the theme, and based on our current site I don't know that that will change), and it was unclear how it would work in IE11.

The second method is just some old school _cascading_ styles, which is what I went with. It isn't fancy, but it's early enough we could always revisit.

### Known Issues

- The Storybook Backgrounds add-on doesn't really work with `t-dark`, and I haven't researched into whether or not I can _extend_ it at all.
- Even if it did work, it would only apply to the canvas (not the docs). So I had to make a somewhat hack-y little theme container. It's possible that in the future this could be simplified with a custom decorator.
- This is just cosmetic, but I sure would like the theme background in the docks page to meet the edges!

## Screenshots

Screen Shot 2020-03-05 at 2 46 43 PM

Screen Shot 2020-03-05 at 2 43 27 PM


src/objects/button/button.stories.mdx
```diff
@@ -35,3 +37,15 @@ Some buttons may not require as much emphasis as others. The `o-button--secondar
     })}
   </Story>
 </Preview>
+
+All visual styles are optimized for [our dark theme](/?path=/docs/themes--dark). We avoid simply inverting the visual appearance so buttons will appear distinct from white content containers or text inputs.
```
Paul-Hebert: >We avoid simply inverting the visual appearance so buttons will appear distinct from white content containers or text inputs.

For some reason my brain struggles to tell if "buttons will appear distinct from white content containers or text inputs" happens when we do or don't invert.

Maybe "Instead of simply inverting their appearance we have different dark theme styles so our buttons appear distinct from white content containers or text inputs."

Might just be me though.

src/themes/_dark.scss
```diff
@@ -0,0 +1,16 @@
...
+
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6,
+  a {
+    color: colors.$text-light-emphasis;
+  }
```
Paul-Hebert: Potentially premature nits:

- I wonder if links should still have a color distinguishing them or if underlines are enough?
- Are there other elements that need to change colors? `blockquote`? `li`?

(I'm fine punting on this until we need to handle it but figured I'd mention it.)

Paul-Hebert: Re: links I just looked at the screenshots again and I think they look great white. Feel free to ignore that part.

src/themes/demo/fill-canvas.twig
```diff
@@ -0,0 +1,16 @@
+<div{% if theme %} class="t-{{theme}}"{% endif %} style="
```
Paul-Hebert: Might be nice to incorporate some of the PR context into a comment here for future devs

src/objects/button/button.stories.mdx
tylersticka: I re-wrote this paragraph, it was very clunky before.

src/themes/_dark.scss
tylersticka: I see this as just a starting point!

src/themes/demo/fill-canvas.twig
tylersticka: I've added a comment!

PR #530, opened 2020-03-07
Add stylelint-config-prettier

### LINKED ISSUE #518: Improve SCSS linting
This is one of our first projects since we've started [feeling Sassy again](https://cloudfour.com/thinks/feeling-sassy-again/). Our Stylelint config (see #507) is having some odd effects on our Sass. For example, it seems to require a line break before any @-rule, so if we `@include` a mixin at the beginning of a class it's just adding a random line break beforehand.

spaceninja:
## Overview

This commit adds `stylelint-config-prettier`, which disables Stylelint rules that conflict with Prettier. It also updates our `package.json` lint scripts to run Prettier as a separate task.

## Testing

1. check out the branch and try running the lint scripts from the command line

---

- Fixes #518


spaceninja:
Note that linting is currently failing, which weirdly, is a good thing, because it shows that I didn't break the linting setup with this change. I'll make a followup PR with lint fixes after this is merged.

spaceninja:
Oh you're right! Hah, okay well, I'll do lint fixes on Monday, then.

calebeby:
@spaceninja should stylelint-config-prettier be integrated into the cloud four stylelint config rather than this repo?

spaceninja:
@calebeby My gut reaction is no. Not everyone likes Prettier for CSS files. I updated the readme to include instructions on how to work with stylelint-config-prettier, which should give people using our config (including us) the most flexibility. Holler if you think I'm missing something, though.

spaceninja:
@tylersticka this is ready for re-review. It now includes lint fixes, and updated our linters to ignore stuff in the `.gitignore` file, which avoids the problem I was running into where the linters were trying to lint the `dist` and `storybook-static` folders.

PR #537, opened 2020-03-09
Add MDX linting

calebeby:
## Overview

`npm run lint`

You can verify that prettier formatting is working by messing up the formatting in the jsx in the .mdx files, and making sure prettier puts it back.

You can verify that eslint checking is working by adding this to one of the JS sections of the .mdx files:

```js
let foo = bar;
```

ESLint should flag a few issues with that line.

You can verify that remark checking is working by adding this to one of the markdown sections of the .mdx files:

https://example.com/

Remark should flag that it should be surrounded with `<>` because not all markdown implementations auto-link things that look like links. It should also auto-fix it if you run `npm run lint`.


Paul-Hebert:
Thanks @calebeby !

I can see that Prettier is formatting my file. I also see that running `npm lint` will complain about JS errors and fix markdown errors.

---

Is this also meant to visually flag errors when I'm editing a file? (like a red underline?) If so I'm not seeing that when I test. If that's not part of this effort then it all looks good to me.

calebeby:
@Paul-Hebert It should show squiggles in your editor. Do you have the eslint VSCode plugin installed?

Paul-Hebert:
@calebeby ,

Hmm, yeah I believe I do. This one, right?

Screen Shot 2020-03-09 at 4 14 57 PM

This is what it looks like when I'm editing a file:

Screen Shot 2020-03-09 at 4 15 42 PM

I have the file language set to `Markdown React (MDX)` in the editor:

Screen Shot 2020-03-09 at 4 18 14 PM

I've also done an `npm ci`. 

I'm not sure why I'm not getting the squiggles. Any ideas?

calebeby:
I think in order to get squiggles you have to tell VSCode's ESLint plugin to allow checking .mdx files. Here is the relevant part of my vscode settings:

```json
  "eslint.validate": [
    "javascript",
    "javascriptreact",
    "typescript",
    "typescriptreact",
    "mdx"
  ],
```

Paul-Hebert:
Oh yeah, that does the trick, thanks @calebeby ! This may be a separate discussion but I wonder whether we should check in a `.vscode/settings.json` file? Or perhaps check it in as `.vscode/settings.json.default` so devs could use it or override it?

That doesn't need to be decided here though. This looks good to me!

src/themes/themes.stories.mdx
```diff
@@ -14,8 +14,13 @@ While selector depth is best kept to a minimum, it's permissable to style object
 
 `t-dark` is intended primarily for page headers.
 
+export const parameters = { themes: { disable: true } };
+// Parameters has to be defined separately because of this prettier bug:
+// https://github.com/prettier/prettier/issues/7731
+// https://mdxjs.com/getting-started#defining-variables-with-exports
```
calebeby: @Paul-Hebert I wanted to point out this change in case you didn't see it, since it was not an automated change

spaceninja:
Nice work on this, Caleb! I copied the setup to my side project Storybook, and it worked like a charm!

src/themes/themes.stories.mdx
Paul-Hebert: Cool, thanks. That seems fine.

calebeby:
Ready for re-review!

calebeby:
@tylersticka Hmm, I would have thought that would fail the build 🤔

calebeby:
@tylersticka I switched to using prettier from master on github because that fixed 2 bugs that it had when formatting mdx files.

Issue #466, opened 2020-02-20
Define class naming convention


tylersticka:
The current iteration of our pattern library uses [SUIT CSS](https://suitcss.github.io/) for class names. We like SUIT because:

- The components are independent of one another and easy to install as needed.
- The class naming convention was easier to grasp for our customers, who found double punctuation in traditional BEM off-putting.
- The distinction between utility classes and non-utilities (`u-`) felt useful.
- The percentage based sizing classes felt in alignment with our ways of thinking about things.

Having used the convention across multiple projects in the last four years, we've encountered some issues with it:

- The distinction between descendant (`Parent-child`) and modifier (`Parent--modifier`) is sometimes too subtle.
- The breakpoint prefixing strategy for utilities (`u-md-size1of2`) has no equivalent for modifiers.
- Camel-case utilities are a bit less readable since they lack any segments after the prefix: `u-flexAlignContentBetween` versus `flex-align-content-between`
- We end up customizing components enough that I sometimes question their helpfulness once a project is far enough along.

With that in mind, here are some questions I'm interested in answering...

- [ ] Should we stick with SUIT-style (`ComponentName-child`) or BEM-style (`component-name__child`) formatting? (I go back and forth.)
- [ ] Should we stick with SUIT-style breakpoint prefixing (`u-md-textGrow1`) or use [BEMIT suffixes](https://csswizardry.com/2015/08/bemit-taking-the-bem-naming-convention-a-step-further/) (`u-textGrow1@md`) (I think this would be a positive change.)
- [ ] For classes that include steps (such as modular scale utilities), can we switch from a leading `0` for negative steps (`text-grow-01`) to the character `n` (`text-grow-n1`)? (I say yes!)

---

@cloudfour/design @cloudfour/dev

tylersticka:
Also ~dragging @spaceninja down with me~ assigning @spaceninja since I know [he cares about this stuff](https://spaceninja.com/2018/09/18/what-is-modular-css/).

spaceninja:
BWAHAHAHA YOU FOOLS SHOULD NEVER HAVE INVITED ME IN YOU'LL REGRET THE DAY YOU ASKED SCOTT FOR HIS OPINIONS ON CLASS NAMES!!!

Just kidding.

My primary concern when it comes to class naming conventions is that we _have a convention_. Whether we use SUIT or BEM is not crazy important to me. I personally prefer BEM over SUIT, but I will happily go with the overall decision the team makes.

[...]

PR #554, opened 2020-03-18
Add css selector bem linting

### LINKED ISSUE #517: Add CSS selector linting
In #466 we agreed on a new standard for CSS selectors.

In #507 we installed stylelint.

Our other projects have used [stylelint-selector-bem-pattern](https://github.com/simonsmith/stylelint-selector-bem-pattern) to help enforce selector conventions. It would be really nice if this project could also flag when selectors don't meet our convention.

calebeby:
#63  Overview

Closes #517 
I had to pass a lot of options to https://github.com/postcss/postcss-bem-linter in order for it to accept [our class naming convention](https://github.com/cloudfour/cloudfour.com-patterns/issues/466#issuecomment-589361857). There easily could have been things that I missed.

`postcss-bem-linter` does not want to accept non-component classes with prefixes like `.is-open`, `.t-dark`, etc. The only way I could figure out how to make it allow those was to ignore selectors with those entirely.

So it will give you an error if you do this:

```scss
// components/button.scss
.button {}
// ^ should be .u-button, will error
```
but it does not give you an error if you do this:
```scss
// components/button.scss
.button.is-open {}
```

because `.is-` is on the ignore-list for selectors. I could not find a better way to do this without forking `postcss-bem-linter`.

This also only lints selectors that are in `components` or `utilities`

## Testing

1. Create `components` and `utilities` folders
2. Add files to those folders
3. Add incorrect selectors and make sure it flags them
4. Add correct selectors and make sure it doesn't flag them

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,0 +1,45 @@
+module.exports = {
+  plugins: ['stylelint-selector-bem-pattern'],
+  extends: ['stylelint-config-cloudfour', 'stylelint-config-prettier'],
+  rules: {
+    'plugin/selector-bem-pattern': {
+      preset: 'bem',
+      /** These files are automatically assumed to have components */
+      implicitComponents: 'src/components/**/*.scss',
+      /** These files are automatically assumed to have utility selectors */
+      implicitUtilities: 'src/utilities/**/*.scss',
+      utilitySelectors: /^\.u-[a-z-]+(?:\\@[a-z]+)?$/,
+      // https://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces#the-namespaces
+      ignoreSelectors: [
+        /^\.t-/,
+        /^\/.s-/,
+        /\.is-/,
+        /\.has-/,
+        /\.js-/,
+        /\.o-/,
+        /\.qa-/,
+        /_/
+      ],
+      componentSelectors: {
+        initial: bemSelector,
+        combined: () => /.*/
+      }
+    }
+  }
+};
+
+// Copied from https://github.com/postcss/postcss-bem-linter/blob/master/lib/preset-patterns.js
+// Had to copy so that we can pass `initial` and `combined` separately,
+// otherwise `combined` defaults to `initial`
+
+/**
+ * @param {String} block
+ * @returns {RegExp}
+ */
+function bemSelector(block) {
+  const WORD = '[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)*';
+  const element = `(?:__${WORD})?`;
+  const modifier = `(?:(?:_|--)${WORD}){0,2}`;
+  const attribute = '(?:\\[.+\\])?';
+  return new RegExp(`^\\.c-${block}${element}${modifier}${attribute}$`);
+}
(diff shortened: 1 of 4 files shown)


tylersticka:
I plan to take a closer look at this as soon as I'm able to. I encourage others to provide feedback before then, but I'd prefer not to merge this till I've had a chance to dig in. Thanks!

tylersticka:
I think you've done the best job you can with this, @calebeby. But I'm wondering if the PostCSS BEM Linter just isn't ready for this convention yet.

In my ideal world, we'd be able to `/** @define objects */` or `/** @define themes */` and have the linter treat those namespaces with the same detail as `/** @define utilities */` or whatever. But as-is, we're sort of just linting two namespaces and ignoring everything else.

To be clear, I don't think this is your fault… it's clear from other issues like https://github.com/postcss/postcss-bem-linter/issues/117 that this is a known limitation. But I fear implementing a partial solution like this makes it _less_ intuitive to understand how conventions are enforced, and may give us a sense of false confidence depending on where the bulk of our styles end up living.

On a side note, I've messaged Harry Roberts on Twitter in case he's aware of existing tools for making this happen: https://twitter.com/tylersticka/status/1240658351819640832

calebeby:
Yeah, that's kind of where I landed too, it's really not something the library wants to do, what I have is not ideal.

If you want, I could fork it and add that functionality

tylersticka:
@calebeby My knee-jerk reaction to maintaining a fork of a BEM selector linting plugin is negative, unless it's comparable in complexity to something like our lint config repos.

_Or_ if we could propose it as a PR, I suppose that'd put it outside the realm of our responsibility.

I dunno, I'm torn. Would love @spaceninja's opinion on this.

spaceninja:
It depends on how open the maintainers are to us submitting a PR.

If it seems like there's a good chance we could do the work and get it merged into the main repo, then game on!

But if it seems like it would languish for a long time, it might be more hassle than it's worth.

FWIW, I've never worked on a project with class name enforcement in linting, it was just understood that contributors should follow BEM rules, and if/when someone slipped up, we'd catch it in code review. As a result, I'm personally not super invested in this aspect of linting, but if the team wants it or sees value, I'm not opposed. (Just to explain why I view the cost/benefit of maintaining a fork this way).

calebeby:
> FWIW, I've never worked on a project with class name enforcement in linting, it was just understood that contributors should follow BEM rules, and if/when someone slipped up, we'd catch it in code review.

I like this. I feel like this PR gets us 80% of the way there to get automated catches for this kind of thing, but implementing the ability for it to catch the remaining 20% would take a lot more time, so I don't think it is worth it. I think that human reviewers can catch those cases

tylersticka:
I've encountered many cases (both on internal and client-facing projects) where automated selector linting came in handy, or when it was added later and flagged dozens of missed issues. Based on those experiences, I'm not sure I'd agree that human review is an adequate substitute.

That said, I feel more confident in our ability to catch those issues manually than I do our ability to consistently maintain a linting plugin. So I think we should close this PR and revisit the issue at a later date.

Sound good?

calebeby:
I think that what we have in this PR is better than nothing. I think we should merge it. Like you said, later on we can revisit

tylersticka:
When I tried this PR locally, I personally found it confusing to have two namespaces linted while others are ignored.

calebeby:
Ok 🤷‍♂

Issue #556, opened 2020-03-18
Find code pattern for knob reuse


Paul-Hebert:
# The Problem

Storybook knobs allow us to pass props into code demos. This is really cool for allowing users to interact with code examples and try out different options! It works great for pattern modifiers, pattern content, and more.

I was recently working on a new Table pattern. I wanted to have several demos showing all the different modifiers and options. I wanted each demo to show all of the options. However, for each demo I wanted to hard code the value of a specific knob.

For example, for the Striped Table example, I wanted to hard code the `rowModifier` knob to `o-table--striped`. This was my first, quick and dirty implementation:

```mdx
# Table Object

The `o-table` class applies default styling to HTML `table` elements.
A table can optionally have headers and footers.

<Preview>
  <Story name="Basic">
    {basicDemo({
      hasHeader: boolean('Include Header', true),
      hasFooter: boolean('Include Footer', true),
      rowModifier: select('Table Row Modifier Class', [
        '',
        'o-table--ruled',
        'o-table--striped'
      ]),
      numericCells: boolean(
        'Add o-table__cell--numeric modifier to cells',
        false
      ),
      rightAlignRowHeadings: boolean('Right-align row Headings:', false)
    })}
  </Story>
</Preview>

# Striped Table

The `o-table--striped` modifier adds alternating background colors to table rows.

<Preview>
  <Story name="Striped">
    {basicDemo({
      hasHeader: boolean('Include Header', true),
      hasFooter: boolean('Include Footer', true),
      rowModifier: 'o-table--striped',
      numericCells: boolean(
        'Add o-table__cell--numeric modifier to cells',
        false
      ),
      rightAlignRowHeadings: boolean('Right-align row Headings:', false)
    })}
  </Story>
</Preview>
```

This quickly became ungainly as I added more examples and it seemed like there was an opportunity to DRY this up. Here's what I'd like to be able to do:

```mdx
// This function returns all of our table knobs
// By passing in custom knobs you can override the default knob values
// This is helpful for setting a knob for a specific demo.
const knobs = customKnobs => {
  const defaultKnobs = {
    hasHeader: boolean('Include Header', true),
    hasFooter: boolean('Include Footer', true),
    rowModifier: select('Table Row Modifier Class', [
      '',
      'o-table--ruled',
      'o-table--striped'
    ]),
    numericCells: boolean(
      'Add <code>o-table__cell--numeric</code> modifier to cells',
      false
    ),
[...]

PR #557, opened 2020-03-18
New Table Pattern

### LINKED ISSUE #491: Table pattern
Create a table pattern.

tables

Paul-Hebert:
## Overview

This PR creates and documents a new table pattern. The Table pattern has new styling, as well as a number of modifiers and options. 

As part of this work, I tweaked `gray-lighter` and `gray-lightest`. I'd appreciate design review on those color changes as well as the new pattern itself.

## Screenshots

Screen Shot 2020-03-18 at 12 35 12 PM
Screen Shot 2020-03-18 at 10 37 23 AM
Screen Shot 2020-03-18 at 10 37 28 AM
Screen Shot 2020-03-18 at 10 37 34 AM
Screen Shot 2020-03-18 at 10 37 40 AM

## Testing

1. Review the Table patterns for issues (canvas and docs)
2. Review the color docs for issues

## Known Issues

This code could be DRYed up a lot. I made an issue to dig into this: https://github.com/cloudfour/cloudfour.com-patterns/issues/556

Fixes #491


src/design-tokens/colors.yml
```diff
@@ -1,140 +1,140 @@
...
-  orange_darker: "#622018"
-  gray_lighter: "#dee7ec"
-  gray_light: "#a5b9c6"
-  gray: "#738da0"
-  gray_dark: "#485968"
-  gray_darker: "#252b32"
-  black_transparent_85: "rgba(0, 0, 0, 0.85)"
-  white_transparent_85: "rgba(255, 255, 255, 0.85)"
-  white: "#fff"
+  blue_lighter: '#8abfff'
```
Paul-Hebert: My code editor made this swap automatically. I'm guessing it's based on Prettier or linting rules? Happy to revert if preferred. The only actual changes are to `gray_light` and `gray_lighter`

src/design-tokens/colors.yml
```diff
@@ -1,140 +1,140 @@
...
+  fuchsia: '#d9118f'
+  fuchsia_dark: '#950f7d'
+  fuchsia_darker: '#5c0a5c'
+  orange_lighter: '#ffad5c'
+  orange_light: '#fa8f47'
+  orange: '#f27041'
+  orange_dark: '#aa3f24'
+  orange_darker: '#622018'
+  gray_lighter: '#f2f6f7'
+  gray_light: '#cedee8'
```
Paul-Hebert: The 2 lines above this comment changes.

src/design-tokens/colors.yml
calebeby: Yeah I think it is because of prettier. I think this change is fine 👍

tylersticka:
## Design Feedback

1. I can't quite articulate why, but there's something I'm not 100% satisfied with in the new gray values.
1. I think maybe `$gray-light` is now _too_ light? It doesn't feel distinct enough from the cell backgrounds as a divider color to me.
1. And maybe the `$gray-lighter` would benefit from a slight hue change (in inspector, changing the hsl color from `192` hue to `203` makes a difference to my eye). But very interested in if @dromo77 has an opinion on that.
1. I'm curious if you tried a lighter weight for the `th` styles? On desktop especially it's feeling maybe a bit inflated.

## Technical Feedback

1. I know _technically_ this contradicts our guidelines, but since tables have pretty specific markup requirements I'm curious if there's actually any value in requiring classes for descendants like `<td>` and `<th>`. It seems like it might simplify usage quite a bit to just require `Table`, unless we thought this would be used on elements that _aren't_ `<table>`.
1. There are a _lot_ of variations shown here. Was this informed by actual table usage on our site, or is this "just in case"? If it's the latter, I fear we're adopting too much up-front complexity.

src/objects/table/demo/basic.twig
```diff
@@ -0,0 +1,160 @@
...
+        <a href="https://github.com/cloudfour/transition-hidden-element/stargazers">34</a>
+      </td>
+      <td class="o-table__cell {% if numericCells %} o-table__cell--numeric {% endif %}">
+        <a href="https://github.com/cloudfour/transition-hidden-element/network/members">0</a>
+      </td>
+      <td class="o-table__cell {% if numericCells %} o-table__cell--numeric {% endif %}">34</td>
+      <td class="o-table__cell {% if numericCells %} o-table__cell--numeric {% endif %}">
+        <a href="https://github.com/cloudfour/transition-hidden-element/issues">0</a>
+      </td>
+    </tr>
```
tylersticka: This isn't a blocker, but if you want to avoid all this repetition in the future, I believe Storybook includes Webpack's JSON loader by default, so you could put a file in the `demo` directory and pipe it to your Twig template to iterate over.

src/objects/table/table.scss
```diff
@@ -0,0 +1,84 @@
...
+ *    tables are clipped.
+ * 3. Override default `th` text alignment
+ */
+
+.o-table {
+  border-collapse: collapse; /* 1 */
+  border-spacing: 0; /* 1 */
+  overflow-x: auto; /* 2 */
+  text-align: left; /* 3 */
+  width: 100%; /* 2 */
```
tylersticka: Honest question: Would any of these be good defaults for `<table>`?

src/objects/table/table.scss
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Table rows can be styled in a couple ways.
+ * This is achieved by adding a class to the table itself.
+ */
+.o-table--ruled .o-table__row {
+  border-bottom: $table-border-thin;
+}
+
+.o-table--striped .o-table__row:nth-of-type(2n - 1) {
+  background: colors.$gray-lighter;
```
tylersticka: Is it necessary to override all of `background`?

src/objects/table/table.scss
```diff
@@ -0,0 +1,84 @@
+@use "../../design-tokens/colors.yml";
+@use "../../design-tokens/font.yml";
+@use "../../design-tokens/sizes.yml";
+@use "../../mixins/ms";
+
+$table-border-color: colors.$gray-light;
+$table-border-thin: sizes.$edge_small solid $table-border-color;
+$table-border-thick: sizes.$edge_medium solid $table-border-color;
```
tylersticka: Honest question I asked myself (but didn't answer) for button: Should these values be stored in a design token file? If so, should that token live with the pattern or inside of tokens?

These are the sorts of questions that make me lie awake at night. 🙃

src/objects/table/table.scss
spaceninja: FWIW, tables are one of the things I've always struggled with in design systems. Bootstrap treats `.table` as an opt-in class, but then provides no default table styles, which always felt strange to me. But then I worked at a job where we actually had default table styles, and it was a hassle having to override them with modifier classes for variations.

I could go either way on treating table styles like a component (meaning fully respecting BEM class naming rules, even though I can't think of a situation where I'd use table classes without table markup) or just applying these as the based table styles (meaning don't use class names at all, but instead target the `table`, `th`, and `td` elements).

TL;DR: There's no "right" answer here, unless we know up-front that we'll have lots of table variations.

Paul-Hebert:
Thanks for all the feedback so far everyone! I'll circle back and respond tomorrow 🙂

src/objects/table/table.scss
tylersticka: > TL;DR: There's no "right" answer here, unless we know up-front that we'll have lots of table variations.

One benefit of this project is that we have 347 posts and 28 pages of existing content we can reference to see how we're using stuff!

That doesn't mean we should necessarily make decisions based on past behavior that make future changes harder, but it is a nice guiding light.

src/objects/table/demo/basic.twig
Paul-Hebert: Oh interesting I'll check that out

src/objects/table/table.scss
Paul-Hebert: No, good call. I can use background-color

src/objects/table/table.scss
Paul-Hebert: I'm not sure...

To be honest I don't have a ton of experience with design tokens.

From my perspective it would be useful to make these design tokens if we wanted to use them in multiple places. e.g. I want to use them for table styles and document them separately in storybook so future devs can use them in similar interfaces.

At this point I don't see a second use case so it seems like a premature optimization but I'm open to other opinions, especially if folks have more experience with design tokens.

src/objects/table/table.scss
Paul-Hebert: Hmmm, I could see the following rules being useful applied to tables in general, since these kinda feel like they should be the defaults:

```css
table {
   border-collapse: collapse;
   border-spacing: 0;
   width: 100%;
   text-align: left;
}
```

I'd be hesitant to apply the overflow or other rules globally. Do y'all think that makes sense?

Paul-Hebert:
@tylersticka, thanks for the in-depth feedback!

> I think maybe $gray-light is now too light? It doesn't feel distinct enough from the cell backgrounds as a divider color to me.
> And maybe the $gray-lighter would benefit from a slight hue change (in inspector, changing the hsl color from 192 hue to 203 makes a difference to my eye). But very interested in if @dromo77 has an opinion on that.

I'm fine exploring alternate colors here. I'll do some experimenting!

> I'm curious if you tried a lighter weight for the th styles? On desktop especially it's feeling maybe a bit inflated.

I can experiment with this as well!

> I know technically this contradicts our guidelines, but since tables have pretty specific markup requirements I'm curious if there's actually any value in requiring classes for descendants like <td> and <th>. It seems like it might simplify usage quite a bit to just require Table, unless we thought this would be used on elements that aren't <table>.

Based on other feedback it sounds like tables are a special case when it comes to BEM. It is a pain having to add a class to every single table cell. I'd be fine simplifying usage by targeting elements instead of classes within `.Table`

> There are a lot of variations shown here. Was this informed by actual table usage on our site, or is this "just in case"? If it's the latter, I fear we're adopting too much up-front complexity.

I did another review of the current variations:

**Headers and Footers**

Reviewing the site lots of tables have styled headers. Articles like [this one](https://cloudfour.com/thinks/sensible-jumps-in-responsive-image-file-sizes/#what-is-the-performance-budget-for-flexible-images) are using `strong` elements to fake table footers. Including the footer styles takes 3 lines of CSS.

**Ruled and Striped Tables**

Both of these variations exist and are in use on the current site.

**Numeric Table Cell Modifier**

We're not using this style anywhere on the site, but reviewing blog posts I found some where I think this modifier could make numeric data in tables easier to scan, compare and understand:

- https://cloudfour.com/thinks/responsive-images-101-part-9-image-breakpoints/
- https://cloudfour.com/thinks/responsive-images-101-part-7-type/
- https://cloudfour.com/thinks/responsive-images-101-part-3-srcset-display-density/

This article appears to be trying to fake this style in some places using `&nbsp;`s for alignment: 

- https://cloudfour.com/thinks/responsive-guide-to-type-sizing/

This modifier requires 4 lines of CSS.

**Tightened Row Headings**

We don't appear to be using row headings very often on the site. I've used similar formatting outside of C4 to make tables easier to scan and understand, but I didn't see any use cases for this on the current site. We could remove this, but at the same time it only takes 3 lines of CSS.

---

Overall I'm inclined to leave these modifiers in place. I could see an argument for removing the tightened row headings or numeric table cell modifiers, but they also only add up to 7 lines of CSS so I'm not sure how much technical complexity they add.

src/objects/table/table.scss
Paul-Hebert: Done!

src/objects/table/table.scss
Paul-Hebert: The rules mentioned above are now applied to `<table>`s globally.

tylersticka:
@Paul-Hebert One more piece of feedback: Based on the outcome of #558, this pattern would be a _component_ (not an object).

src/objects/table/demo/basic.twig
Paul-Hebert: Done! Let me know if there's a more storybooky way to do this!

Paul-Hebert:
Hey @tylersticka , @spaceninja , @calebeby , everybody else!

I made some changes to this PR and would appreciate another look when you have time:
- The CSS uses `background-color` instead of `background`
- Some table styles are set on `<table>` instead of `.c-table`
- `.c-table` child elements (`td`, `th`, `tfoot`, `tbody`) get styles without needing classes
- Tables are components, not objects
- The demo consumes a JSON file to make maintenance easier.
- `th`s are semibold, not bold
- `gray-lighter`'s hue was adjusted
- `gray-light` was darkened slightly.

There are a few open discussions that I've responded to.
- Should the SCSS table vars be design tokens?
- Are all these modifiers worth including?

Let me know your thoughts on all of the above. Thanks!

src/base/_defaults.scss
```diff
@@ -139,3 +139,14 @@ nav,
 section {
   display: block;
 }
+
+/**
+ * Apply some sensible default to tables
```
tylersticka: Nit: "defaults"?

src/components/table/table.scss
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * 1. Tables may overflow horizontally. Keep this from breaking layouts while
+ *    ensuring content is viewable
+ *    In the future we may want to consider making it visually obvious when
+ *    tables are clipped.
+ */
+
+.c-table {
+  overflow-x: auto; /* 1 */
```
tylersticka: Any reason why we wouldn't want this on our default `<table>` elements?

src/components/table/table.scss
```diff
@@ -0,0 +1,86 @@
...
+}
+
+/**
+ * Optionally right align row headings.
+ * This keeps headings closer to the content they label and in some cases
+ * improves scannability.
+ */
+.c-table--tighten-row-headings [scope='row'] {
+  text-align: right;
+}
```
tylersticka: This might be crazy, but do you think there's a way to apply a modifier to `c-table` that applies this change to `th[scope='row']` elements and the `numeric-cell` modifier to `td` elements?

src/components/table/table.stories.mdx
```diff
@@ -0,0 +1,121 @@
...
+<Preview>
+  <Story name="Basic">
+    {basicDemo({
+      hasHeader: boolean('Include Header', true),
+      hasFooter: boolean('Include Footer', true),
+      rowModifier: select('Table Row Modifier Class', [
+        '',
+        'c-table--ruled',
+        'c-table--striped'
+      ]),
```
tylersticka: As written, this makes it seem like these modifiers can't be combined. I don't believe that's the case: Should these be their own distinct options instead of a `<select>` element?

src/components/table/table.scss
Paul-Hebert: I wondered about this as well. My concern was that your table may have some columns that are numeric data and others that are strings (so shouldn't be right aligned or have tabular numbers). I guess it might be good to have the option to either apply this at the cell level or on the table level?

src/base/_defaults.scss
Paul-Hebert: Good catch. Fixed

src/components/table/table.stories.mdx
Paul-Hebert: Good call. I was thinking that might feel heavey-handed but I tried it out and it looked okay. I'll update this to allow both.

src/components/table/table.scss
tylersticka: I'm just trying to think of adding these via a CMS. If the feature is available as a top-level modifier, the WordPress Gutenberg table block can support it (every Gutenberg block has a "class" field). But if it's only available on cells, then it will require editing the block as HTML, writing a custom Gutenberg block or manually coding the table.

src/components/table/table.stories.mdx
Paul-Hebert: Done!

src/components/table/table.scss
Paul-Hebert: Hmm, I suppose not... I can move that over

Paul-Hebert: done!

src/components/table/table.scss
Paul-Hebert: Yeah that's a good point. I hadn't been considering it from that perspective. I'd been imagining populating via JSON or hand-building for blog posts but you're right that it would be nice to be able to use this with the Gutenberg block building UI. I'll look into combining these and applying the class at the table level.

Paul-Hebert:
Thanks for another good round of feedback @tylersticka !

I made the following updates:

- Add support and styling for a table caption
- Fixed a typo
- Combined and simplified the usage of the right aligned row headings and numeric table data
- Allow combining the striped and ruled modifiers
- Fixed a typo
- Moved an overflow styling rule from `.c-table` to `<table`>

If you could take another look when you have a chance I'd appreciate it. Thanks!

PR #565, opened 2020-03-24
Chore: Refactor Button & Icon from "objects" to "components"

### LINKED ISSUE #558: Do BEMIT right
@calebeby asked me a very simple question in a PR I can't be bothered to look up again right now. I'm paraphrasing, but it boiled down to this:

> Why is `button` an object and not a component?

I _thought_ this would be an easy answer. Pulling up [Harry Roberts' namespace article](https://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/), I re-read his initial description of the object and component namespaces...

> - `o-`: Signify that something is an Object, and that it may be used in any number of unrelated contexts to the one you can currently see it in. Making modifications to these types of class could potentially have knock-on effects in a lot of other unrelated places. Tread carefully.
> - `c-`: Signify that something is a Component. This is a concrete, implementation-specific piece of UI. All of the changes you make to its styles should be detectable in the context you’re currently looking at. Modifying these styles should be safe and have no side effects.

Based on _that_ description, I felt confident in my decision that `o-button` was correctly namespaced. We use buttons in a bunch of different contexts, and you shouldn't feel 100% confident you aren't breaking something by viewing a single instance of it.

But then further down there was some elaboration:

> OOCSS is fantastic in that it teaches us to abstract out the repetitive, shared, and purely structural aspects of a UI into reusable _objects_. This means that things like layout, wrappers and containers, the [Media Object](http://www.stubbornella.org/content/2010/06/25/the-media-object-saves-hundreds-of-lines-of-code/), etc. can all exist as non-cosmetic styles that handle the skeletal aspect of a lot of UI components, without ever actually looking like designed ‘things’.

Uh-oh. That _doesn't_ sound like a button. Buttons have cosmetic styles. They look designed. And later on in that article, Harry shows a sample button class called `c-btn`. So the guy who _invented_ BEMIT thinks buttons are components.

## The problem I _thought_ objects solved

When I read that initial description of objects, I was thinking they solved the Input Group problem.

Y'see, several of our design systems have included [an Input Group pattern](https://cpe-patterns.netlify.com/patterns/elements/forms.html#input-group), where two or more inputs and/or buttons appear _unified_ in a row.

The CSS for these usually looks like this:

```scss
.input-group {
  display: flex;
  /* etc. */
}

.input-group__input {
  flex: 1;
  /* etc. */
}
```

Because computers don't sort things the same way humans do by default, if we import all our components with a glob pattern (`@import 'components/**/*.css';`), `input-group.css` would precede `input.css` in our bundle. This would often cause issues where `.input-group__input` would not override necessary `.input` styles unless we manually enforced the source order.

What _I_ thought objects were meant to do was define those low-level components (`o-input`) as a separate group, followed by specific implementations like (`c-input-group`).

## The problem I _think_ objects are meant to solve instead

Upon re-reading, I now believe objects are really meant to solve a layout problem.

Let's consider issues #474 and #493. Those conversations might yield multiple patterns… an _object_ for everything _but_ the visual elements, with other classes used to specify the content itself:

```html
<div class="o-summary">
  <h3 class="o-summary__headline u-text-grow-1">
    <a class="o-summary__action" href="...">
      ...
    </a>
  </h3>
  <div class="o-summary__object c-thumbnail">
    <img class="c-thumbnail__object" src="..." alt="...">
  </div>
  <div class="o-summary__content">
    <p>
      ...
    </p>
  </div>
</div>
```

But maybe I have this wrong.

## So what?

Currently, `o-button`, `o-icon` and the WIP `o-table` (#557) are all specified as _objects_. But I believe this is incorrect. If others agree, we should move these to the _components_ namespace instead.

What do others think?

---

@cloudfour/design @cloudfour/dev

gerardo-rodriguez:
## Overview

This PR includes:
- moves Button and Icon source from `objects/` to `components/`
- renames Button and Icon class names so they are namespaced `c-` instead of `o-`
- updates any documentation referring to these patterns as "objects"
- changes Table heading from "Table Object" to "Table" (figured I could sneak this in per [PR feedback](https://github.com/cloudfour/cloudfour.com-patterns/pull/565#discussion_r397486212))

### Button

Screen Shot 2020-03-24 at 2 43 57 PM

### Icon

Screen Shot 2020-03-24 at 2 44 05 PM

## Testing

1. Confirm the Button now exists under "Components" (e.g. `?path=/docs/components-button--elements`)
1. Confirm the Icon now exists under "Components" (e.g. `?path=/docs/components-icon--basic`)
1. From the **Design > Icons** page, confirm the "see the icon object" link correctly links to the new Icon URL

---

- Closes #558


src/components/icon/icon.stories.mdx
```diff
@@ -3,11 +3,11 @@ import basicDemo from './demo/basic.twig';
 import inlineDemo from './demo/inline.twig';
 import './icon.scss';
 
-<Meta title="Objects/Icon" />
+<Meta title="Components/Icon" />
 
 # Icon Object
```
calebeby: Should this be changed to `Icon` instead of `Icon Object`? Same for table

gerardo-rodriguez: ✅

src/components/table/table.stories.mdx
```diff
@@ -6,7 +6,7 @@ import './table.scss';
 
 <Meta title="Components/Table" decorators={[withKnobs]} />
 
-# Table Object
```
gerardo-rodriguez: This change is not related to the Button or Icon refactor but was suggested as a change to make via PR feedback from @calebeby: https://github.com/cloudfour/cloudfour.com-patterns/pull/565#discussion_r397486212

gerardo-rodriguez:
@calebeby I addressed your PR feedback. Please let me know if you are done reviewing the PR as @tylersticka has already approved. Thanks! 😉

src/components/table/table.stories.mdx
Paul-Hebert: Oh good catch, thanks @gerardo-rodriguez , @calebeby !

PR #568, opened 2020-03-25
New figure pattern

### LINKED ISSUE #482: Create figure pattern
Create a figure pattern that can accompany long bodies of text like an article. It has an optional caption.

figure
figure pancakes

dromo77:
## Overview

This PR adds and documents a figure pattern in storybook. The caption text color has been updated to meet WCAG guidelines.

## Screenshots

figure

fixes #482


tylersticka:
@dromo77 Hooray! Some general questions/comments:

1. Be sure to address linting errors.
1. Any reason this isn't applied to `<figure>` and `<figcaption>`?

dromo77:
>Be sure to address linting errors

@tylersticka I'm a bit confused about linting, I'm not sure how to find or address the errors.

tylersticka:
> I'm a bit confused about linting, I'm not sure how to find or address the errors.

You have a few options:

- Install linting plugins in your editor of choice, which should flag issues as you edit.
- Run `npm run check-lint` (to show issues) or `npm run lint` (to attempt to fix issues automatically) from Terminal.
- At the bottom of this PR, you can click "Show all checks" and click one of the "more details" for a linting option. This should show you the errors that were encountered.

dromo77:
@tylersticka With the confusion around the linting error I forgot to give an update on this PR. The linting error has been resolved, would you mind giving this another look?

tylersticka:
@dromo77 I thought I'd asked this before but I can't find the conversation now: Any reason why we are making this a component and not the default `<figure>` and `<figcaption>` styles?

src/components/figure/figure.twig
```diff
@@ -0,0 +1,8 @@
+<figure class="c-figure {% if figureClass %} {{figureClass}} {% endif %}">
+  <img src="{{src}}" alt="{{alt}}" class="c-figure__object">
```
tylersticka: Figures can include more than images. We should allow for that in our template.

src/components/figure/demo/basic.twig
```diff
@@ -0,0 +1,7 @@
+{% embed '../figure.twig' %}
+  {% if hasCaption %}
```
tylersticka: Since you're already checking for the existence of the caption block in your template, I'm not sure why we've also added a `hasCaption` boolean here?

src/components/figure/demo/styles-dark.twig
```diff
@@ -0,0 +1,11 @@
+{% embed '../../../themes/demo/fill-canvas.twig' with { theme: 'dark' } %}
+  {% block content %}
+    {% embed '../figure.twig' %}
+      {% if hasCaption %}
+        {% block caption %}
+          This is an image with a caption.
+        {% endblock %}
+      {% endif %}
```
tylersticka: It feels like it's probably unnecessary to recreate this content in two different templates?

src/components/figure/figure.scss
```diff
@@ -0,0 +1,34 @@
...
+ */
+.c-figure {
+  margin: 0;
+}
+
+/**
+ * Contain and center image
+ */
+.c-figure__object {
+  display: block;
```
tylersticka: We are already setting `display: block` by default on media elements in our `base/_defaults` stylesheet.

src/components/figure/figure.scss
```diff
@@ -0,0 +1,34 @@
...
+  margin-left: auto;
+  margin-right: auto;
+  max-width: 100%;
```
tylersticka: We are already setting `max-width: 100%` by default on media elements in our `base/_defaults` stylesheet.

src/components/figure/figure.scss
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/**
+ * Styles the caption
+ */
+.c-figure__caption {
+  color: colors.$gray-dark;
```
tylersticka: Instead of setting a hard `color` value here _and_ for `.t-dark`, maybe we could give the caption an opacity?

dromo77:
>Any reason why we are making this a component and not the default `<figure>` and `<figcaption>` styles?

🤔 I thought this would be a good issue to get my feet wet with pattern work. However looking at the styles, I'm not sure a component is necessary. What do you think @tylersticka ?

tylersticka:
> 🤔 I thought this would be a good issue to get my feet wet with pattern work. However looking at the styles, I'm not sure a component is necessary. What do you think @tylersticka ?

We definitely need figure and figcaption patterns, I just think they can be part of our defaults rather than a full on component.

dromo77:
@tylersticka Instead of a component I added styles for `<figure>` and `<figcaption>` to our defaults.

src/base/_defaults.scss
```diff
@@ -151,3 +151,26 @@ table {
...
+figure {
+  margin: 0; /* 1 */
+
+  & > * {
+    margin-left: auto; /* 2 */
+    margin-right: auto; /* 2 */
+  }
+}
+
+/* Apply default styles to <figcaption> */
```
tylersticka: I'm usually all about comments, but I don't know if this is useful.

src/base/_defaults.scss
```diff
@@ -151,3 +151,26 @@ table {
...
+ * 2. Center content
+ */
+
+figure {
+  margin: 0; /* 1 */
+
+  & > * {
+    margin-left: auto; /* 2 */
+    margin-right: auto; /* 2 */
+  }
```
tylersticka: Could you replace this with `text-align: center` on `figure`, which would also cascade to `figcaption`?

src/base/_defaults.scss
```diff
@@ -151,3 +151,26 @@ table {
...
+}
+
+/* Apply default styles to <figcaption> */
+
+figcaption {
+  font-size: ms.step(-1);
+  font-style: italic;
+  margin-top: ms.step(-1);
```
tylersticka: We may want to wrap this in something in case `figcaption` is not the last child. (Believe it or not, it's actually allowed to have `figcaption` be first or last in a `figure` element!)

dromo77: Do you think all the caption styles should be wrapped or just the `margin-top` that provides space between the `figure` element and `figcaption`?

tylersticka: Just the `margin-top`. You could do something like this if you wanted to:

```scss
figcaption {
  // etc.

  &:last-child {
    margin-top: ms.step(-1);
  }
}
```

dromo77:
@tylersticka I added an example of a figure to the typography stories so you should be able to see it there now. 

I took a little different approach to applying space between elements. I gave all child elements (except the first child) a `margin-top` so there is consistent spacing between elements. I thought this might be helpful in instances when `<figcaption>` appears above the `<figure>` element or if there are multiple `<figure>` elements.

src/design/typography/demo/figure.twig
```diff
@@ -0,0 +1,4 @@
+<figure>
+  <img src="https://picsum.photos/600/300">
```
tylersticka: Nit: Can you add a `width`, `height` and `alt` just so this feels in line with our best practices?

src/base/_defaults.scss
```diff
@@ -151,3 +151,28 @@ table {
...
+figure {
+  margin: 0; /* 1 */
+  text-align: center;
+
+  & > * {
+    margin-left: auto; /* 2 */
+    margin-right: auto; /* 2 */
+  }
+
+  & > :not(:first-child) {
```
tylersticka: A shorter way of writing this would be `& > * + *`. This also has _slightly_ lower specificity. (You can compare [at this website](https://specificity.keegan.st/).)

src/design/typography/demo/figure.twig
```diff
@@ -0,0 +1,4 @@
...
+  <figcaption>I'm a picture with a caption.</figcaption>
```
tylersticka: This is a _total_ nit, but from one design nerd to another: Maybe swap the apostrophe for a curly single quote in content? (`’`)

src/design/typography/demo/figure.twig
tylersticka: Another thought: Could we pull an image on our blog to use as FPO for this?

I believe you can add the image to the `demo` directory, then in the `.mdx` file you can say `import 'path/to/image.svg` (or whatever) and it _should_ build with Webpack.

src/design/typography/demo/figure.twig
Paul-Hebert: Chiming in as a dev nerd, I wonder if we could create some sort of helper (or whatever the Twig equivalent is) that prettifies text like this. For example [markdown-it](https://github.com/markdown-it/markdown-it) has a `typographer` option that enables "some language-neutral replacement + quotes beautification"

Though that maybe makes more sense on the WP side of things?

tylersticka: We get this for free in WP land, I'm just being nitpicky. I don't think it's worth justifying another dependency.

dromo77:
@tylersticka I made a couple changes, mind taking another look?

src/base/_defaults.scss
```diff
@@ -151,3 +151,28 @@ table {
...
+
+/*
+ * 1. Removes default margin
+ * 2. Center content
+ * 3. Space between elements inside <figure>
+ */
+
+figure {
+  margin: 0; /* 1 */
+  text-align: center;
```
tylersticka: I see now why this doesn't replace `margin-left: auto` and `margin-right: auto`. Let's move it back to `figcaption`, it made more sense there. Sorry for steering you wrong.

tylersticka:
@dromo77 Be sure to pull this branch before making changes, I clicked the button that sync with the latest `v-next` so I could test it better.

PR #569, opened 2020-03-25
Set up publishing to npm

calebeby:
## Overview

You can run `npm publish --dry-run` to see everything that _would_ happen during publish, except for the pushing to registry part.

Now there is a `npm run build` task which compiles CSS + JS to the `dist/` folder. The CSS generates minified and unminified versions. The JS generates UMD bundles (can be used in browser via AMD or directly using globals, and can be used in node via commonjs), both minified and unminified, and it generates an unminified ESM bundle.


.storybook/main.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
       name: '@storybook/preset-scss',
       options: {
         sassLoaderOptions: {
-          // Dart Sass performs much better than Node Sass
```
calebeby: Source: https://github.com/sass/dart-sass/blob/master/perf.md

"Dart Sass on Node.js" is usually a bit slower than sassc.

Dart Sass is nice because it doesn't require installing and/or compiling native dependencies.

package.json
```diff
@@ -73,15 +93,17 @@
     "preprocess": "run-p preprocess:*",
     "preprocess:svg-to-twig": "gulp svg-to-twig",
     "preprocess:theo-to-mdx": "gulp theo-to-mdx",
-    "watch-preprocess": "gulp watch",
+    "watch-preprocess": "gulp watch-preprocess",
+    "build": "gulp build",
     "check-lint": "run-s check-lint:*",
-    "check-lint:prettier": "prettier --ignore-path .gitignore --list-different '**/*.{scss,js,mdx}'",
+    "check-lint:prettier": "prettier --ignore-path .gitignore --check '**/*.{scss,js,mdx}'",
```
calebeby: This change is minor and somewhat unrelated, I just noticed that last year Prettier added the `--check` option which is similar to `--list-different` but it prints useful output like "prettier is checking that these files match formatting" ... "These files did not match prettier, did you forget to run the formatter"

gulpfile.js/tasks/build.js
```diff
@@ -0,0 +1,58 @@
...
+sass.compiler = require('sass');
+
+const outDir = 'dist';
+
+const buildSass = () => {
+  return src('./src/index.scss')
+    .pipe(
+      sass({
+        // Import Theo design tokens as SCSS variables
+        importer: require('../../.storybook/theo-importer'),
```
tylersticka: Since this is now a dependency of both the build task and Storybook, should we move it out of the `.storybook` directory?

src/index.scss
```diff
@@ -0,0 +1 @@
+@use "./base";
```
calebeby: This file was needed because the gulp sass implementation requires the entry file to not start with an underscore

tylersticka: Does this mean we aren't including any component CSS yet?

.gitignore
```diff
@@ -11,3 +11,6 @@ npm-debug.log
 
 # Local configurations
 .vscode
+
+# Ignore generated MDX docs
+/src/design-tokens/generated/*.mdx
```
calebeby: Previously this ignore was in a different .gitignore in that folder. But, that did not work for Prettier + ESLint because they are set to only ignore files that are ignored in _this_ file

src/index.scss
calebeby: That was a mistake, I fixed it in https://github.com/cloudfour/cloudfour.com-patterns/pull/569/commits/78210ab34358d7f7e002086460254cb99839ed51

gulpfile.js/tasks/build.js
calebeby: Fixed in https://github.com/cloudfour/cloudfour.com-patterns/commit/78210ab34358d7f7e002086460254cb99839ed51

calebeby:
Ready for re-review ✅

src/index.scss
```diff
@@ -0,0 +1,4 @@
...
+// The "as _" is necessary because otherwise
+// Sass tries to use the * from the glob as the namespace and fails
+@use './components/**/*.scss' as _;
```
tylersticka: Would this be a better fit for [`@forward`](https://sass-lang.com/documentation/at-rules/forward)?

calebeby: My understanding is that `@forward` is like an `export * from "..."` in JS. Initially I didn't think that is what we want because for this file, I intended that the only target usage is for it to be generated into the css file in `dist`. But then it occurred to me that people might also want to import this file directly as a sass file. Is that a valid use case?

tylersticka: I'm honestly not sure, as I'm just getting back into the Sass world. But it seems like our options are:

- `@use`, which requires a namespace to be specified (as mentioned in your comment) even though we'll never be using it
- [`@import`](https://sass-lang.com/documentation/at-rules/import), which does not require a namespace but Sass discourages its continued use
- `@forward`, which does not require a namespace and will make whatever was included available when imported and is intended to be used when splitting a single package up among multiple files

Of the three, `@forward` felt the most appropriate to me. But I might be wrong!

calebeby: I think that in this context the only difference between @use and @forward is that if another module @use'd the `index.scss` file, then that other module would have access to mixins from the components and other files. But I only created this `index.scss` file to generate the `dist` css file, it isn't intended to be `@use`'d.

tylersticka: So if I am understanding this correctly (and I may not be)...

- Both techniques result in the same bundled `*.css` file.
- If we use `@use` (heh), you won't be able to access mixins, variables, etc. via Sass except through a leading underscore namespace.
- If we use `@forward`, you _would_ be able to access mixins, variables, etc. via Sass.

If that's accurate, I still think `@forward` sounds like what we want? Or am I missing somethin?

calebeby: I was planning on adding this file to the npm exclude list so it doesn't even get published. This file is not intended to be used, other than to generate the css file in dist. So there is no need to export mixins, variables, etc. from this file.

calebeby: This is for the same reason that I wouldn't export anything from an entry-point JS file. It's not going to get imported anywhere, so there is no need to export anything

tylersticka: Okay, so is this more accurate then?

- Both techniques result in the same bundled `*.css` file.
- If we use `@use`, we have to specify a `_` namespace.
- If we use `@forward`, it just works as expected.

calebeby: "it just works as expected"

In this scenario, I don't want to export anything, so @forward does not do what I expect.

tylersticka: > In this scenario, I don't want to export anything, so @forward does not do what I expect.

Maybe we need to hop on a call at some point, I'm very confused.

I find this line of SCSS:

```scss
@use './components/**/*.scss' as _;
```

More confusing than this one:

```scss
@forward './components/**/*.scss';
```

It sounds like you're worried about `index.scss` exporting things. But you also say that it is "not going to get imported anywhere." So I don't understand what the issue is.

Is the bundle CSS different? Are there regressions? Is there an API key in our Sass variables we don't want to expose? Do underscores give us street cred? What am I missing?

calebeby: I realized that the underscore may be confusing so I changed it to `components`: https://github.com/cloudfour/cloudfour.com-patterns/pull/569/commits/36b6c394877356b61d2617224edc3ec5548ff392#diff-5ab26e17c78a72ee4bc1774adbcf305eR4

For me, `@forward` raises more questions when I see it because it makes me think "where is this forwarding to, what is supposed to be importing this". `@use` is more clear to me because it just says what it does, it is only loading in another file.

It's like in JS using `export * from "..."` as an import rather than using `import`. I wouldn't use an export statement just for the purpose of importing. I would use an import statement instead.

tylersticka: I think the majority of us interacting with the Sass won't have that same point of confusion. But just to double check, I decided to see what other large UI frameworks are doing.

- [Bootstrap is sticking with `@import` for now](https://github.com/twbs/bootstrap/issues/30025#issuecomment-574825600), as are Salesforce, Mozilla Protocol and Shopify Polaris.
- Material uses `@use` but no globs (just manually lists packages 🤔): https://github.com/material-components/material-components-web/blob/master/packages/material-components-web/material-components-web.scss

In light of that, I think I'm okay with this. I still disagree, but I no longer feel passionate about it.

tylersticka:
This is in need of developer review.

@spaceninja seems like a good fit since he's written about npm publishing, but anyone from @cloudfour/dev is welcome to take on.

Thank you!

package.json
```diff
@@ -22,13 +22,28 @@
...
+    "!src/index.scss"
+  ],
   "dependencies": {
     "focus-visible": "5.0.2",
     "prismjs": "1.19.0"
   },
   "devDependencies": {
     "@babel/core": "7.9.0",
     "@cloudfour/eslint-plugin": "8.0.0",
+    "@rollup/plugin-multi-entry": "3.0.0",
```
spaceninja: Are we using rollup and gulp? That seems odd.

package.json
```diff
@@ -73,15 +95,17 @@
...
-    "check-lint:prettier": "prettier --ignore-path .gitignore --list-different '**/*.{scss,js,mdx}'",
+    "check-lint:prettier": "prettier --ignore-path .gitignore --check '**/*.{scss,js,mdx}'",
     "check-lint:css": "stylelint --ignore-path .gitignore '**/*.scss'",
     "check-lint:js": "eslint --ignore-path .gitignore '**/*.{js,mdx}'",
     "lint": "run-s lint:*",
     "lint:prettier": "prettier --ignore-path .gitignore --write '**/*.{scss,js,mdx}'",
     "lint:css": "stylelint --ignore-path .gitignore --fix '**/*.scss'",
-    "lint:js": "eslint --ignore-path .gitignore --fix '**/*.{js,mdx}'"
+    "lint:js": "eslint --ignore-path .gitignore --fix '**/*.{js,mdx}'",
+    "prepublishOnly": "npm ci && npm run check-lint && npm run preprocess && npm run build"
```
spaceninja: Could you add a CONTRIBUTING.md file with instructions on how to publish to npm? (even if it's just "run `npm version` and `npm publish`)

package.json
calebeby: Yes, we are using both, they serve different purposes.

We are using Gulp as a task runner, to pipe code from one tool to another.
We are using Rollup as a bundler, to resolve dependencies of js code, tree-shake, and combine.

Josh68: @spaceninja, I'm curious, have you used Rollup without a task runner? In the past I've done this with Webpack, relying on NPM scripts for orchestration. But @calebeby set up other projects with this combination, and it seems like a good approach.

gulpfile.js/tasks/build.js
```diff
@@ -0,0 +1,62 @@
...
+
+sass.compiler = require('sass');
+
+const outDir = 'dist';
+
+const buildSass = () => {
+  return src('./src/index.scss')
+    .pipe(
+      sass({
+        // Import Theo design tokens as SCSS variables
```
Josh68: Is this an accidental duplicate comment, or intentional?

sass-importers/glob.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {string} url Raw path in source code
+ * @param {string} currentPath Path to the current file
+ * @param {(result: ResolveResult) => void} done Callback function for async resolvers
+ * @returns {ResolveResult | void}
+ */
+const globImporter = (url, currentPath, done) => {
+  // This is a really lazy check but it is easy and good enough
+  const isGlob = url.includes('*');
+  if (!isGlob) return null;
+  glob(url, { cwd: path.dirname(currentPath) }).then((matchingPaths) => {
```
Josh68: I could be totally off, but does there need to be a `return` before the `glob` function?

Josh68: I think I am totally off, missing the call to `done`

package.json
spaceninja: I have very little experience, I was asking because in my head I think of Rollup and Gulp as basically doing the same thing. But this approach makes sense (and I trust Caleb)

spaceninja:
@calebeby I ran the dry run, and I got five files in `/dist`. Can you confirm this is what's expected?

- cloudfourPatterns.js - contains an empty factory function?
- cloudfourPatterns.min.js - same, but minified
- cloudfourPatterns.mjs - empty file
- standalone.css - looks to be a compiled version of all our Sass in one file
- standalone.min.css - same, but minified

I'm guessing this is all correct, and the odd JS files are just because there isn't any JS in our pattern library yet?

Did we want to also add the sass files to the `/dist` folder so consumers have the option to add the sass directly (ala Bootstrap?)

calebeby:
> I'm guessing this is all correct, and the odd JS files are just because there isn't any JS in our pattern library yet?

Yeah. To test this out I created a `button.js` file in the button component folder and exported something from it, and made sure it showed up in the built js files

> Did we want to also add the sass files to the /dist folder so consumers have the option to add the sass directly (ala Bootstrap?)

@spaceninja the sass files in the `src` directory are included in the npm package so consumers should still be able to access it without us copying them to `dist`

spaceninja:
Awesome — in that case, I'm happy to approve once you add that CONTRIBUTING.md file

package.json
```diff
@@ -22,13 +22,28 @@
     "type": "git",
     "url": "https://github.com/cloudfour/cloudfour.com-patterns.git"
   },
+  "main": "./dist/cloudfourPatterns.js",
+  "unpkg": "./dist/cloudfourPatterns.min.js",
```
gerardo-rodriguez: > `"unpkg": "./dist/cloudfourPatterns.min.js",`

TIL 🎓 

https://kentcdodds.com/blog/unpkg-an-open-source-cdn-for-npm

package.json
Josh68: When I started doing this kind of stuff (not nearly as complex as what @calebeby is doing), I was using Grunt and then brought Webpack into that system. After I while, I wanted to ditch Grunt and turned to NPM scripts (and some script utilities) to orchestrate things, but I think Gulp provides a cleaner, more composable way of doing this.

sass-importers/glob.js
```diff
@@ -0,0 +1,27 @@
...
+    // Generates a fake file that has the expanded @use statements
+    const contents = matchingPaths
```
Josh68: The comment references `@use`, while the code adds `@forward`. Is this correct?

package.json
gerardo-rodriguez: @calebeby Should we be using `&&` here? 🤔 

My (recent) understanding is this won't work on Windows? I might also be wrong. ¯\\\_(ツ)_/¯

I gathered this from https://github.com/peerigon/scriptlint/wiki/no-unix-double-ampersand

tylersticka: If this is a concern, [`npm-run-all`](https://www.npmjs.com/package/npm-run-all) (which is already installed) is cross-platform. I believe `run-s ci check-lint preprocess build` would work, but I haven't checked.

spaceninja: Good catch, @gerardo-rodriguez!

Josh68: I also wonder whether this applies: https://stackoverflow.com/questions/48104380/single-quotes-not-working-in-package-json (escape double-quotes vs use single-quotes). Should I try grabbing this and running on my Windows 10 laptop?

calebeby: @Josh68 Did the `&&` in the script work? I googled it and found conflicting answers

Josh68: My only problem (which I'm not going to try to solve right now) is rebuilding node-gyp in Windows 10. Everything else seems fine, and also ran perfectly on WSL2 (Ubuntu).

Josh68: @calebeby, with the `node-gyp` hangup, I may not be able to give you a definitive answer. I thought maybe I'd resolved that, but not yet. It's a Python thing.

Josh68: I am able to run `npm run build-storybook` without issue. It contains `&&`

tylersticka: FWIW, I think a lot of people (myself included) who develop on Windows have switched to WSL, which handles ampersands just fine.

I think given the audience for this (us), optimizing for unix in the short-term is acceptable.

gerardo-rodriguez: > I think given the audience for this (us), optimizing for unix in the short-term is acceptable.

Fine by me! I'll resolve the convo. 👍 🙂

package.json
```diff
@@ -73,15 +95,17 @@
     "preprocess": "run-p preprocess:*",
     "preprocess:svg-to-twig": "gulp svg-to-twig",
     "preprocess:theo-to-mdx": "gulp theo-to-mdx",
-    "watch-preprocess": "gulp watch",
+    "watch-preprocess": "gulp watch-preprocess",
+    "build": "gulp build",
     "check-lint": "run-s check-lint:*",
-    "check-lint:prettier": "prettier --ignore-path .gitignore --list-different '**/*.{scss,js,mdx}'",
+    "check-lint:prettier": "prettier --ignore-path .gitignore --check '**/*.{scss,js,mdx}'",
```
gerardo-rodriguez: @calebeby Thoughts on the following setup to reduce the duplicate Prettier setup? 🙂 

```js
{
  "prettier": "prettier --ignore-path .gitignore '**/*.{scss,js,mdx}'",
  "check-lint:prettier": "npm run prettier -- --check",
  "lint:prettier": "npm run prettier -- --write"
}
```

gerardo-rodriguez:
@calebeby I see a bunch of comments at the top of the `dist/standalone.css` file that seem a bit out of place. Is this expected? 🤔 

Screen Shot 2020-03-30 at 9 05 23 AM

I'm not sure if this is a concern for this PR or not but figured I'd ask. 🙂

gerardo-rodriguez:
@calebeby Is this a concern at all? 🤔 

Screen Shot 2020-03-30 at 9 12 16 AM


```
Filename is not in kebab case. Rename it to `cloudfour-patterns.js`.
eslint@cloudfour/unicorn/filename-case
```

tylersticka:
> I see a bunch of comments at the top of the `dist/standalone.css` file that seem a bit out of place. Is this expected? 🤔

@gerardo-rodriguez Here's what I believe is happening:

1. We've added comments to Theo.
1. When Theo is imported into Sass, it converts the comments to comment strings that accompany the variables.
1. The variables are compiled but the comments are not.

As long as these go away in the minified file, I don't think it's an issue.

Potentially could be resolved by changing the comment formatting to `\\` instead of `\* *\`, but I'm not sure it's worth it (and certainly out of scope for this PR).

gerardo-rodriguez:
> As long as these go away in the minified file, I don't think it's an issue.

Works for me! 🙂

package.json
calebeby: Good idea, added ✅

gulpfile.js/tasks/build.js
calebeby: That was an accident, I fixed it ✅

package.json
calebeby: I changed prettier to pass the directory rather than globs (so it automatically runs on all files it supports), so I think this is no longer necessary.

sass-importers/glob.js
calebeby: Fixed ✅

calebeby:
I created separate .eslintignore and .prettierignore files (rather than pointing the CLI's to .gitignore) so that the CLI's and editors activate on the same files (because the editor plugins do not look at .gitignore)

calebeby:
I also resolved a bunch of merge conflicts, so this is ready for re-review

src/design-tokens/_easing.yml
```diff
@@ -25,11 +25,11 @@ aliases:
   ease_in_out_back: cubic-bezier(0.68, -0.55, 0.265, 1.55)
 props:
   ease_in_out:
-    value: "{!ease_in_out_quad}"
+    value: '{!ease_in_out_quad}'
```
calebeby, line 28: These changes in .yml files are because now prettier is being run on all files it supports

spaceninja:
@calebeby what does the github action in this PR do?

calebeby:
@spaceninja it runs the build, so that if there is a build error we find out about it in CI

spaceninja:
Awesome.

Also, what's the difference between the `build` script and the `build-storybook` script?

calebeby:
The `build` script creates files to be used by npm (compiled css+js files)

The `build-storybook` builds the storybook site which gets published to netlify

spaceninja:
Great, thanks for clarifying — perhaps that distinction should live in a comment or the README?

spaceninja:
I've kinda lost track of where this PR sits in terms of approval, but I've just gone over it, and it does what it says on the label — a dry run confirms the package is created as expected. I'm happy to approve if there are no objections.

I'm really excited to see this trigger get pulled and the pattern library become an npm package we can consume in the wp app!

calebeby:
@Josh68 @gerardo-rodriguez Did you want to take another look?

gerardo-rodriguez:
> @Josh68 @gerardo-rodriguez Did you want to take another look?

Thanks, @calebeby. I won't be able to review today. I'm confident in the team review that has been provided. Feel free to merge without my re-review/approval. My concern was addressed. 😉

PR #570, opened 2020-03-27
Define breakpoints and max-widths

### LINKED ISSUE #547: Draft standard breakpoints and max widths
Our pattern library has traditionally included standard breakpoints, which are helpful for the reasons [I describe in this article](https://cloudfour.com/thinks/breakpoints-dont-box-me-in/). As of today, these have been:

```css
@custom-media --sm-viewport (width >= 30em);
@custom-media --md-viewport (width >= 48em);
@custom-media --lg-viewport (width >= 64em);
@custom-media --xl-viewport (width >= 80em);
@custom-media --xxl-viewport (width >= 100em);
@custom-media --xxxl-viewport (width >= 120em);
```

It has also included patterns for establishing maximum widths that prevent content from becoming too wide, either to maintain the integrity of the design or to maintain readability:

```css
.u-containSpread {
  max-width: 80em; /* ~ 1600 @ 20px base */
}

.u-containProse {
  max-width: 40rem; /* ~ 640 @ 16px base */
}
```

Historically, these have been thought of as separate decisions. But I suspect they should be thought of together: Not only should the prose width cleanly divide into the maximum spread/layout width, but both of those should have some relationship to our standard breakpoints.

## Tasks

- [ ] Define a set of standard breakpoints
- [ ] Define a set of maximum content widths for prose and layout
- [ ] Translate these into design token files
- [ ] Update existing CSS breakpoints (currently used for typography) to leverage token values

We can tweak these later, but we need a starting point!

dromo77:
## Overview

This PR  creates tokens for standard breakpoints and maximum prose and spread widths. These can always be adjusted later on as needed.

## Screenshots

layout

---

- Fixes #547


src/design-tokens/layout.yml
```diff
@@ -0,0 +1,45 @@
...
+  width_64: '64em'
+  width_80: '80em'
+  width_100: '100em'
+  width_120: '120em'
+props:
+  breakpoint_xs:
+    value: '{!width_30}'
+    type: media-query
+    category: media-query
+  breakpoint_sm:
```
tylersticka: On a recent client project, we reduced some of these abbreviations down to a single letter. So instead of `xs`, `sm`, `md`, `lg`, `xl`, `xxl`, it was `xs`, `s`, `m`, `l`, `xl`, `xxl`.

I didn't really miss the extra letters. What do you think?

dromo77:
Thanks for the feedback @tylersticka , would you mind taking another look?

src/design-tokens/breakpoint.yml
```diff
@@ -0,0 +1,21 @@
...
+    value: '{!width_40}'
+  - name: breakpoint_m
+    value: '{!width_48}'
+  - name: breakpoint_l
+    value: '{!width_64}'
+  - name: breakpoint_xl
+    value: '{!width_80}'
+  - name: breakpoint_xxl
+    value: '{!width_100}'
+  - name: breakpoint_xxxl
```
tylersticka: Are the `breakpoint_` prefixees in these names necessary now that they live in their own file?

src/design-tokens/_widths.yml
```diff
@@ -0,0 +1,8 @@
+aliases:
+  width_30: '30em'
+  width_40: '40em'
+  width_48: '48em'
+  width_64: '64em'
+  width_80: '80em'
+  width_100: '100em'
+  width_120: '120em'
```
tylersticka: Just a thought: Would these be more helpful if they were disconnected from the actual width? So like `width_xs`, `width_s`, etc.

dromo77:
@tylersticka Made a couple more changes, can you take another look?

src/design-tokens/generated/layout.yml
```diff
@@ -0,0 +1,11 @@
...
+  - './_widths.yml'
+
+props:
+  - name: max_width_prose
+    value: '{!width_40}'
+  - name: max_width_spread
+    value: '{!width_80}'
+global:
+  type: relative
+  category: max-width
```
tylersticka: I don't think this file is supposed to be here?

src/design-tokens/generated/breakpoint.yml
```diff
@@ -0,0 +1,21 @@
...
+    value: '{!width_64}'
+  - name: xl
+    value: '{!width_80}'
+  - name: xxl
+    value: '{!width_100}'
+  - name: xxxl
+    value: '{!width_120}'
+global:
+  type: relative
+  category: media-query
```
tylersticka: I don't think this file is supposed to be here?

src/design-tokens/_widths.yml
```diff
@@ -0,0 +1,8 @@
+aliases:
+  width_xs: '30em'
+  width_s: '40em'
+  width_m: '48em'
+  width_l: '64em'
+  width_xl: '80em'
+  width_xxl: '100em'
+  width_xxxl: '120em'
```
tylersticka: Do you think there's an opportunity to make these jumps more consistent?

Right now the gaps between steps are 10, 8, 16, 16, 20, 20.

I went through our old pattern CSS and markup to see which breakpoints we were actually using. Here's what I found...

Step | Value | CSS | Markup
--- | --- | --- | ---
sm | 30em | 22 | 102
md | 48em | 28 | 312
lg | 64em | 22 | 92
xl | 80em | 5 | 0
xxl | 100em | 2 | 0
xxxl | 120em | 2 | 0

What that tells me is that the < 100em values should definitely influence our scale the most. Right now, the scale feels least consistent in the smaller values. It also seems like

Maybe something like this could work?

Step | Ems | Pixels | Notes
--- | --- | --- | ---
xs | 20 | 320 | Old iPhone size, start of responsive font scaling
s | 30 | 480 | Old iPhone landscape
m | 40 | 640 | Maximum prose width, small tablet widths
l | 60 | 960 | Classic "desktop" design width
xl | 80 | 1280 | Approaching classic laptop resolution
xxl | 100 | 1600 | Common desktop monitor resolution
xxxl | 120 | 1920 | HD width

Then the gaps are 10, 10, 20, 20, 20, 20. Crazy-ness?

dromo77: That makes sense to me 👍

dromo77:
@tylersticka I updated the breakpoints to scale more consistently.

src/base/_typography.scss
```diff
@@ -15,8 +16,8 @@
 
 html {
   @include fluid.font-size(
-    20em,
-    75em,
+    breakpoint.$s,
```
tylersticka: Can this be `breakpoint.$xs`? I'd love the type to start sizing upward from `20em` as it did before.

PR #573, opened 2020-03-30
Replace centered add-on with paddings

### LINKED ISSUE #546: Add padding to Storybook previews
As @Paul-Hebert pointed out while reviewing #545, some wider patterns are reaching the edge of the Storybook canvas:

Some ways we could address this:

- Via an add-on like [storybook-addon-paddings](https://github.com/rbardini/storybook-addon-paddings)
- Via whitespace utility classes, which we've yet to add or discuss

### LINKED ISSUE #566: Icon is cut off on non-retina display in FF and Safari
I'm not sure if this is a high-priority issue, but I noticed the heart icon is cut off on the right side when viewing through my non-retina display at wider viewports.

It seems to only happen in the following browsers for me:
- FF 74.0
- Safari 13.0.5

## FireFox 74.0

Screen Shot 2020-03-24 at 2 57 00 PM

## Safari  13.0.5

Screen Shot 2020-03-24 at 3 10 36 PM

tylersticka:
## Overview

This PR replaces [the centered add-on](https://github.com/storybookjs/storybook/tree/master/addons/centered) with [the paddings add-on](https://github.com/rbardini/storybook-addon-paddings) now that the issue I encountered there [has been fixed](https://github.com/rbardini/storybook-addon-paddings/issues/6). This has a few benefits:

- Canvas contents will now have padding by default (instead of running up against the edges).
- Padding can now be adjusted from the canvas toolbar.
- Full-width elements are no longer unexpectedly smooshed.
- Elements that exhibited visual bugs and rounding errors from the centering technique alone no longer do, which should make it easier to diagnose those sorts of things. (See #566)
- While the theme demo template hack is still a requirement until Themes work in docs mode, it has been greatly simplified.

## Screenshots

paddings

## Testing

View deploy preview and confirm fixes described in overview.

---

- Fixes #546 
- Fixes #566 

/CC @Paul-Hebert @gerardo-rodriguez @derekshirk


Josh68:
I have bandwidth to look at this now.

.storybook/preview.js
```diff
@@ -1,14 +1,26 @@
...
-addDecorator(centered);
+// Accessibility testing via aXe
 addDecorator(withA11y);
 
 // Theme selection from stories
 const themes = [{ name: 'Dark', class: 't-dark', color: colors.primaryBrand }];
 addParameters({ themes });
+
+// Padding values from modular scale
+const paddings = [];
```
Josh68: Nit, optional approach:

```js
const paddings = new Array(10).fill(0).map((k,i) => {
  const idx = i - 3;
  return { 
    name: `Step ${idx}`,
    value: `${Math.pow(ms.ratio, idx).toFixed(2)}em`,
    default: idx === 0
  };
});
```

Josh68: Not absolutely sure, but think the `+` for casting isn't needed before `Math.pow`

.storybook/preview.js
```diff
@@ -1,14 +1,26 @@
 import { addDecorator, addParameters } from '@storybook/html';
 import { withA11y } from '@storybook/addon-a11y';
-import centered from '@storybook/addon-centered/html';
+import { withPaddings } from 'storybook-addon-paddings';
 import * as colors from '../src/design-tokens/colors.yml';
+import * as ms from '../src/design-tokens/modular-scale.yml';
```
Josh68: Just to see how this works (with whatever webpack loader I assume is responsible), I tried this

```js
import { ratio } from '../src/design-tokens/modular-scale.yml';
//...
value: `${Math.pow(ratio, idx).toFixed(2)}em`,
```
Seems to work. If the explicitness of `ms.ratio` is preferred, I'm not picky about which way it's imported.

src/themes/themes.stories.mdx
```diff
@@ -15,7 +15,11 @@ While selector depth is best kept to a minimum, it's permissable to style object
 `t-dark` is intended primarily for page headers.
 
 <Preview>
-  <Story name="Dark" height="200px" parameters={{ themes: { disable: true } }}>
+  <Story
+    name="Dark"
+    height="200px"
+    parameters={{ themes: { disable: true }, paddings: { disabled: true } }}
```
Josh68: I'm probably missing it, but can't tell what the `paddings` property is changing here. Also, FYI (nit) see spelling `permissable` -> `permissible` in this file.

tylersticka: It is disabling the paddings add-on for this story because it doesn't work with the theme preview.

I'll fix the typo 👍

.storybook/preview.js
tylersticka: @Josh68 I find that alternative approach a lot harder to read. It's also slightly more code, so I kept the version I have.

I added a comment explaining what `+` is doing here (it prevents numbers that don't need decimals like `1.00` from having them).

.storybook/preview.js
tylersticka: You're correct, the Theo loader is currently set up to write `export` statements for all of the design tokens in a file.

I've simplified in this case because I'm not sure what other values this value may export in the short term.

.storybook/preview.js
Josh68: You're right, it is harder to read. I've gotten used to preferring array helpers over loops, but in this case, the loop makes perfect sense.

src/themes/themes.stories.mdx
Josh68: Just to clarify, I found this setting made a visible difference when I toggled it in the demo that's imported (`fill-canvas`), but couldn't see what difference it was making here, when I toggled it. I may have been missing it.

tylersticka:
> I had already approved, but my approval seemed to have disappeared.

@Josh68 Approvals are dismissed when changes are made.

Issue #490, opened 2020-02-21
Input pattern(s)


dromo77:
Create patterns for input fields.

<img width="638" alt="forms" src="https://user-images.githubusercontent.com/42841342/75058979-ddb6f400-5490-11ea-9ad0-8948a8260f37.png">

tylersticka:
This should go hand in hand with #475 to make sure the elements feel comparable in size.

We should take extra care to avoid the "floating label" pattern in its current form because there are accessibility and obviousness concerns with that pattern.

We should also try to embrace better contrast. The unchecked box/radio state is particularly tough to spot with the current design.

tylersticka:
I think I'm becoming a speed bump for this work, so I'm going to un-assign myself and add some tasks for this. Don't hesitate to reach out if you have questions or want to run ideas past me.

## Types
- [ ] Text
- [ ] Email
- [ ] Search
- [ ] Text Area
- [ ] Select
- [ ] Checkbox
- [ ] Radio

## States
- [ ] Default
- [ ] Focus
- [ ] Disabled
- [ ] Read-only
- [ ] Hover
- [ ] Active

## Goals, features and ideas
- [ ] Improve contrast
- [ ] Avoid "floating label" pattern (ux issues)
- [ ] Avoid "input group" pattern (never used)
- [ ] Consistent layout (dimensions, border widths, etc.) with buttons (consider using design tokens to share values)
- [ ] Distinct appearance from buttons
- [ ] Explore using `text-indent` so overflow passes over padding
- [ ] Consider design of labels (maybe prototype a comment form?)
- [ ] Support dark theme

tylersticka:
Helpful reading: https://css-tricks.com/custom-styling-form-inputs-with-modern-css-features/

tylersticka:
Stole a half-hour to start making progress on this today. Still work to do but having fun designing something higher contrast than before.

<img width="431" alt="Screen Shot 2020-03-27 at 4 41 57 PM" src="https://user-images.githubusercontent.com/69633/77808759-fb581a00-7049-11ea-8689-1e6e3c6df7d7.png">

PR #579, opened 2020-04-01
Modular scale value transforms for design tokens

### LINKED ISSUE #549: Support modular scale steps as Theo transforms
This project uses a modular scale, but Theo has no built in way of defining modular scale steps. But it _does_ have a method of [defining custom value transforms](https://github.com/salesforce-ux/theo#custom-transforms--value-transforms).

In theory if we used Theo to read the value of the `ms.ratio` token first, we could then define `ms`, `ms/rem` and `ms/em` transforms that we apply based on a certain type or category value.

So something like this (just as a for instance):

```yaml
# design-tokens/sizes.yml
props:
  - name: pad_control_x
    value: 0
    type: ms-em
    category: sizing
  - name: pad_control_y
    value: -3
    type: ms-em
    category: sizing
```

Then referenced like this:

```scss
@use "../../design-tokens/sizes.yml";

.o-button {
  padding: sizes.$pad-y sizes.$pad-x;
}

.o-input {
  padding: sizes.$pad-y sizes.$pad-x;
}
```

Would compile to:

```css
.o-button {
  padding: .512em 1em;
}

.o-input {
  padding: .512em 1em;
}
```

Alternatively, there's no reason we couldn't just define those steps and then pipe the values into the `ms.step` mixin we already have. But having the transforms occur in our tokens might simplify our CSS and documentation a bit.

tylersticka:
## Overview

Adds modular scale value transforms for design tokens. This means you can specify certain types and they will automatically convert to a usable value based on our modular scale.

For example, these design tokens:

```yaml
props:
  - name: example_unitless
    value: 0
    type: modular
  - name: example_em
    value: 2
    type: modular/em
  - name: example_rem
    value: -1
    type: modular/rem
  - name: example_px
    value: 3
    type: modular/px
global:
  category: example 
```

Would result in the following values when used:

Sass Variable | Value
--- | ---
$example-unitless | 1
$example-em | 1.5625em
$example-rem | 0.8rem
$example-px | 31px

I didn't think this would be necessary until I started #490, which requires input sizes to coordinate with buttons. Since `height` is _still_ the only reliable way to do this, it became challenging to coordinate sizes that would align nicely with adjacent typography. Being able to specify shared dimensions via tokens without doing manual calculations is awesome.

In order to pull this off for both CSS and JavaScript, I centralized all of our Theo configuration into a single directory that exports the configured Theo object from its `index.js`. This has an additional benefit of moving the Sass importer out of `.storybook`, which was an issue I had with #569.

## Screenshots

theo-ms

## Testing

1. Check out this branch locally.
1. Confirm that `npm run storybook` is still successful.
1. Open the `sizes.yml` token file.
1. Change the `control_pad_vertical` token value to a different step.
1. Confirm that the change takes effect for both the button component and the design token sizes documentation page.
1. Change the `type` of the token and confirm it still works. Valid types for padding are `modular/em`, `modular/rem` and `modular/px`.

---

- Resolves #549


tylersticka:
I pushed a couple of small updates this morning:

- I renamed the transforms from `ms/*` to `modular/*`. This more closely resembles Theo's built-in formats (`relative/*`, `floating/*`, etc.), does not obscure the meaning with an acronym, and avoids confusion with the abbreviation for milliseconds.
- I added the `.theo` directory to the README's "Project Structure" section.

PR #581, opened 2020-04-01
Text inputs

tylersticka:
## Overview

Introduces a `c-input` component that covers text-based inputs. It works in default and dark themes and has `readonly`, `disabled`, `focus` and `hover` styles.

Also normalizes heights and dimensions between inputs and buttons so they should look nice in proximity regardless of type or browser.

It also adds a downward caret icon intended for `select` support, but I started going down a rabbit hole of SVG inlining and decided to pause and submit a PR.

## Screenshots

Screen Shot 2020-04-01 at 3 27 32 PM

## Testing

1. View Component/Input story in deploy preview.
1. Try various knobs and confirm that they work as expected.
1. Try various knobs again after switching to the Dark theme from the Storybook toolbar.

---

- See #490


Josh68:
@tylersticka, unless I'm not understanding correctly, it looks like a deployment failed. Did your PR get picked up?

tylersticka:
@Josh68 I don’t see any deploy failure or other issue. Not sure what you mean by picked up.

Josh68:
> @Josh68 I don’t see any deploy failure or other issue. Not sure what you mean by picked up.

That was unclear. I meant deployed. Should I see the `input` component here in the preview?

image

src/components/button/button.scss
```diff
@@ -15,6 +15,8 @@
  * 2. Although we will be overtly setting some typographic properties, the
  *    shorthand is helpful for completely overriding all system font styles as
  *    a starting point.
+ * 3. We set a height to coordinate consistently with `c-input`. Unfortunately
```
Josh68, line 18: 👍

src/components/input/input.scss
```diff
@@ -0,0 +1,99 @@
...
+  padding: sizes.$control-pad-vertical sizes.$control-text-inset;
+  transition-duration: motion.$speed-quick;
+  transition-property: background-color, border-color;
+  transition-timing-function: motion.$ease-out;
+  width: 100%;
+
+  /**
+   * For certain single-line input types, using `text-indent` instead of
+   * `padding-left` gives a more natural and intuitive appearance when text
+   * content overflows.
```
Josh68, line 41: Nit: Not sure I get this -- testing with or without the override on an email input I can't really see the difference, but I well may be missing it.

src/components/input/input.scss
```diff
@@ -0,0 +1,99 @@
...
+@use "../../mixins/ms";
+@use "../../mixins/unit";
+
+/**
+ * 1. Browsers apply certain default styles unless appearance is set to `none`.
+ *    Unfortunately Chrome, Safari and Mozilla all still rely on the
+ *    vendor-prefixed version of this property at the time of this writing.
+ * 2. Input heights vary between different types in certain browsers unless a
+ *    height is explicitly set.
+ * 3. Non-normal line heights break placeholder text alignment in Safari.
```
Paul-Hebert, line 14: Oof, that's annoying!

src/components/input/input.scss
tylersticka: With `padding-left`, overflowing text gets awkwardly cut off on the left:

padding

With `text-indent`, it does not:

text-indent

src/components/input/input.scss
```diff
@@ -0,0 +1,99 @@
...
+   */
+
+  &[type='email'],
+  &[type='text'] {
+    padding-left: 0;
+    text-indent: sizes.$control-text-inset;
+  }
+
+  &:hover:not(:disabled):not([readonly]) {
+    background-color: #fff;
```
Paul-Hebert, line 54: Nit: Should we be using a design token for our white color?

src/components/input/input.scss
```diff
@@ -0,0 +1,99 @@
...
+
+    .t-dark & {
+      background-color: colors.$primary-brand-lighter;
+      border-color: colors.$primary-brand-dark;
+    }
+  }
+}
+
+/**
+ * This is ugly, but unfortunately we can't target an attribute like `type` for
```
Josh68, line 92: Nit. I assume it would be considered worse (and uglier) to require an additional class with a modifier for `textarea`s, so I might just start the comment with something like "Using element selector because we can't ..." You could also reference the comments about why non-`textarea` inputs need a height set, but not necessary.

src/components/input/input.scss
tylersticka: Probably, yes!

dromo77:
From a design standpoint I think this looks good!

src/components/input/input.scss
Paul-Hebert: Ahh thanks. I didn't fully understand what this was doing either. Adding that sentence to the comment may be helpful

src/components/input/input.scss
tylersticka: ✅ Updated to use a design token

src/components/input/input.scss
tylersticka: ✅ Revised comment

src/components/input/input.scss
Josh68: > With padding-left, overflowing text gets awkwardly cut off on the left:

Thanks, I missed that. This could be in the comment, too.

tylersticka:
> You may or may not agree, and it may be out of the scope of this PR, but thought I'd mention it.

@Josh68 Thank you for your feedback. I respectfully disagree, but I do think there is opportunity for refinement once the button is in use in actual interfaces. Either way, definitely out of scope for this PR.

src/components/input/input.scss
Josh68: That explains it, thanks

PR #584, opened 2020-04-02
Fix c-input:disabled color in Safari

tylersticka:
## Overview

I noticed while testing a separate component that the text color of a disabled input appeared lighter in Safari than elsewhere. I discovered a quirk in that browser where certain states prefer `-webkit-text-fill-color` over `color`. (Oy.)

Unfortunately, fixing that problem resets the browser's default `::placeholder` styles in Safari and Chrome, so I also had to specify styles there. I added a new `text_dark_muted` token to facilitate this.

On the plus side, placeholder text now has sufficient contrast. 👍 

Oh, I also noticed I had `@use` lines at the top of my input CSS that I wasn't actually using. I removed those.

### Before

Screen Shot 2020-04-02 at 3 33 54 PM

### After

Screen Shot 2020-04-02 at 3 34 28 PM

## Testing

1. View the Input/Text Elements story canvas from the deploy preview in Safari and another browser.
1. Confirm the colors are consistent for placeholder text and normal value text when disabled and not disabled.


Paul-Hebert:
@tylersticka looking at the preview deploy I'm still seeing some color differences on the placeholder text between Safari (13.0.5) and FF (74):

Screen Shot 2020-04-03 at 10 00 56 AM
Screen Shot 2020-04-03 at 10 01 27 AM

Is that expected?

tylersticka:
@Paul-Hebert Good catch! No it is not. Fix coming shortly…

tylersticka:
@Paul-Hebert Should be fixed now!

PR #585, opened 2020-04-02
Add PostCSS and SVG inlining

### LINKED ISSUE #582: SVG inlining
I went down a bit of a rabbit hole with this and I'd love some feedback on the best approach (and possibly some help implementing that approach).

## Context

There are occasionally times when it's convenient to embed an SVG element within a stylesheet. It eliminates a request, it allows us to maybe do some light find/replace of colors, and it lets us put SVGs where they can't go in markup (for example, a background image or a `<select>` caret).

When this project used PostCSS, we made use of [the PostCSS Inline SVG plugin](https://github.com/TrySound/postcss-inline-svg). I thought this would be straightforward to do in Sass, but it appears I was mistaken.

Another piece of helpful context is the state of Sass (as I understand it, this may be wrong). For years and years, Sass was written in Ruby. Then eventually it was ported to C, which came to Node as `node-sass`. More recently it's been rewritten in Dart, which is now considered the "core" version, and is on npm as simply `sass`. That one (Dart Sass) is what we're using.

## Existing Sass solutions

The first thing I did was look for an equivalent to our old Inline SVG Plugin. Here's what I found out there along with reasons why they won't work or why I have reservations about them…

- A lot of searches result in [Compass](http://compass-style.org/) because it included a lot of inlining functions, but that's a very old library that only works with Ruby Sass.
- [Node SASS Asset functions](https://www.npmjs.com/package/node-sass-asset-functions) attempts to recreate the asset functions of Compass in Node Sass, but it does not work with Dart Sass.
- [sass-asset-inliner](https://github.com/lassehaslev/sass-asset-inliner) also does not work with Dart Sass. It also appears to base64-encode SVGs, which is unnecessary, results in larger files and performs poorly in browser.
- [node-sass-svg](https://github.com/zemax/node-sass-svg) works with anything, but it doesn't allow any sort of color customization or replacement unless Sass variables are embedded in the SVG to begin with.
- [sass-inline-svg](https://github.com/haithembelhaj/sass-inline-svg) does do color replacement but it does not work with Dart Sass.
- [sass-inline-svg-utf8](https://github.com/franzheidl/sass-inline-svg-utf8) is much leaner than the previous solution, but it only customizes fill colors (no strokes) and does so in a less reliable way than our previous solutions.

## Alternatives

Based on the solutions I tried or researched, a few different approaches came to mind. I'd love to know if any of these seem like the way to go to others on the team, or if there's a different solution I'm missing.

### 1. Add PostCSS to stack and use the old plugin

PostCSS called itself a "post-processor" at launch because it was initially intended to compliment (rather than replace) tools like Sass. While that purpose has evolved somewhat, there's no reason why we couldn't process our stylesheets with both Sass and _then_ PostCSS.

The downside to this approach is that _both_ tools in the stack will be necessary for the CSS to be usable. This isn't a matter of optimizing (cssnano), polyfilling (autoprefixer) or automating repetitive selectors (foft-classes)... the CSS that Sass outputs will essentially be malformed until it's processed by a _second_ tool.

I can't decide if that's a useless philosophical concern or a valid one.

### 2. Fix one of the broken solutions

As near as I can tell, the only thing stopping plugins like `sass-inline-svg` or `sass-inline-svg-utf8` from working is that it does not allow the user to override their Sass dependency. There's nothing about what they're doing under the hood with Node Sass that Dart Sass couldn't do. Potentially we could fork and PR a version of one of these tools that fixes the issues we have with them.

### 3. Write our own Sass function or importer

We've written [our own inliner functions before](https://github.com/cloudfour/core-hbs-helpers/blob/master/lib/svg.js). It wouldn't be crazy to do it again. But it is another bit of weighty configuration to maintain within this project, and I don't know if we feel passionate enough about it to maintain it outside of this project.

### 4. Preprocess our SVGs

We're already converting SVGs to Twig format. In theory we could also convert them to one or more Sass files in a similar way.

I know there's already some concern over the Twig assets making that directory feel cluttered. This would only exacerbate that problem, although we could in theory do a sweep over _all_ of the SVG assets and compile them into a single file. But at that point it's probably as much code to accomplish (if not more) than option 3?

### 5. Repeat ourselves (for now)

Once an SVG string is in Sass, it isn't particularly difficult to encode it [from Sass functions and mixins directly](https://codepen.io/kevinweber/pen/dXWoRw). So if we wanted to avoid some complexity, we could just copy and paste the SVG sources we need into our Sass and leave a note like "make sure to update this if the source file changes."

While this is not the most elegant solution, it does have the benefit of leaving us to see how often we actually use this feature. On the other hand, it does feel like a downgrade from what we have today.

## Thoughts?

Do others have opinions? Have I overlooked a different solution?

tylersticka:
## Overview

Adds PostCSS to our Storybook CSS loader stack, and adds the PostCSS Inline SVG plugin. This will be used in the short-term for embedding a custom `<select>` caret and `<input type="checkbox">` check mark in our CSS with context-appropriate colors.

## Testing

1. Check out branch locally.
1. Install dependencies: `npm ci`
1. Run Storybook: `npm run storybook`
1. Navigate to http://localhost:6006/?path=/docs/components-input--text-elements
1. In your code editor, open `src/components/input/input.scss`
1. Add `background-image: svg-load('icons/caret-down.svg', fill=red);` to the `.c-input` block and save.
1. Observe that the input is now hideously covered in red triangles.

---

- Resolves #582


gerardo-rodriguez:
Thank you for the excellent testing instructions, @tylersticka! 🎉  

I'm taking a moment to review now. 🙂

Paul-Hebert:
>Observe that the input is now hideously covered in red triangles.

🤣

PR #587, opened 2020-04-03
Fix lint scripts and VS Code integration

tylersticka:
## Overview

Running `check-lint` or `lint` scripts would always return an error because they were not properly ignoring auto-generated files. @calebeby included a fix as part of #569, but that PR has fallen out of date. I did make two small changes to his solution: I deleted _all_ additional `.gitignore` files (so there is now only one at the root), and I simplified the `generated` string to ignore the whole directory.

Linting was also not working properly in VS Code because the eslint plugin does not yet support the `eslintConfig` property in `package.json`. This moves eslint and prettier configs to their own files, which resolves the issue and makes their definitions more consistent with that of stylelint.

TL;DR: Linting from the command line or editor should work now.

## Testing

1. Check out branch locally
1. Run `npm run check-lint`, confirm no errors.
1. Run `npm run lint`, confirm no errors.
1. In VS Code with eslint and Prettier configured, confirm error reporting and linting works as expected.

---

/CC @calebeby @Paul-Hebert


.gitignore
```diff
@@ -2,6 +2,10 @@
 dist
 storybook-static
 
+# Generated files
+/src/assets/**/*.twig
```
Paul-Hebert, line 6: I take it these are our SVG to Twig generated files? Might be nice to make a note so it's clear what exactly we're generating and ignoring there (since generally I'd assume twig files should not be ignored)

Paul-Hebert:
Code looks good. Pulling this down and trying it out

tylersticka:
> Is this happening for you as well or is this something with my setup?

That is not happening for me. 😕 

Screen Shot 2020-04-03 at 10 24 21 AM

Here are my VS Code settings in case they are helpful (the `editor.formatOn*` settings are toggled by the formatting toolbar plugin):

```json
{
    "editor.fontSize": 14,
    "editor.fontFamily": "'Fira Code', Menlo, Monaco, 'Courier New', monospace",
    "editor.fontLigatures": true,
    "workbench.colorTheme": "One Dark Pro",
    "editor.rulers": [
        80
    ],
    "projectManager.git.baseFolders": ["$home/Repos"],
    "projectManager.sortList": "Recent",
    "projectManager.git.maxDepthRecursion": 1,
    "editor.formatOnPaste": true,
    "editor.formatOnSave": true,
    "editor.formatOnType": true,
    "window.zoomLevel": 1,
    "css.validate": false,
    "less.validate": false,
    "scss.validate": false,
    "eslint.lintTask.enable": true,
    "[javascript]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    },
    "[jsonc]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    },
    "[json]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    },
    "git.autofetch": true,
    "git.confirmSync": false,
}
```

Paul-Hebert:
Thanks! I can look into that. Glad it's me and not the project :)

calebeby:
@Paul-Hebert Do you have the latest version of this VSCode extension? https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint

Paul-Hebert:
Good call @calebeby ! It looks like I had a _different_ stylelint extension installed. The version I have appears to no longer be available online. Weird🤔 

Changing extensions fixed it! Thanks!

PR #589, opened 2020-04-03
Checkbox component

### LINKED ISSUE #461: Unchecked checkboxes and radio buttons have low contrast
These form elements may be difficult to use due to their low contrast. (They get a 1.25 contrast ratio on a white background)

We should investigate using a darker color or increasing contrast some other way to make these easier to see and use (This will help accessibility for sight impaired users as well as users in high glare situations)

Screen Shot 2019-10-16 at 2 12 46 PM

tylersticka:
## Overview

Adds a `c-checkbox` class intended for `<input type="checkbox">` elements.

Originally I was going to style `radio` inputs as well, but I noticed we have _never_ used those on our current site, and it increased the size of the CSS enough that I thought it seemed off. I've also always struggled with naming a single class for both use cases… if we eventually add `c-radio`, [mixins might be a more performant solution](https://csswizardry.com/2016/02/mixins-better-for-performance/) anyway.

I originally had `checked` as a Storybook knob, but I decided to remove it because manipulating the property via Storybook does not animate. I'd rather devs click the element to see the animation.

## Screenshots

Screen Shot 2020-04-03 at 1 34 54 PM
Screen Shot 2020-04-03 at 1 35 07 PM
Screen Shot 2020-04-03 at 1 35 13 PM

## Testing

1. Try out Checkbox component in deploy preview.

---

- See #490 
- Resolves #461

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -9,7 +9,9 @@
  * 2. Modern browsers let us apply wholly custom styles to certain elements,
  *    but only when `appearance` is set to `none`. Because this property is not
  *    a finalized standard, vendor prefixes are still required.
- * 3. Without this, checkboxes appear too small unless hard-pixel sizes are
+ * 3. Some browsers will squish the checkbox in flex containers, which we never,
+ *    ever want.
+ * 4. Without this, checkboxes appear too small unless hard-pixel sizes are
  *    used (ew).
  */
 
@@ -25,7 +27,8 @@
     border-radius: sizes.$radius_medium;
     color: colors.$text-dark;
     cursor: pointer;
-    font: inherit; /* 3 */
+    flex: none; /* 3 */
+    font: inherit; /* 4 */
     height: sizes.$toggle-medium;
     margin: 0;
     padding: 0;
@@ -33,6 +36,7 @@
     transition-duration: motion.$speed-quick;
     transition-property: background-color, border-color, color;
     transition-timing-function: motion.$ease-out;
+    vertical-align: middle;
     width: sizes.$toggle-medium;
 
     /**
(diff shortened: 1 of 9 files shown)


src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,163 @@
...
+     * We use `focus-visible` here because checkboxes do not inherently require
+     * keyboard interactions.
+     */
+
+    @include focus.focus-visible {
+      box-shadow: 0 0 0 sizes.$edge-large colors.$primary-brand-lighter;
+    }
+
+    /**
+     * State: Disabled
```
Paul-Hebert, line 119: > Originally I was going to style radio inputs as well, but I noticed we have never used those on our current site, and it increased the size of the CSS enough that I thought it seemed off

This made me wonder: do we ever use disabled checkboxes on the site?

tylersticka: We do not. This might seem silly, but I like having `disabled` states in case we ever feel like enhancing a form for client-side functionality... its useful having an easy "don't click me, I'm posting data" state to switch to.

Maybe that's silly!

src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,163 @@
...
+      content: '';
+      left: sizes.$edge-medium; /* 1 */
+      opacity: 0; /* 2 */
+      position: absolute;
+      right: sizes.$edge-medium; /* 1 */
+      top: sizes.$edge-medium; /* 1 */
+      transform: scale(0); /* 2 */
+      transition-duration: motion.$speed-quick;
+      transition-property: opacity, transform;
+      transition-timing-function: motion.$ease-out;
```
Paul-Hebert, line 70: I'd be tempted to give this a little bounce when animation in, but that might be overkill.

tylersticka: I like the idea, but my understanding is that this would require [`@keyframes` to accomplish](https://easings.net/en#easeOutBounce), and it would need to be different depending on whether we were transitioning from or to a `:checked` state, which I'm not sure I can accomplish through CSS alone.

Let me know if I'm missing something obvious!

PR #591, opened 2020-04-04
Add select support to c-input

### LINKED ISSUE #490: Input pattern(s)

tylersticka:
## Overview

Adds support for `<select>` elements to `c-input`.

## Screenshots

Screen Shot 2020-04-03 at 4 57 09 PM

## Testing

1. Try it out from the deploy preview!

---

- Resolves #490


src/components/input/input.scss
```diff
@@ -51,6 +52,38 @@
...
+
+  @at-root select#{&} {
+    background-image: svg-load('icons/caret-down.svg', fill=colors.$text-dark);
+    background-position: right sizes.$control-icon-inset center;
+    background-repeat: no-repeat;
+    background-size: sizes.$control-icon-size 100%; /* 1 */
+    padding-bottom: sizes.$control-pad-vertical / 2; /* 2 */
+    padding-right: sizes.$control-icon-inset + sizes.$control-icon-size; /* 3 */
+    padding-top: sizes.$control-pad-vertical / 2; /* 2 */
+  }
```
Paul-Hebert, line 85: Thanks for the detailed comments on these CSS rules!

src/components/input/input.scss
```diff
@@ -51,6 +52,38 @@
     text-indent: sizes.$control-text-inset;
   }
 
+  /**
+   * Multi-line types should be taller
+   */
+
+  @at-root textarea#{&} {
```
Paul-Hebert, line 59: Nit: I find this harder to read than the previous `textarea.c-input`. Not a blocker by any means though

tylersticka: I agree, but I wanted all of the type-specific CSS to be together so the various states would apply at the same point in the cascade.

Issue #580, opened 2020-04-01
Set viewport add-on sizes based on breakpoint design tokens


tylersticka:
We have a viewport add-on installed that lets us easily test our components in Storybook at different sizes:

<img width="1120" alt="Screen Shot 2020-04-01 at 1 01 34 PM" src="https://user-images.githubusercontent.com/69633/78181274-eea43f00-7418-11ea-84ab-48d664c13f6f.png">

Now that we have design tokens for our standard breakpoints (see #570) we should consider configuring the add-on to use these sizes.

Josh68:
https://github.com/cloudfour/cloudfour.com-patterns/pull/597

PR #597, opened 2020-04-06
Storybook Addon Viewport - Use Design Token Breakpoints

Josh68:
## Overview

Pulls in Theo breakpoints and assigns them as the viewports (with `height: 100%`) for `@storybook/addon-viewport`.

Retains the `defaultViewport` (Responsive, which is 100% x 100%) on page load and which is what the "Reset viewport" link returns to after a specific breakpoint viewport size is selected.

Note: The tokens have no "pretty" names, so the addon is defaulting to the names used elsewhere (e.g., `xs`, `m`, `l`).

If pretty names are desired, we should discuss where best to store / map those.

## Screenshots

viewports

## Testing

1. [Preview theme](https://deploy-preview-597--cloudfour-patterns.netlify.com/)
1. Verify that you can select among the viewports defined in Theo
1. Verify that the default is full-screen responsive and that you can return to that default by choosing "Reset viewport"

---

- Fixes [#580](https://github.com/cloudfour/cloudfour.com-patterns/issues/580)

/CC @tylersticka @dromo77


.storybook/helpers/breakpoint-viewports.js
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * List of viewports to use with addon-viewport, based on Theo breakpoints tokens
+ */
+export const breakpointViewports = Object.keys(breakpoints).map((name) => {
+  return {
+    name,
+    styles: {
+      width: breakpoints[name],
+      height: '100%',
+    },
```
Josh68, line 12: https://www.npmjs.com/package/@storybook/addon-viewport#viewport-model

The Viewport Model indicates `type` ('desktop' | 'mobile' | 'tablet' | 'other') is not optional, but without strict typing, nothing here is broken. I wonder whether we need to indicate what `type` each of these is supposed to represent or whether we should just ignore that property.

.storybook/preview.js
```diff
@@ -27,3 +28,12 @@ for (let i = -3; i <= 6; i++) {
 }
 addDecorator(withPaddings);
 addParameters({ paddings });
+/**
+ * Default viewport is Responsive: 100% x 100%. When a breakpoint is chosen, this
+ * will show as the Reset Viewport option
+ */
```
tylersticka, line 34: It's unclear to me what this comment is trying to convey relative to the adjacent code.

.storybook/preview.js
```diff
@@ -2,6 +2,7 @@ import { addDecorator, addParameters } from '@storybook/html';
 import { withA11y } from '@storybook/addon-a11y';
 import { withPaddings } from 'storybook-addon-paddings';
 import * as colors from '../src/design-tokens/colors.yml';
+import { breakpointViewports } from './helpers/breakpoint-viewports';
```
tylersticka, line 5: Can we include this logic alongside the setting itself? I find it odd to have _just_ these nine lines of configuration in a separate file.

.storybook/preview.js
Josh68: Sorry, no it's not necessarily obvious. We don't set this viewport, but it's the stock default, and I thought it would be good to indicate that somewhere. In other words, if we don't actually set the default, this is what it will be on page load (100% x 100%). If I rephrased slightly, would it make sense, and do you think it's worth adding?

```

File: .storybook/preview.js (hunk 3 above)
Comment (Josh68) [reply .storybook/preview.js:5]: Personal preference and something we've been doing recently in other projects - trying to keep small, focused modules of code to avoid having files grow too large. I'm fine with moving this back into `preview.js` if you like that better.

Comment (Josh68) [reply .storybook/preview.js]: I have done this, and also updated comments.

File: .storybook/preview.js [hunk 4]
```diff
@@ -27,3 +28,24 @@ for (let i = -3; i <= 6; i++) {
...
+      width: breakpoints[name],
+      height: '100%',
+    },
+    type: 'other',
+  };
+});
+/**
+ * The default viewport (which is not overridden here) is Responsive: 100% x 100%.
+ * After a breakpoint is chosen, the "Reset Viewport" option will use this default.
+ */
```
tylersticka, line 46: I understand what this is saying now, but I still don't understand its relevance to the code below. It seems like it's just describing a feature of the add-on. Is there something about the "Reset Viewport" feature that impacted how you wrote lines 47–51 that I'm missing?

Josh68:
@tylersticka, when you have a moment, I've updated according to your comments. Let me know if the revised comment makes more sense, or if you think it's not needed. Thanks

.storybook/preview.js
Josh68: @tylersticka, I'm just trying to explain that the default viewport (aka, `defaultViewport` in the config, which is an option) is not something we're setting, and that it's both the default on page load and the viewport that is used when you use "Reset viewport" (which shows up after you've selected one of the breakpoint-based viewport options).

I tried overriding the default at one point, but decided it was best to leave it as is, and this option wasn't clear to me until I read the addon's docs. If you think this comment isn't helpful, should I just remove it?

Josh68: I've just removed this comment.

Josh68:
@tylersticka, I simply removed the comment that wasn't making sense to you. I felt it made sense, but as it didn't to you, no need to keep it. Not sure whether there was anything else you wanted here, or if it's good to approve for merge. Thanks

.storybook/preview.js
```diff
@@ -27,3 +28,20 @@ for (let i = -3; i <= 6; i++) {
...
+    styles: {
+      width: breakpoints[name],
+      height: '100%',
+    },
+    type: 'other',
+  };
+});
+addParameters({
+  viewport: {
+    viewports: breakpointViewports,
```
tylersticka, line 45: @Josh68 Could we prepend this to the existing list of minimal viewports? See: https://github.com/storybookjs/storybook/tree/master/addons/viewport#use-detailed-set-of-devices

.storybook/preview.js
```diff
@@ -27,3 +28,20 @@ for (let i = -3; i <= 6; i++) {
 }
 addDecorator(withPaddings);
 addParameters({ paddings });
+
+// Create viewports using widths defined in design tokens
+const breakpointViewports = Object.keys(breakpoints).map((name) => {
+  return {
+    name,
```
tylersticka, line 35: Could we set the name to be something like:

```js
{
  name: `breakpoints.${name}`
}
```

This would align with our usage of breakpoints in Sass, which might be more helpful?

Josh68: That sounds good. Thanks for the suggestion.

.storybook/preview.js
Josh68: Yes, I wasn't sure whether we wanted that, but assumed we'd be replacing. I'll use the other approach

Josh68:
@tylersticka, thanks for the feedback. All updated and ready for re-review.

.storybook/preview.js
```diff
@@ -27,3 +29,23 @@ for (let i = -3; i <= 6; i++) {
 }
 addDecorator(withPaddings);
 addParameters({ paddings });
+
+// Create viewports using widths defined in design tokens
+const breakpointViewports = Object.keys(breakpoints).map((name) => {
+  return {
+    name: `breakpoints.${name}`,
```
tylersticka, line 36: Missing a dollar sign in the output (should display as `breakpoints.$xs`, see: https://v-next--cloudfour-patterns.netlify.com/?path=/docs/design-tokens-breakpoint--page)

Josh68: Ah, I read over that too quickly. Saw a dollar sign in a literal and went with it. Fixing now.

Josh68: Deployed with this update.

Josh68:
@tylersticka, all updated and pulled in the latest from `v-next`. Hope this is good to go.

PR #599, opened 2020-04-07
Container object

### LINKED ISSUE #571: Container pattern
While working on issue #547, there was a discussion about how we currently use our `u-contain*` utilities. To summarize, we use `u-contain-spread` and `u-contain-prose` to establish maximum widths for content.  However, these utility classes preform multiple functions. Not only do they establish maximum widths, but they also center content. Because utilities are meant to only do one thing there may be a better approach, a container pattern. 

Creating the container pattern as an object would allow us to have the opinionated styles that we need.

>In his article on namespaces, Harry Roberts describes the "objects" namespace as the best place for "things like layout, wrappers and containers." They are more opinionated than utilities, but they are nothing without content to fill them.

See #547 for the full discussion.

dromo77:
## Overview

This PR adds a new container object. It includes a modifier for prose content and updates the `max_width_prose` token value.

## Screenshots

container

- Fixes #571


src/objects/container/container.stories.mdx
```diff
@@ -0,0 +1,50 @@
...
+# Container
+
+<Preview>
+  <Story
+    name="Basic"
+    height="300px"
+    parameters={{ paddings: { disabled: true } }}
+  >
+    {basicDemo({
+      prose: boolean('Add o-container--prose modifier to the container', false),
```
tylersticka, line 18: Could we shorten this label so it looks a bit nicer when the sidebar is aligned to the right?

src/objects/container/container.stories.mdx
```diff
@@ -0,0 +1,50 @@
+import { Story, Preview, Meta } from '@storybook/addon-docs/blocks';
+import { withKnobs, boolean } from '@storybook/addon-knobs';
+import basicDemo from './demo/basic.twig';
+import './demo/styles.scss';
+import './container.scss';
+
+<Meta title="Objects/Container" decorators={[withKnobs]} />
+
+# Container
```
tylersticka, line 9: Can we add some more explanatory text? What is this pattern intended for? What's the difference between prose and the default? If the default is not for prose, what is it for? That sort of stuff would be helpful.

src/objects/container/container.scss
```diff
@@ -0,0 +1,16 @@
+@use "../../design-tokens/layout.yml";
+
+.o-container__content {
```
tylersticka, line 3: I notice you are only styling the `__content`. What is the outer container for?

src/objects/container/container.stories.mdx
```diff
@@ -0,0 +1,50 @@
...
+    name="Themes (Dark)"
+    height="300px"
+    parameters={{ themes: { disable: true }, paddings: { disabled: true } }}
+  >
+    {basicDemo({
+      prose: boolean('Add o-container--prose modifier to the container', false),
+      theme: 'dark',
+    })}
+  </Story>
+</Preview>
```
tylersticka, line 50: It doesn't look like the container has any theme-specific styles, so it's probably overkill to include this here. One can always opt into the dark theme from the canvas view if they want to test that it works there.

src/objects/container/container.scss
dromo77: I imagine we will the ability to add margins to the container. Although the outer container doesn't have any styles, if we removed it and added space utilities to the `__content` it negates the margins that help to center the content.

tylersticka: That's a fine explanation.

If I were a newcomer to the project, where would I find that explanation or direction for where to add margin/padding?

dromo77: Do you think that's something that should be added to the container description?

tylersticka: I think it'd be good to mention in the story docs (the `.mdx` file) _and_ probably as a comment in this file to explain why we're _starting_ with `o-container__content`. That way it's clearly communicated to CSS developers _and_ to template designers!

dromo77: Sounds good thanks, I updated the story docs already, but I'll add an explanation to this file as well.

Issue #494, opened 2020-02-21
Thumbnail patterns


dromo77:
We have several different styles of thumbnails, some are within a contained shape and some have borders. These may need to be different patterns, especially the calendar.

<img width="615" alt="Thumbnails" src="https://user-images.githubusercontent.com/42841342/75062496-c2031c00-5497-11ea-968f-ef2fbfe19bac.png">

Paul-Hebert:
I wonder how/if this overlaps with the media object pattern?

https://github.com/cloudfour/cloudfour.com-patterns/issues/474

tylersticka:
It very well may!

And do these overlap with #477?

These inventory issues are starting points. I've been taking the strategy of hoarding the ones I see as overlapping or related, and either creating a larger issue encompassing them, or just tackling all three in a WIP component.

Something we haven't tried yet (but works fine in my local testing) is prototyping in Storybook, so that's an approach we could take to some of these as well!

tylersticka:
## Tasks

- [x] Embed
- [x] Rounded
- [x] Circle
- [ ] Date

PR #601, opened 2020-04-08
Embed object

tylersticka:
## Overview

Introduces an `o-embed` class for embedding imagery and other media responsively while maintaining aspect ratios.

This deviates from [our Flex Embed component](https://cloudfour-patterns.netlify.com/patterns/components/flex-embed.html) in a few ways:

- It removes the word `flex-`, which felt a little confusing since it doesn't involve flexbox at all.
- It foregoes the separate ratio element in favor of a pseudo element. In reality we never used the `max-height` features of this technique.
- It targets child elements directly instead of requiring an element class. While this is a little messier, in reality we were already doing this because it is very challenging modifying classes of child elements consistently when content is defined via a CMS.
- The modifier classes use nicknames defined in tokens and based on [standard aspect ratios](https://en.wikipedia.org/wiki/Aspect_ratio_(image)).
- The modifier classes support responsive breakpoints, useful for interactive embeds like CodePens or for art direction.
- In addition to modifier classes, aspect ratio may be defined via a custom property. This allows for arbitrary one-off ratios without requiring any new CSS.
- It is now an object instead of a component because it defines layout (no appearance without content).

## Screenshots

Responsive modifier classes in action:

responsive

The docs:

Screenshot_2020-04-08 Storybook

---

- See #494


src/objects/embed/embed.scss
```diff
@@ -0,0 +1,119 @@
...
+ * Makes ratio modifier classes based on the `aspect-ratio` and `breakpoint`
+ * design tokens. Intended to be included from within the base class.
+
+ * We iterate over breakpoints before ratios so wider breakpoints will be later
+ * in the cascade. If we iterated over aspect ratios first, then a narrower
+ * breakpoint would override a wider one if its aspect ratio came first in our
+ * design token file.
+ *
+ * 1. For some reason Sass does not compile `$ratio` without this string
+ *    interpolation syntax. I'm not entirely sure why.
```
Paul-Hebert, line 24: I ran into this the other day. This might be a helpful link: https://sass-lang.com/documentation/breaking-changes/css-vars

src/objects/embed/embed.scss
```diff
@@ -0,0 +1,119 @@
...
+    @if $ratio != $default-aspect-ratio {
+      &--#{$name} {
+        --aspect-ratio: #{$ratio}; /* 1 */
+      }
+    }
+  }
+  @each $suffix, $breakpoint in meta.module-variables('breakpoint') {
+    @media (min-width: $breakpoint) {
+      @each $name, $ratio in meta.module-variables('aspect-ratio') {
+        &--#{$name}\@#{$suffix} {
```
Paul-Hebert, line 40: I like this usage of `@` in the class name. I was confused by `\@#` at first though and wondered if it was some weird SASS syntax. I assume the `\` is just escaping the `@`? It might be nice to add a note or example of what this will output

src/objects/embed/embed.scss
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Fallback behavior: Browsers that do not support custom properties will use
+ * the element's intrinsic aspect ratio. These styles cause the element to fill
+ * the available width.
+ */
+
+.o-embed > * {
+  display: block;
+  width: 100%;
+}
```
Paul-Hebert, line 65: Question: Down below we target `picture > img` specially. Do we need to do something similar in our fallback? Though I guess IE11 doesn't support `picture` anyways...

tylersticka: Great question!

src/objects/embed/embed.scss
tylersticka: I had a note here originally and decided it wasn't necessary, I should have trusted my instincts!

src/objects/embed/embed.scss
tylersticka: That is a helpful link!

Issue #510, opened 2020-03-02
Messaging and alert pattern(s)


tylersticka:
<img width="1378" alt="Screen Shot 2020-03-02 at 2 21 01 PM" src="https://user-images.githubusercontent.com/69633/75723420-250f6280-5c91-11ea-8d18-3b90f94826a5.png">

Currently used to convey an offline state, but accessible contrast needs work.

derekshirk:
This looks like a good first issue for me to get my feet wet with.

PR #604, opened 2020-04-08
Pattern: message

derekshirk:
## Overview

Introduces a (fullwidth) message pattern ideal for offline network status notifications.

This PR is open to both design and development feedback. 

## Screenshots

Screen Shot 2020-04-30 at 11 38 10 AM

Close button `:hover` style

message-with-close-hover

Close button with active `:focus`

Screen Shot 2020-05-04 at 3 07 59 PM

## Testing

1. View Netlify deploy preview. 
https://deploy-preview-604--cloudfour-patterns.netlify.com/?path=/docs/components-message--message

2. Confirm all Storybook accessibility checks pass.

---

Issue #510


src/components/message/message.scss
```diff
@@ -0,0 +1,60 @@
...
+ * 1. Reset default button styles (remove once reset/button CSS is inplace)
+ * 2. Close icon should be the same color as the parent text color.
+ * 3. Easy way to collapse extra whitespace between button edge and icon.
+ * 4. Make icon a little larger than the parent text size.
+ * 5. Put some space between message and close button.
+ * 6. Removes padding so that element maintains consistent spacing with other
+ *    inner elments
+ */
+
+.c-message-close {
```
tylersticka, line 51: Should use BEM syntax? (`.c-message__close`)

src/components/message/message.scss
```diff
@@ -0,0 +1,60 @@
...
+ * Inherit parent text color for links when blue is hideous against background.
+ */
+
+.c-message--warning a {
+  color: inherit;
+}
+
+/**
+ * 1. Reset default button styles (remove once reset/button CSS is inplace)
+ * 2. Close icon should be the same color as the parent text color.
```
tylersticka, line 43: Did you already try using `c-button c-button--link`?

src/components/message/message.scss
```diff
@@ -0,0 +1,60 @@
...
+ * 3. Condense line height in case it breaks to multiple lines.
+ *
+ * @TODO
+ * 1. Style `:focus` state
+ */
+
+.c-message {
+  align-items: center; /* 1 */
+  background: colors.$gray-light;
+  border-top: 4px solid colors.$gray-dark;
```
derekshirk, line 19: Can I do something "smarter" than `4px` here? Note to self to look at design token options.

src/components/message/message.twig
```diff
@@ -0,0 +1,23 @@
+{# 
+  @TODO
+  1. Generate an example of a message with a decorative icon.
+  #}
+
+{% set tagName = tagName | default('div') %}
+
+<{{tagName}} aria-describedby="MessageInfo" {% if not stopAnnouncements %} aria-live="polite"{% endif %} class="c-message{% if class %} {{class}}{% endif
+%}" tabindex=”0”{% if isWarning %} role="alert"{% else %} role="status"{% endif
+%}>
```
tylersticka, line 10: - I worry that these attributes will slow down approval of the basic CSS/markup pattern because they are so intertwined with how this pattern may be used in context. Is there any way we can simplify for now?
- I see some smart quotes where dumb quotes should be.

src/components/message/message.twig
```diff
@@ -0,0 +1,23 @@
...
+  <p id="MessageInfo">
```
tylersticka, line 11: Including an ID in markup by default prevents the pattern from being used more than once in a page.

src/components/message/message.twig
```diff
@@ -0,0 +1,23 @@
...
+    {% block content %}
+      {{label | default('Hi, I\'m a message!')}}
+    {% endblock %}
+  </p>
+  {% if includeClose %}
+    <button aria-label="Close message" class="c-message-close" type="button">
```
tylersticka, line 17: Close message?

src/components/message/demo/warning-message.twig
```diff
@@ -0,0 +1,13 @@
+{% embed '../message.twig' with {
+  includeClose: true,
+  isWarning: true,
+  class: 'c-message--warning' } 
+%}
+  <p id="MessageInfo">
+    {% block content %}
+      You seem to be offline at the moment.
+      <span class="u-hidden u-md-inlineBlock">We&rsquo;ll keep trying to
```
tylersticka, line 9: - These utilities don't exist yet. Can we remove this content?
- Can we replace `&rsquo;` with a unicode `’`?

src/components/message/message.scss
```diff
@@ -0,0 +1,60 @@
...
+ * 2. Push message and close button as far apart as possible.
+ * 3. Condense line height in case it breaks to multiple lines.
+ *
+ * @TODO
+ * 1. Style `:focus` state
+ */
+
+.c-message {
+  align-items: center; /* 1 */
+  background: colors.$gray-light;
```
tylersticka, line 18: Should we use `$gray-lighter` for consistency with table backgrounds?

src/components/message/message.scss
tylersticka: Can you explain the design thinking behind the top border? It feels strange to me.

src/components/message/message.scss
```diff
@@ -0,0 +1,60 @@
...
+  color: colors.$text-dark;
+  display: flex; /* 1, 2 */
+  justify-content: space-between; /* 2 */
+  line-height: ms.step(1, 1); /* 3 */
+  padding: ms.step(-8) ms.step(0);
+}
+
+.c-message--warning {
+  background: colors.$fuchsia;
+  border-top: 4px solid colors.$fuchsia-darker;
```
tylersticka, line 29: Consider only overrriding necessary properties, for example `background-color` and `border-color`.

src/components/message/message.scss
```diff
@@ -0,0 +1,60 @@
...
+  color: colors.$text-light-emphasis;
+}
+
+/**
+ * Inherit parent text color for links when blue is hideous against background.
+ */
+
+.c-message--warning a {
+  color: inherit;
+}
```
tylersticka, line 39: Would this be better accomplished via a theme?

src/components/message/message.scss
derekshirk: Sure. My intent was to add some visual emphasis to the pattern. 

That being said I'm not too attached to it and will be happy to remove them. 

@tylersticka

src/components/message/message.scss
derekshirk: 🤔 I did not see a `c-button--link` or related selector in `button.scss`.

src/components/message/message.twig
derekshirk: @tylersticka 

> I worry that these attributes will slow down approval of the basic CSS/markup pattern because they are so intertwined with how this pattern may be used in context.

> Is there any way we can simplify for now?

I see two options: 

1. I could remove all accessibility related attributes
2. I could remove some of the attributes. 

If the later, I'm not sure which ones should be considered default. 

@tylersticka

src/components/message/message.scss
derekshirk: 1. See [previous follow-up comment](https://github.com/cloudfour/cloudfour.com-patterns/pull/604#discussion_r407718436)

2. 👍

src/components/message/message.scss
tylersticka: Sorry, I meant `c-button--tertiary`.

src/components/message/message.twig
tylersticka: Let's remove all of them for now.

src/components/message/message.scss
derekshirk: 🤔 Perhaps?

Since this pattern is likely to sit above all all page content, instead of within page content, I wasn't sure theme variants were appropriate here.

src/components/message/message.scss
derekshirk: Example of (close) button with `c-button--tertiary`

Screen Shot 2020-04-13 at 3 24 18 PM

Example of (close) button with `c-message__close`

Screen Shot 2020-04-13 at 3 24 38 PM

derekshirk:
@tylersticka This is ready for another review. Thank you.

src/components/message/message.scss
```diff
@@ -0,0 +1,57 @@
...
+@use "../../mixins/ms";
+@use "../../mixins/unit";
+
+/**
+ * 1. Display message and close button in a centered row.
+ * 2. Push message and close button as far apart as possible.
+ * 3. Condense line height in case it breaks to multiple lines.
+ *
+ * @TODO
+ * 1. Style `:focus` state
```
tylersticka, line 14: Can this be one line? I spent a few moments trying to figure out how this `@todo` related to comment 1, then I realized [you weren't using sassdoc conventions](http://sassdoc.com/annotations/#todo).

src/components/message/message.scss
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+.c-message {
+  align-items: center; /* 1 */
+  background-color: colors.$gray-lighter;
+  color: colors.$text-dark;
+  display: flex; /* 1, 2 */
+  justify-content: space-between; /* 2 */
+  line-height: ms.step(1, 1); /* 3 */
+  padding: ms.step(1) ms.step(0);
```
tylersticka, line 24: It feels a little off visually to me to have less padding on the sides than the top.

src/components/message/message.scss
```diff
@@ -0,0 +1,57 @@
...
+}
+
+.c-message--withClose {
+  padding-right: ms.step(0);
```
tylersticka, line 28: Is this the same amount of padding as `c-message` already has on the right?

src/components/message/message.scss
```diff
@@ -0,0 +1,57 @@
+@use "../../design-tokens/colors.yml";
+@use "../../mixins/focus";
+@use "../../design-tokens/font.yml";
+@use "../../design-tokens/sizes.yml";
+@use "../../mixins/ms";
+@use "../../mixins/unit";
+
+/**
+ * 1. Display message and close button in a centered row.
```
tylersticka, line 9: I'm not sure this comment is necessary.

src/components/message/message.scss
```diff
@@ -0,0 +1,57 @@
...
+ * 2. Push message and close button as far apart as possible.
+ * 3. Condense line height in case it breaks to multiple lines.
```
tylersticka, line 11: This comment should include the _why_. The why isn't really "in case it breaks to multiple lines," the why is probably more like "Our default `line-height` may visually compete with this element's padding, so we condense it."

src/components/message/message.scss
```diff
@@ -0,0 +1,57 @@
+@use "../../design-tokens/colors.yml";
+@use "../../mixins/focus";
+@use "../../design-tokens/font.yml";
+@use "../../design-tokens/sizes.yml";
+@use "../../mixins/ms";
+@use "../../mixins/unit";
```
tylersticka, line 6: Is this ever used?

src/components/message/message.scss
```diff
@@ -0,0 +1,57 @@
+@use "../../design-tokens/colors.yml";
+@use "../../mixins/focus";
+@use "../../design-tokens/font.yml";
```
tylersticka, line 3: Is this ever used?

src/components/message/message.scss
```diff
@@ -0,0 +1,57 @@
...
+ * 1. Button background-color should inherit the same color as its parent.
+ * 3. Close icon should be the same color as the parent text color.
+ * 4. Easy way to collapse extra whitespace between button edge and icon.
+ * 5. Make icon a little larger than the parent text size.
+ * 6. Put some space between message and close button.
+ * 7. Removes padding so that element maintains consistent spacing with other
+ *    inner elments
+ */
+
+.c-message__close {
```
tylersticka, line 41: I would love to see this element's click area be larger, ideally filling the height of the containing element.

src/components/message/message.scss
tylersticka: Also, I'm not sure what focus state the message would have? The close button already has one. 🤔

src/components/message/message.scss
```diff
@@ -0,0 +1,57 @@
...
+  background-color: inherit; /* 1 */
+  border-color: transparent;
+  border-radius: sizes.$radius-medium;
+  border-style: solid;
+  border-width: 0;
```
tylersticka, line 46: Is it necessary to style so much of the border if its width is `0`?

src/components/message/message.scss
```diff
@@ -0,0 +1,57 @@
...
+  color: inherit; /* 4 */
+  cursor: pointer;
+  display: flex; /* 5 */
```
tylersticka, line 49: It's unclear how comment 5 relates to this.

src/components/message/message.stories.mdx
```diff
@@ -0,0 +1,38 @@
...
+<Meta title="Components/Message" decorators={[withKnobs]} />
+
+# Message
+
+Use to convey general information or messages that suggest action for users.
+Message content should focus on a single theme, piece of information, or
+required action to avoid overwhelming users.
+
+Additionally, the message pattern should be used thoughtfully and sparingly for
+only the most important information.
```
tylersticka, line 16: I think we can condense, and we should include the inspiration for this component so its purpose is not hypothetical. Example:

> Containing element for a short, actionable chunk of text. Originally designed to notify the user when they've lost their network connection.

I'm not sure that's exactly right, but hopefully you get the idea.

src/components/message/message.stories.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+<Preview>
+  <Story name="Message" height="120px">
```
tylersticka, line 19: Can we give this story a `name` that's distinct? "Text" or "Default" or "Without Close"?

src/components/message/message.stories.mdx
```diff
@@ -0,0 +1,38 @@
...
+    })}
+  </Story>
+</Preview>
+
+## Message with close button
+
+Most messages should allow users to dismiss them with a close button.
+
+<Preview>
+  <Story name="Message with close" height="120px">
```
tylersticka, line 32: I'd rather see more of this story's uniqueness in the name (example: "With Close Button") than repeating the name of the component.

src/components/message/message.twig
```diff
@@ -0,0 +1,21 @@
+{# 
+  @TODO
+  1. Generate an example of a message with a decorative icon.
+  #}
```
tylersticka, line 4: This feels more relevant to the story file than this template to me.

src/components/message/message.twig
```diff
@@ -0,0 +1,21 @@
...
+
+{% set tagName = tagName | default('div') %}
+
+<{{tagName}} class="c-message{% if class %} {{class}}{% endif %}">
+  {% block content %}
+  <p>
+    {{label | default('Hi, I\'m a message!')}}
+  </p>
```
tylersticka, line 12: Nit: Please indent when nesting in block tags.

src/components/message/message.twig
```diff
@@ -0,0 +1,21 @@
...
+  {% endblock %}
+  {% if includeClose %}
+  <button aria-label="close" class="c-message__close u-pad-none" type="button">
+    {% include '../../assets/icons/close.svg.twig' with {
+        class: 'c-icon'
+      } %}
+  </button>
```
tylersticka, line 19: Nit: Please indent when nesting in block tags.

src/components/message/message.twig
```diff
@@ -0,0 +1,21 @@
...
+{% set tagName = tagName | default('div') %}
+
+<{{tagName}} class="c-message{% if class %} {{class}}{% endif %}">
+  {% block content %}
+  <p>
+    {{label | default('Hi, I\'m a message!')}}
+  </p>
+  {% endblock %}
+  {% if includeClose %}
+  <button aria-label="close" class="c-message__close u-pad-none" type="button">
```
tylersticka, line 15: Why is `u-pad-none` a utility when all the other button styles are defined in `c-message__close`?

tylersticka: @Paul-Hebert I feel like recently you shared some information on when to use `aria-label` versus visually hidden text? Am I imagining that?

src/components/message/message.twig
```diff
@@ -0,0 +1,21 @@
...
+    {% include '../../assets/icons/close.svg.twig' with {
+        class: 'c-icon'
+      } %}
```
tylersticka, line 18: @Paul-Hebert Do you remember if it's best to use `aria-hidden="true"` or `role="presentation"` on icons in buttons like this?

src/components/message/message.twig
```diff
@@ -0,0 +1,21 @@
...
+  </button>
+  {% endif %}
+</{{tagName}}>
```
tylersticka, line 21: Missing newline at tend of file. (The editor config plugin manages this for me.)

src/components/message/message.twig
Paul-Hebert: I think visually hidden text within the button would be preferable here.

The primary downside of `aria-label` is that it's [not reliably translated into other languages](https://heydonworks.com/article/aria-label-is-a-xenophobe/). (I think the sotry with Google Translate has gotten better but still isn't great)

So a screen reader user reading our site in Spanish would read the button text as "close" instead of "cerrar." If we used visually hidden text instead they'd hear "cerrar."

src/components/message/message.twig
Paul-Hebert: If we want this to be hidden from screen readers we should use `aria-hidden`. 

My understanding is that `role="presentation"` tells a screen reader “ignore the semantics of this element” while `aria-hidden="true"` says “ignore this completely.”

https://timwright.org/blog/2016/11/19/difference-rolepresentation-aria-hiddentrue/

src/components/message/message.twig
derekshirk: Would this be the appropriate comment style _and_ location for this comment within the (.mdx) story file?

Screen Shot 2020-05-01 at 3 07 59 PM

src/components/message/message.twig
```diff
@@ -0,0 +1,18 @@
...
+      {{label | default('Hi, I\'m a message!')}}
+    </p>
+  {% endblock %}
+  {% if includeClose %}
+    <button class="c-message__close" type="button">
+      {% include '../../assets/icons/close.svg.twig' with {
+          class: 'c-icon',
+          aria_hidden: true
+        } %}
+      <span class="u-hidden-visually">Close</span>
```
gerardo-rodriguez, line 15: Might we be able to push this a bit further? 🔍  

If I were to close my eyes using a screen-reader and tabbed onto this button, I'd hear "close". I might start wondering though, what am I closing? 🤔 

If I can see, I visually understand the context _around_ the close button. It is clear that the close button resides within the message/notification element. 

Perhaps we could add some extra context so that screen reader users can receive a more equivalent experience? What about something like "Close the notification"? 🙂

src/components/message/message.scss
gerardo-rodriguez: Great idea, @tylersticka! I agree!

We can also reference the [Web Content Accessibility Guidelines (WCAG) 2.1](https://www.w3.org/TR/WCAG21/#target-size) in this instance which suggests a minimum of at least 44 by 44 CSS pixels in size. 🙂

derekshirk:
I have addressed all current feedback. I will review this again myself Monday morning before re-requesting review. Thank you everyone for your feedback!

derekshirk:
@gerardo-rodriguez @Paul-Hebert @tylersticka, 

I am tagging each of you because you had previously commented on this PR. 

Ready for re-review.

src/components/message/demo/default-message.twig
```diff
@@ -0,0 +1,8 @@
+{% embed '../message.twig' %}
+  {% block content %}
+    <p>
+      Messages contain short, consice information used to highlight important
```
tylersticka, line 4: Typo: concise

src/components/message/demo/message-with-close.twig
```diff
@@ -0,0 +1,7 @@
+{% embed '../message.twig' with { includeClose: true } %}
+  {% block content %}
+    <p>
+      Message with Close Button
```
tylersticka, line 4: Nit: Capitalization seems odd?

src/components/message/demo/message-with-close.twig
```diff
@@ -0,0 +1,7 @@
+{% embed '../message.twig' with { includeClose: true } %}
```
tylersticka, line 1: Nit: I am _woefully_ bad at remembering this and there are merged examples of the contrary, but Twig files should generally favor [underscores over camelcase](https://twig.symfony.com/doc/3.x/coding_standards.html).

(I've got a to-do to create an issue for the existing violations to this, but I keep thinking I need to write this info down first, and then something else comes up... 😅)

src/components/message/message.scss
```diff
@@ -0,0 +1,55 @@
...
+@use "../../mixins/ms";
+
+/**
+ * 1. Push message and close button as far apart as possible.
+ * 3. When text breaks onto multiple lines our default line-height visually
+ *    competes with this element's padding, so we condense it a little.
+ */
+
+.c-message {
+  align-items: stretch;
```
tylersticka, line 13: I thought `stretch` was the default? 🤔

src/components/message/message.scss
```diff
@@ -0,0 +1,55 @@
+@use "../../design-tokens/colors.yml";
+@use "../../mixins/focus";
+@use "../../design-tokens/sizes.yml";
+@use "../../mixins/ms";
+
+/**
+ * 1. Push message and close button as far apart as possible.
+ * 3. When text breaks onto multiple lines our default line-height visually
```
tylersticka, line 8: What happened to `2`?

src/components/message/message.scss
```diff
@@ -0,0 +1,55 @@
...
+ *    competes with this element's padding, so we condense it a little.
+ */
+
+.c-message {
+  align-items: stretch;
+  background-color: colors.$gray-lighter;
+  color: colors.$text-dark;
+  display: flex; /* 1 */
+  justify-content: space-between; /*  */
```
tylersticka, line 17: Empty comment or ASCII art?

src/components/message/message.scss
```diff
@@ -0,0 +1,55 @@
...
+  margin-left: ms.step(1); /* 6 */
+  padding: 0 ms.step(1);
+
+  @include focus.focus-visible {
+    box-shadow: 0 0 0 sizes.$edge-large colors.$primary-brand-lighter;
+  }
+}
+
+.c-message__close:hover {
+  .c-icon {
```
tylersticka, line 51: It's not really in the spirit of our guidelines to directly style a separate component from another. Some alternatives to consider:

1. Add a new class to the `c-icon` instance, for example `c-message__close-icon`. ([More info](https://www.smashingmagazine.com/2016/06/battling-bem-extended-edition-common-problems-and-how-to-avoid-them/#4-cross-component-components))
1. Apply the transform to `c-message__close` directly.

src/components/message/message.scss
```diff
@@ -0,0 +1,55 @@
...
+    transform: scale(0.9);
```
tylersticka, line 52: It feels a little unpleasant to me that the icon is _shrinking_ when I interact with it. I'd also love to see any transitions animate.

src/components/message/message.scss
```diff
@@ -0,0 +1,55 @@
...
+    transform-origin: center;
```
tylersticka, line 53: I believe the initial value of `transform-origin` is `50% 50%`, which is equivalent?

src/components/message/message.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+# Message
+
+A containing element intended for short, actionable user messages. Originally
+designed to notify the user of offline network connection status.
+
+<Preview>
+  <Story name="Default" height="120px">
+    {messageDemo({
+      label: text('Message text'),
+      includeClose: boolean('Include close button'),
```
tylersticka, line 20: Consider making the label the same as the property name so it's a more helpful reference

src/components/message/message.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+# Message
+
+A containing element intended for short, actionable user messages. Originally
+designed to notify the user of offline network connection status.
+
+<Preview>
+  <Story name="Default" height="120px">
+    {messageDemo({
+      label: text('Message text'),
```
tylersticka, line 19: This doesn't appear to function?

src/components/message/message.stories.mdx
```diff
@@ -0,0 +1,36 @@
+import { Story, Preview, Meta } from '@storybook/addon-docs/blocks';
+import { withKnobs, text, boolean } from '@storybook/addon-knobs';
+import messageDemo from './demo/default-message.twig';
+import messageWithCloseDemo from './demo/message-with-close.twig';
+import './message.scss';
+
+<!-- @TODO Generate an example of a message with a decorative icon. -->
```
tylersticka, line 7: If this is more of a feature than a code change, it might be better as a new issue or a todo on an existing issue?

src/components/message/message.scss
derekshirk: 789, then 7 was still hungry.

src/components/message/message.stories.mdx
derekshirk: If I've understood this correctly, it the knob should read as follows: 

Screen Shot 2020-05-04 at 4 20 37 PM

tylersticka: I mean if the property is `includeClose`, the label should also be `includeClose`?

src/components/message/message.stories.mdx
derekshirk: I decided to add a new issue: 
https://github.com/cloudfour/cloudfour.com-patterns/issues/684

derekshirk:
@tylersticka Thank you for the continued review. I have addressed the most current batch of feedback.

src/components/message/message.scss
```diff
@@ -0,0 +1,58 @@
...
+  line-height: ms.step(1, 1); /* 2 */
+}
+
+.c-message__content {
+  padding: ms.step(1);
+}
+
+/**
+ * 1. Button background-color should inherit the same color as its parent.
+ * 3. Close icon should be the same color as the parent text color.
```
tylersticka, line 27: - These two comments are still missing the "why."
- Missing 2

src/components/message/message.scss
```diff
@@ -0,0 +1,58 @@
...
+
+.c-message__close-icon {
+  transition-duration: motion.$speed-quick;
+  transition-property: transform;
+  transition-timing-function: motion.$ease-out;
+}
+
+.c-message__close:hover .c-message__close-icon {
+  transform: scale(1.1);
+}
```
tylersticka, line 58: Can we also style the `:active` state like we do `c-button`?

derekshirk:
Thanks @tylersticka! I have addressed the two most recent pieces of feedback.

src/components/message/message.scss
```diff
@@ -0,0 +1,65 @@
...
+
+.c-message__content {
+  padding: ms.step(1);
+}
+
+/**
+ * 1. To ensure desired appearance, the button background-color should inherit
+ *    the same color as its parent.
+ * 2. To ensure desired appearance, the close icon should be the same color as
+ *    the parent text color.
```
tylersticka, line 29: I'm sorry to harp on this, but "desired appearance" doesn't really say much in terms of "why."

I can assume from `background-color: inherit;` that the `background-color` should `inherit` from the parent, and that you wouldn't have set that style if it was not the "desired appearance." I still don't know _why_.

You have a couple of next steps I would find acceptable:

1. You add the real reason why. "Transparent background and inherited text color promote harmony with adjacent text and won't compete with any adjacent calls to action." Just as an example. Choose this if you have a real reason "why" you can articulate. (This would also be acceptable to include in the story rather than the CSS if you feel it helpful for more than designer-developers.)
1. Remove the comment. If you don't have anything to say beyond a re-stating of the rule and property, that's fine!

derekshirk: No need to apologize! I appreciate the feedback.

derekshirk: > Just as an example

I chose to copy your suggestion verbatim. 

- It achieves the intended goal completely and concisely. 
- I wasn't convinced that re-wording what you wrote was worth it, when the example feels appropriate to me
- I learned the lesson here 😄

derekshirk:
closes #510

Issue #493, opened 2020-02-21
Card pattern


dromo77:
I'm calling this a card, but there may be a better name for it. This is a common grouping of elements seen throughout the site. They all look a little different, but they share similar elements: 

- An image
- Heading
- Link
- Descriptive text

<img width="1256" alt="cards" src="https://user-images.githubusercontent.com/42841342/75061099-dbef2f80-5494-11ea-8be4-81e095c4c2e5.png">

tylersticka:
We need to take extra care with this pattern to make sure links are [implemented in an accessible way](https://adrianroselli.com/2020/02/block-links-cards-clickable-regions-etc.html) and that heading elements come first (before images).

Related: #474

tylersticka:
@derekshirk I saw this move to "In Progress." I think that's probably premature in advance of #609 (which did not exist till a minute ago).

derekshirk:
Sounds good! Thank you @tylersticka.

tylersticka:
Another point of reference when it comes to implementing this in an accessible but usable way (not recommending this solution, just food for thought): https://css-tricks.com/block-links-the-search-for-a-perfect-solution/

tylersticka:
Completed as of #1044

Issue #477, opened 2020-02-20
Grid/layout patterns


dromo77:
Create a grid system that can be used to layout content.

<img width="636" alt="grids" src="https://user-images.githubusercontent.com/42841342/74988623-1d2f0300-53f3-11ea-8f7d-1f1fd05db3ca.png">

tylersticka:
I don't know if we _should_ have one big "Grid" pattern anymore!

Maybe in a world with CSS Grid Layout, we can have different layout patterns for different use cases?

Perhaps we can _start_ with our article listing, which I think would benefit the most from CSS Grid! What cool things can we do to surface illustrations where available, to support articles of various titles, etc.?

tylersticka:
We've decided to add layout object patterns as needed for as long as possible rather than creating a thin class-based expression of CSS Grid Layout to satisfy our old habits.

PR #606, opened 2020-04-09
Article prototype example

tylersticka:
## Overview

Introduces a new prototypes section including a single Article prototype example. This example is meant as a starting point for others to "save as" and change, it isn't a proposed design.

The feed data was harvested from our RSS feed but with changes applied for prototyping based on JSON feed format.

## Screenshots

Screenshot_2020-04-09 Prototypes Articles - Example ⋅ Storybook

---

- See #477 #493


src/prototypes/articles/example/example.twig
```diff
@@ -0,0 +1,42 @@
...
+    {% for article in feed.items %}
+      <article>
+        <header>
+          <h2>
+            <a href="{{article.url}}">
+              {{article.title}}
+            </a>
+          </h2>
+        </header>
+        <p>
```
derekshirk, line 12: 🤔 Would it be appropriate to wrap content after `<header>` and before `<footer>` in a `<section>` tag?

tylersticka: No, `<section>` would not be appropriate there.

derekshirk: Gotcha, I saw [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article#Examples) example and wasn't sure.

tylersticka: That example confuses me.

The good news is that this is just an example: For prototyping purposes, you can "save as" and write whatever markup you want! 🎉

derekshirk: 😄 Thank you sir.

Issue #578, opened 2020-03-31
Improve sorting of Storybook sidebar


tylersticka:
Currently the sidebar content is sorted with "Getting Started" first (great!), followed by Components, Design Tokens, Design and Themes.

Ideally, I'd love to see these groups sorted in the following order (also including things we haven't created yet):

1. Getting Started
1. Design
1. Design Tokens
1. Objects
1. Components
1. Utilities
1. Themes
1. Third-Party
1. Prototypes

Sorting logic should _only_ apply to the top level.

Josh68:
Hopefully, this is just some fancy sorting foo in `preview.js`, so adding a size 1. Will update if it turns out to be harder.

PR #607, opened 2020-04-09
Update menu sort

Josh68:
## Overview

Uses a predefined list of categories to update the Storybook top-level menu order.

As new categories are added, the ones already defined here should be correctly sorted, and new ones can be added, as needed.

Note: https://github.com/storybookjs/storybook/pull/9188 could supplant this update if we were to upgrade to Storybook v6 in the future.

## Screenshots

Existing menu:

image

Updated menu order:

image

## Testing

1. [Preview theme](https://deploy-preview-607--cloudfour-patterns.netlify.com/)
1. Confirm the order of the menu matches the list, as specified, in issue [578](https://github.com/cloudfour/cloudfour.com-patterns/issues/578)

---

- Fixes # [578](https://github.com/cloudfour/cloudfour.com-patterns/issues/578)

/CC @tylersticka


.storybook/preview.js
```diff
@@ -9,13 +10,65 @@ import 'focus-visible';
...
+  'Getting Started',
+  'Design',
+  'Design Tokens',
+  'Objects',
+  'Components',
+  'Utilities',
+  'Themes',
+  'Third-Party',
+];
+/**
```
Josh68, line 46: This code could be simplified by getting rid of the `sanitize` functionality, but I thought it might be a good idea. Up for review.

.storybook/preview.js
```diff
@@ -9,13 +10,65 @@ import 'focus-visible';
 import '../src/index.scss';
 import './preview.scss';
 
+/**
+ * @typedef { [String, import('@storybook/client-api/dist/types').StoreItem] } StoryItem
```
Josh68, line 14: I could not find a type for what is actually passed to `addParameters.options.storySort`, but this is its shape, and the `StoreItem` is the value that's used in the sorting function.

.storybook/preview.js
```diff
@@ -1,4 +1,5 @@
 import { addDecorator, addParameters } from '@storybook/html';
+import { sanitize } from '@storybook/csf';
```
tylersticka, line 2: Shouldn't this dependency be added to `package.json`?

.storybook/preview.js
tylersticka: Forgive my ignorance about `@typedef`, but it's unclear to me what this is conveying, or why it's higher up in the file. Is there a link to a source file or something we could include closer to the actual logic instead?

.storybook/preview.js
```diff
@@ -9,13 +10,66 @@ import 'focus-visible';
...
+  'Utilities',
+  'Themes',
+  'Third-Party',
+  'Prototypes',
+];
+/**
+ * Sanitize (slufigy) each entry in the list of categories, to normalize odd characters,
+ * capitalization, and use of spaces vs dashes. Do this outside of the `storySort`
+ * sorting function
+ */
```
tylersticka, line 51: This comment focuses on _what_ you're doing, but doesn't include _why_. What do odd characters, capitalization, spaces, etc. impact in terms of sorting?

.storybook/preview.js
Josh68: It appears it was included by default with Storybook, but I can try to make it a separate dependency.

tylersticka: Because it's a namespaced dependency, it would be good to track it in `package.json` just in case Storybook ever moves away from a monorepo strategy (and for consistency).

.storybook/preview.js
Josh68: @tylersticka I can move this, and I can add a comment if that would help. 

We've been using JSDoc in the Decathlon project to add inspectable and verifiable typing information. 

In this case, having this `@typedef`, which is essentially an interface definition, means that you can inspect the code further down and see what types of values are actually expected when I have mystery code like this `story[1].kind.split('/')[0]`. Hovering over this reveals that the `kind` property is a `string`, as expected, on the `StoreItem` type that's the 2nd member of each `StoryItem` passed into `storySort` (that's a mouthful).

.storybook/preview.js
Josh68: Added

.storybook/preview.js
Josh68: Moved the `@typedef` and added comments.

.storybook/preview.js
Josh68: I've updated the comment here and added one inside the `storySort` function, in an attempt to explain _why_ I did this. Whether this is "needed" is unclear to me. I just felt a little uneasy about having hard-coded strings that might deviate from the strings in `Meta` `title` attributes in simple ways, but then again, those are also just entered by hand. Makes me wish these were all tokenized, somehow.

If you feel this is excessive/unnecessary, I'll just pull it out and simplify.

Josh68:
@tylersticka, when you have a moment, I've updated according to your feedback. Thanks

.storybook/preview.js
tylersticka: > Sanitize (slufigy) each entry in the list of categories, to normalize odd characters, capitalization, and use of spaces vs dashes.

You've added more detail about _what_ you are doing, but that wasn't the issue. The issue is _why_ you are doing it. _Why_ are we normalizing these things? What happens if we don't?

I think you're strategy of slug-ifying is fine. What I can't tell is whether or not it's even necessary, because I don't know what prompted it. Was the sorting wrong beforehand? Are there certain characters it is incompatible with?

If the answer is "I just had a feeling this _might_ be necessary someday," that probably doesn't justify the inclusion of this logic here. But if you had a non-hypothetical reason, like "spaces weren't sorted the way we would expect them to be" or something like that, it's critical we include that detail in our comments.

Josh68: > If the answer is "I just had a feeling this might be necessary someday," that probably doesn't justify the inclusion of this logic here

This is the case. I'm going to remove the sanitization and streamline. If needed in the future, we can re-include.

Josh68:
This is ready for re-review, with the `sanitize` dependency removed

.storybook/preview.js
```diff
@@ -16,6 +16,58 @@ addDecorator(withA11y);
 const themes = [{ name: 'Dark', class: 't-dark', color: colors.primaryBrand }];
 addParameters({ themes });
 
+/**
+ * Define the type of value passed to `addParameters.options.storySort`, which is
+ * parsed to get to the name of each story's category
+ * @typedef { [String, import('@storybook/client-api/dist/types').StoreItem] } StoryItem
+ */
```
tylersticka, line 23: Can you please include a link to where you found this information?

Josh68: TBH, I searched for the shape that matched what I was logging when testing how to get the value I needed for sorting. This led me to the `StoreItem` interface: https://github.com/search?q=StoreItem+in:file+extension:ts+repo:storybookjs/storybook&type=Code

To get the type back, I needed to reference the `/dist` rather than the `/src`.

I realize this project is not set up to use TypeScript by default, as Decathlon is. If this is too much magic for you, I can revert the use of the typings altogether.

This is the reference that shows the way to get at the values for an example sort: https://storybook.js.org/docs/configurations/options-parameter/#sorting-stories

Josh68: For context, this code example just made me think "Can I get some explanation of this magic?":

```js
storySort: (a, b) =>
      a[1].kind === b[1].kind
...
```

tylersticka: Yeah, if this is a typescript definition that this information isn't relevant to this project.

Josh68: Then I'll revert this, as well. Understand that you can use TypeScript with JSDoc and plain JS files, but consuming type definitions. This is what we ended up doing elsewhere, but also had things set up to verify typings.

calebeby: FWIW I think these jsdoc annotations are useful, they are being used to power vscode's autocomplete and for me they are useful to help understand the code

tylersticka: Thanks, @calebeby.

Trying to re-phrase so I'm clear:

It's fine to include `@typedef`, but for those of us unfamiliar with TypeScript it would help a lot to have some sort of `@see` link to a relevant URL, even if it is just the source file on GitHub where the type is defined.

calebeby: If you ctrl-click on `StoreItem` on that line in VSCode it will go to where it is defined in `node_modules`

Josh68: OK, I hope this update isn't going to spawn much more discussion. I went down the road of adding typings because I wanted to use JSDoc annotations. If that's still preferred, I can define a type and just describe the shape of the value used for sorting (`object` with a `kind`:`string` property). As it is, I ended up just describing what's going on and removed the JSDoc annotations. Added a reference to the `storySort` example in Storybook docs.

tylersticka: My dudes: You're killing me here. As the person primarily responsible for maintaining this file, I'm asking for context. Not explanations of what `@typedef` is. Not instructions on how to use my editor. Concise, clear context for why the code is written.

Ideally, this would be in the comment itself. For example:

```
/**
 * The parameters for Storybook sorting functions are sparsely documented.
 * Including a detailed type definition in lieu of clear documentation to
 * link to.
 *
 * @typedef (...)
 */
```

For some reason, this feedback is not being addressed. So I've asked for a URL for context instead. Now I'm hearing that this isn't necessary because some of the team is aware of editor functions in a specific editor.

How are we still spinning on this?

Josh68: > If you ctrl-click on StoreItem on that line in VSCode it will go to where it is defined in node_modules

I took @tylersticka's question to mean "How did you find this," and the truth was, I did some digging. I may not have known the appropriate way to easily get to the interface definition, or it may be that something's not quite right with Storybook's typings, not sure.

I fear I may have gone too far in the other direction with my last update. I can revert that commit and revise as desired, if that would be better.

Sorry for the churn. I'm trying to provide decent documentation and not add confusion.

Josh68: > I'm sorry for sounding so exasperated about this, but I'm at a loss for why we're still spinning on such a small request.

I'm sorry, as well. If adding the comment you outlined above to the code as I had it in https://github.com/cloudfour/cloudfour.com-patterns/pull/607/commits/c3fa4d9783e501747975bc980c0a11d852267023, I will do that.

Josh68: @tylersticka, I have reverted back to the previous state, added the description on the `@typedef` you suggested, and added links to the GitHub source files for the interfaces that make up `StoreItem`. I hope this provides what you're looking for.

Josh68:
Thanks @tylersticka and @calebeby.

PR #623, opened 2020-04-13
Rounded corner utilities

### LINKED ISSUE #494: Thumbnail patterns

tylersticka:
## Overview

Adds `u-round`, `u-round-full` and `u-round-none` utilities.

Updates `o-embed` to work with these utilities.

This gives us what we need to make circular elements like avatars. See included docs for more info.

Name change inspired by [similarly named Tailwind utilities](https://tailwindcss.com/docs/border-radius/), but feedback on that is welcome.

I originally named `u-round-none` as `u-round-initial` (and it had a value of `initial` rather than `0`), but after writing the docs I decided that was not as immediately intuitive. Happy to discuss that further if others disagree.

## Screenshots

Screen Shot 2020-04-13 at 2 18 49 PM

### Objects/Embed

Screen Shot 2020-04-13 at 2 50 30 PM

---

- See #494


src/design-tokens/sizes.yml
```diff
@@ -1,49 +1,53 @@
...
     type: relative
     category: sizing
-  radius_medium:
+  - name: radius_medium
     value: 0.25em
     type: relative
     category: radius
-  control_height:
+  - name: radius_full
+    value: 9999px
```
derekshirk, line 19: For my own knowledge gathering, can you comment on why `9999px` was preferred here, instead of a high percentage based value?

tylersticka:
> My only hang-up is applying `u-round` doesn't result in a "round" object.
> Did you consider `u-rounded`, `u-rounded-full` and `u-rounded-none`?

I'll walk you through the naming journey I went on:

`u-border-radius`
Hmm, that's kind of a mouthful.

`u-radius`
That isn't a very friendly word, is it? Maybe I should make it more about _what it does_ than _how it works_.

`u-rounded-corners`
Are corners necessary? What else can CSS round?

Hmm, what do other frameworks call this…

(Looks up Tailwind.)

`u-rounded`
Oh, cool. "Rounded." That's simple.

It's a little funny to have it be past-tense. Are all our utilities going to be past-tense? `u-spaced`, `u-padded`, `u-text-colored-red`?

Then again, I never say "that button has round corners." But I _do_ say "maybe we should round the corners." It works more as a verb than a noun, which sort of makes sense for a class.

Maybe rounded has more meaning than I think. I should look up the definition...

> Shaped into the form of a circle or sphere; made round.

Guess not! `u-round` it is.

That said, something _making sense_ is not the same thing as something _being easy to understand_. I'd love to hear more opinions.

src/design-tokens/sizes.yml
tylersticka: Great question, I will add a `comment` value (and a new issue to _display_ comments when they exist in our design token docs)

Paul-Hebert:
Chiming in on the naming conversation:

Rounded is clearer to me. When I hear something is "round" I assume it's circular. When I hear "rounded" I assume the corners are rounded.

Josh68:
> Chiming in on the naming conversation:
> 
> Rounded is clearer to me. When I hear something is "round" I assume it's circular. When I hear "rounded" I assume the corners are rounded.

FWIW, I agree with @Paul-Hebert

tylersticka:
All right, it's called `u-rounded` now!

But I better not see classes like `u-padded` show up!

src/design-tokens/sizes.yml
tylersticka: ✅ Comment added

src/design-tokens/sizes.yml
```diff
@@ -1,49 +1,57 @@
 props:
-  edge_small:
+  - name: edge_small
```
Josh68, line 2: I just want to ask a naive question. Is the `name` key just an optional organizing principle for Theo YAML files (meaning the consumer output is the same with or without)? I saw this and immediately thought "This must not break anything, right?" Thanks

tylersticka: Theo can be an array or an object.

If it is an object, the syntax is more terse but the order of tokens is not preserved.

If it is an array, the order is preserved but you need to use `name` for the variable.

More info: https://github.com/salesforce-ux/theo#spec

Josh68: Ah, I think I saw that in a JSON example, but didn't realize this was the YAML equivalent. Makes sense. Thanks.

Josh68: I missed the leading `-`s, which turned the properties from an object into an array.

Josh68:
> All right, it's called `u-rounded` now!
> 
> But I better not see needlessly past-tense classes like `u-padded` show up!
> 
> 

`u-failed`

PR #625, opened 2020-04-14
Remove Private Flag

spaceninja:
## Overview

Remove the `private` flag from `package.json` so we can publish this package to the npm registry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "cloudfour.com-patterns",
+  "name": "@cloudfour/patterns",
   "version": "0.1.0",
   "author": "Cloud Four",
   "description": "Front-end patterns for cloudfour.com",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,6 @@
 {
   "name": "cloudfour.com-patterns",
   "version": "0.1.0",
-  "private": true,
   "author": "Cloud Four",
   "description": "Front-end patterns for cloudfour.com",
   "homepage": "https://github.com/cloudfour/cloudfour.com-patterns",


package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "cloudfour.com-patterns",
```
tylersticka, line 2: Note: Before we publish, should we namespace this for our org? `@cloudfour/patterns` or something like that?

Just a thought, not a blocker for this PR.

spaceninja: I thought about that — I could go either way. I didn't change it because I thought `@cloudfour/patterns` (which just says to me "patterns owned by the cloudfour org") was less specific than `cloudfour.com-patterns` (which says to me "patterns specifically for cloudfour.com, and implies ownership by the cloudfour org).

That said, I'm not feeling particularly strongly about this, so I'm happy to change it. Anyone have strong feelings?

gerardo-rodriguez: What about `@cloudfour/cloudfour.com-patterns`. 

That, in theory, would allow to have others in the future that follow the same pattern:
- `@cloudfour/site.com-patterns`
- `@cloudfour/other-site.com-patterns`

Too much? 🤔¯\\\_(ツ)_/¯

spaceninja: I like that you're thinking towards the future — I feels repetitive to me on the face, but thinking about adding other sites using that naming pattern it makes sense. I'm down for that suggestion. @tylersticka?

gerardo-rodriguez: Or, a different twist... 
- `@cloudfour-patterns/cloudfour.com`
- `@cloudfour-patterns/site.com`
- `@cloudfour-patterns/other-site.com`

spaceninja: @tylersticka settled this on Slack when he pointed out that these patterns are mostly intended for cloudfour.com but may also be used for PWA Stats, our handbook, etc.

Changing the package name to `@cloudfour/patterns`

spaceninja:
@tylersticka would you mind re-approving?

PR #629, opened 2020-04-14
Show token comments and font families in docs

tylersticka:
## Overview

Makes several improvements to our Theo design token translation task:

- If any token in a category includes comments, they will be displayed in a new "comment" column.
- If a token is in category `font-family`, the value will have each step of the stack applied for preview purposes.
- All token names and non-font-family token values are wrapped in `<code>` so they are more consistent line to line.

## Screenshots

Screen Shot 2020-04-14 at 4 33 48 PM

Screen Shot 2020-04-14 at 4 33 59 PM

### DIFF
--- a/.theo/formats/stories.mdx.js
+++ b/.theo/formats/stories.mdx.js
@@ -40,10 +40,30 @@ function mdxColors(props) {
  * @returns {String}
  */
 function mdxProp(prop, includeComment) {
-  let content = `
-    <td style={{width:'10%'}}><code>$${kebabCase(prop.name)}</code></td>
-    <td>${prop.value}</td>
-  `;
+  // First cell has small percentage to give values and comments more room.
+  let content = `<td style={{width:'10%'}}><code>$${kebabCase(
+    prop.name
+  )}</code></td>`;
+
+  if (prop.category === 'font-family') {
+    // Create an array for each of the fonts in the family string
+    let fonts = prop.value.split(',');
+    // Start a new array to hold the preview elements we're about to create
+    const fontStrings = [];
+    // For each font string we create, we are going to remove the first font
+    // element using `shift`. This allows each segment to be previewed along
+    // with its remaining fallbacks.
+    while (fonts.length) {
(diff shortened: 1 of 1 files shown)


.theo/formats/stories.mdx.js
```diff
@@ -54,12 +79,15 @@ function mdxProp(prop) {
  * @returns {String}
  */
 function mdxProps(props) {
-  const rows = props.map(mdxProp);
+  const commentedProps = props.filter((prop) => !!prop.comment);
```
Josh68, line 82: Nit. I doubt this matters, but `'comment' in props` would return `true` for props with an empty comment, while this code wouldn't. But what you have is probably the behavior you want.

tylersticka: Yeah, I don't want it to return true for an empty comment.

Is this blocking approval?

.theo/formats/stories.mdx.js
```diff
@@ -36,15 +36,40 @@ function mdxColors(props) {
...
+    content += `<td>${fontStrings.join(', ')}</td>`;
+  } else {
+    content += `<td><code>${prop.value}</code></td>`;
+  }
+
+  if (includeComment) {
+    content += `<td>${prop.comment || '&nbsp;'}</td>`;
+  }
+
+  return `<tr>${content.trim()}</tr>`;
```
Josh68, line 72: Not sure, but is there any need to add `.trim()` back after the `return`ed `<tr>`-wrapped string?

.theo/formats/stories.mdx.js
Josh68: No, approved. Just trying to be thorough.

spaceninja: Can confirm this is working as intended — to filter the props array and return only props with comments.

.theo/formats/stories.mdx.js
Josh68: On second look, there is no need for that. Resolving this comment

PR #644, opened 2020-04-23
Rhythm object

tylersticka:
## Overview

- Overrides default `margin` for more elements.
- Adds an `o-rhythm` object class for adding whitespace vertically between adjacent child elements.
- Adds modifiers based on common use cases to promote consistency.

### Thought Process

I decided to abandon our previous utility-based approach to this after reviewing our live site and making a few observations:

- The previous utilities included 42 separate classes. Of these, only ten were ever used.
- Of the ten classes we used, four were the same step (`u-staggerItems1` compared to `u-spaceItems1`) and made little to no visible difference if swapped for the opposite utility of the same step.
- Of those six remaining classes, three were used only once or twice. These outliers all seem to be within a step of more common values, which suggests to me that having so much choice was resulting in unintentional inconsistencies.

I chose the name `rhythm` because I think it just _feels_ like what the pattern is trying to accomplish. Every time I'm outvoted on this, I find myself missing it… so I decided to suggest it again in hopes that I'll gradually wear everyone else down. 😉 

To make sure I was covering the common use cases, I made a spreadsheet tracking which modular scale steps were used most often:

Scale step | Occurrences in templates | Usage Notes
--- | --- | ---
-6 | 4 | Lists and tiny content chunks
-2 | 1 | Our Approach asides
-1 | 9 | Intros and medium content chunks
1 | 30 | Article content, page content, etc.
3 | 1 | Article headings
4 | 2 | Sections

I translated the intent of these to the following modifier classes:

Scale step | Modifier  | Intent
--- | --- | ---
-6 | `compact` | Lists and tiny content chunks
-1 | `condensed` | Intros and summary content
1 | (default) | Article content, page content, etc.
3 | `generous`<br>`generous-headings` | Sections and article headings

## Screenshots

Screenshot_2020-04-22 Storybook


src/base/_defaults.scss
```diff
@@ -33,6 +33,27 @@ a {
...
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+ol,
+p,
+pre,
+ul {
```
Paul-Hebert, line 53: Is there a reason we want to override these but not other [block level elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements)?

tylersticka: I pulled this list from our existing patterns, which we had pulled from Suit CSS, but I see no reason not to extend this to other elements!

tylersticka: (Assuming those other elements _have_ default margin, of course)

tylersticka:
> If you could look into that it would be great, but it seems like it might be a weird fluke so I'm approving.

It's probably because of the YouTube embed in the article, but I can take a look.

Paul-Hebert:
>It's probably because of the YouTube embed in the article, but I can take a look. 

Ahh, makes sense!

src/base/_defaults.scss
tylersticka: @Paul-Hebert Thanks for this, I made a few changes:

1. I consolidated all `margin` overrides from this file to the same place.
1. I expanded the list of elements based on browser default stylesheets, excluding elements I don't anticipate we'd ever use. (See comment.)
1. I removed `margin: 0` declarations from components that are redundant. This also makes the `o-rhythm` layout class easier to use since it won't have the whitespace overruled immediately by a nested `c-button` or `c-input`.

tylersticka:
> If you could look into that it would be great, but it seems like it might be a weird fluke so I'm approving.

@Paul-Hebert This error should be gone in the latest version. There is a warning about including aXe Core in the iframe to test its content, but since that content is served from YouTube, I don't think we need to worry about that.

tylersticka:
Tests are failing due to a GitHub Actions outage: https://www.githubstatus.com/incidents/zdxk6xq21405

Will trigger a re-run later today.

tylersticka:
I have no idea why the one lint test is still failing. GitHub Status says Actions should be fine now. All the others passed. It isn't failing on linting, it's failing on `npm ci`. I try running that command locally and it works fine.

Issue #646, opened 2020-04-23
Polyfill CSS for targeted browsers


tylersticka:
Originally our intention was to leave polyfilling of CSS to the projects themselves, but our recent decision to adopt [logical properties and values](https://preset-env.cssdb.org/features#logical-properties-and-values) demands _some_ sort of polyfilling for the latest version of Safari. Without that, it'll be pretty impossible for us to test.

We already have PostCSS added to our stack, so the most obvious solution would be to add the [preset env](https://preset-env.cssdb.org/) plugin and a [browserslist](https://github.com/browserslist/browserslist). We do not have to determine our final shipping list of browsers, but ideally we'd use a minimal configuration for the time being that covers the latest version of modern browsers with significant market share.

PR #651, opened 2020-04-24
Spacing utilities

tylersticka:
## Overview

This reintroduces spacing utilities to the project with the following changes from the production version:

- New class naming convention.
- Retains step `0` instead of omitting it.
- Does not reintroduce `space-items` (see #644).
- Uses logical property names instead of physical ones.
- Uses relative units (`em`).

This PR does not include responsive variations of these utilities because it was getting a little large as-is.

## Screenshots

Screenshot_2020-04-24 Storybook

## Testing

Logical property support varies between browsers. This should be resolved by #646. For now, it's best to test in Firefox.

### DIFF
--- a/src/utilities/spacing/demo/directions.twig
+++ b/src/utilities/spacing/demo/directions.twig
@@ -1,0 +1,22 @@
+{% if step < 0 %}
+  {% set step = 'n' ~ step|abs %}
+{% endif %}
+
+{% embed '../../../objects/rhythm/rhythm.twig' with {
+  class: 'demo-u-' ~ name ~ '-directions'
+} %}
+  {% block content %}
+    {% for direction in ['', '-block', '-block-start', '-block-end', '-inline', '-inline-start', '-inline-end'] %}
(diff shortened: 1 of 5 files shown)


src/utilities/spacing/demo/directions.twig
```diff
@@ -0,0 +1,22 @@
+{% if step < 0 %}
+  {% set step = 'n' ~ step|abs %}
+{% endif %}
+
+{% embed '../../../objects/rhythm/rhythm.twig' with {
+  class: 'demo-u-' ~ name ~ '-directions'
```
Josh68, line 6: Nit: When we were doing similar includes in Liquid, we somehow documented the variables that were passed into the include, in a form that looked a bit like JSDoc. It would just be nice to know what `name` is / can be, in this case, and that it's a required "param" for the include.

tylersticka: This seems like overkill for templates created for demonstration purposes.

Josh68: Fair enough.

PR #652, opened 2020-04-24
CSS polyfills

### LINKED ISSUE #646: Polyfill CSS for targeted browsers

Josh68:
## Overview

Adds configuration for postcss-preset-env, initial .browserlistrc, and example SCSS update (to be reverted before merge)

## Screenshots

In Chrome, latest

**Output**
image

**Source (mapped)**
image

## Testing

1. [Preview theme](https://deploy-preview-652--cloudfour-patterns.netlify.app/)
1. Go to Utilities -> Spacing -> Padding (e.g.)
1. Inspect a `u-pad-block` or `u-pad-inline` element and not separate directional padding values
1. Click to go to the source `spacing.scss` and note the original properties are `padding-block` and `padding-inline`

---

- Fixes #646

/CC @tylersticka


src/components/button/button.scss
```diff
@@ -33,7 +33,12 @@
   font-weight: font.$weight-semibold;
   height: sizes.$control-height; /* 3 */
   line-height: normal;
-  padding: sizes.$control-pad-vertical sizes.$control-pad-horizontal;
+  /* @TODO - restore original, below */
```
Josh68, line 36: This file should be completely restored before approval and merge.

.browserlistrc
```diff
@@ -0,0 +1,4 @@
+# Browsers that we support
```
Josh68, line 1: Use of a `.browserlistrc` is aligned with setup in other projects. "Recommended" approach (not sure why) is to add to `package.json`, but I prefer the separation. However anyone prefers, I can move this.

Also, these are only initial tests. Can be updated here and later.

.storybook/main.js
```diff
@@ -60,6 +61,17 @@ module.exports = {
...
+              // @see https://github.com/csstools/postcss-preset-env/blob/master/INSTALL.md#webpack
+              ident: 'postcss',
+              plugins: () => [
+                /**
+                 * Enables Stage 2 options, by default.
+                 * This includes logical properties and values.
+                 * @see https://preset-env.cssdb.org/features#logical-properties-and-values
+                 * @see https://github.com/csstools/postcss-preset-env#stage
+                 */
+                postcssPresetEnv(),
```
Josh68, line 73: We could, alternatively, spec stage 3 or 4 and add `logical-properties-and-values` support separately.

.browserlistrc
tylersticka: I know `browserlist` is easier to say, but is there a reason you went with that instead of `browserslist` (which matches the name of the config)?

Josh68: Confusion I guess I share with some others: https://github.com/browserslist/browserslist/issues/138

Fixed the name

.storybook/main.js
tylersticka: @Josh68 Any reason we wouldn't add this to our `postcss.config.js` instead?

src/components/button/button.scss
tylersticka: Now that #651 has been merged, you should have some properties to test without needing to introduce this code.

.browserlistrc
Josh68: This is a logical OR, and results in the following list (`npx browserslist`)

```
and_chr 79
and_ff 68
and_qq 1.2
and_uc 12.12
android 76
baidu 7.12
bb 10
chrome 80
chrome 79
edge 80
firefox 72
ie 11
ie_mob 11
ios_saf 13.3
ios_saf 13.2
kaios 2.5
op_mini all
op_mob 46
opera 66
safari 13
samsung 10.1
```

src/components/button/button.scss
Josh68: Right you are. I'll revert that file now.

.storybook/main.js
Josh68: I hadn't even seen that was in our file structure. Will try that out now.

Josh68:
@tylersticka, all updated with instructions to test with new spacing utils.

.storybook/main.js
Josh68: Updated

.browserslistrc
```diff
@@ -0,0 +1,4 @@
+# Browsers that we support
+
+last 1 version
+> 3%
```
calebeby, line 4: Hey Josh, I know this just got merged, but I think this browserslist might support more browsers than we want to support in this repo.

> covers the latest version of modern browsers with significant market share.

[`last 1 version`](https://browsersl.ist/?q=last+1+version) includes outdated browsers like IE11 and IE Mobile and Blackberry Browser. This will result in more transpilation than I think is necessary.

Can we limit this to evergreen browsers? If we decide that our main cloudfour site needs to support more browsers than this repo supports, we can add more transpilation there.

How about:

```
>2%
last 1 safari version
last 1 ios version
last 1 firefox version
last 1 chromeandroid version
```

[Preview](https://browsersl.ist/?q=%3E2%25%2C+last+1+safari+version%2C+last+1+ios+version%2C+last+1+firefox+version%2C+last+1+chromeandroid+version)

cc @tylersticka

calebeby: I'm willing to submit a PR if that is helpful

tylersticka: @calebeby Go for it! Or create an issue if time is a factor, either way.

calebeby: ✅ https://github.com/cloudfour/cloudfour.com-patterns/pull/658

PR #662, opened 2020-04-27
Fix spacing utility demos

### LINKED ISSUE #661: *-inline-start and *-inline-end utilities not functioning
Not sure what happened, but some of the spacing utilities are no longer functioning as expected.

Examples circled in green are working correctly, in red are not:

Screen Shot 2020-04-27 at 1 13 25 PM

Possibly related to recent polyfill additions?

---

/CC @calebeby @Josh68

Josh68:
## Overview

- Reuses postcss plugin config for both Webpack and Gulp tasks (not necessary for this update, but these were out of sync)
- Removes [postcss-preset-env](https://github.com/csstools/postcss-preset-env) plugin
- Updates spacing util demos to replace shorthand examples with `-start` and `-end` combinations, to fix issues across supported browsers. 
    - `postcss-preset-env` with [postcss-logical](https://github.com/csstools/postcss-logical) was unable to solve these issues and introduced others

## Screenshots

Utilities -> Spacing -> Padding demo, PR 651:

image

Same demo, this PR (note `.u-pad-block-1` and `.u-pad-inline-1`):

image

## Testing

1. [Preview theme utils-spacing-padding preview](https://deploy-preview-662--cloudfour-patterns.netlify.app/?path=/story/utilities-spacing--padding)
1. [PR 651 utils-spacing-padding preview](https://deploy-preview-651--cloudfour-patterns.netlify.app/?path=/story/utilities-spacing--padding) - From before the introduction of `postcss-preset-env`
1. [Preview theme utils-spacing-margin preview](https://deploy-preview-662--cloudfour-patterns.netlify.app/?path=/story/utilities-spacing--margin)
1. [PR 651 utils-spacing-margin preview](https://deploy-preview-651--cloudfour-patterns.netlify.app/?path=/story/utilities-spacing--margin)
1. See that shorthand examples (`.u-pad-inline-1`, `.u-pad-block-1`, `.u-space-pad-1`, and `.u-space-block-1`) all look correct in the previews for this PR
1. Confirm in Chrome, FF, and Safari / iOS Safari (+ new Edge, if you like)

---

- Fixes #661

/CC @person


gulpfile.js/tasks/build.js
```diff
@@ -28,7 +29,7 @@ const buildSass = () => {
     )
     .pipe(rename({ basename: 'standalone' }))
     .pipe(dest(outDir))
-    .pipe(postcss([cssnano()]))
+    .pipe(postcss([cssnano(), postcssPresetEnv()]))
```
calebeby, line 32: @Josh68 Can we make it read from our postcss.config.js so we have one shared postcss config for both storybook/webpack and gulp?

Josh68: Yeah, I was thinking that. It's not ready (hence, still in draft). I need to condition `cssnano`, I think.

Josh68: If I add an extra property name/value pair in the mixin, I get output 🤔 

```css
 &-inline-start-#{$name} {
    position: relative;
    @if $property == margin {
      margin-inline-start: $value;
    } @else {
      padding-inline-start: $value;
    }
  }
```

```css
.u-pad-inline-start-1 {
    position: relative;
}
```

Josh68: This is still a WIP with issues remaining in some examples. Using the `preserve` option for the PostCSS logical properties module, some properties are being left as-is and some converted to directional properties. There may be a bug in the module, but I'm not entirely clear on that yet.

Ideally, everything would be converted with fallbacks for non-conforming browsers. There may also be a need to refactor with `dir` attributes set on parent elements for each example. However, the current issues are showing up with logical block properties (which appear to be converted only to directional properties, without preserving the logical versions), and I'm not clear whether `dir` makes a difference in these cases. Adding extra markup throws off existing demo CSS, so I've abandoned that approach for the moment.

.storybook/main.js
```diff
@@ -59,7 +60,9 @@ module.exports = {
           {
             loader: 'postcss-loader',
             options: {
+              ident: 'postcss',
```
Josh68, line 63: Documentation says this `ident` line is needed. If anyone wants a reference, I can add, but thought it might be ok to leave off.

.storybook/main.js
```diff
@@ -59,7 +60,9 @@ module.exports = {
...
               sourceMap: true,
+              plugins: () => postcssConfig.plugins
```
Josh68, line 65: Use plugins from our centralized config.

gulpfile.js/tasks/build.js
```diff
@@ -3,8 +3,7 @@ const { src, dest, parallel } = require('gulp');
 const sass = require('gulp-sass');
 const rename = require('gulp-rename');
 const postcss = require('gulp-postcss');
-const cssnano = require('cssnano');
-
+const postcssConfig = require('../../postcss.config');
```
Josh68, line 6: For using plugins from our centralized postcss config, shared with the webpack setup.

src/utilities/spacing/demo/styles.scss
```diff
@@ -9,7 +9,7 @@
     border-radius: sizes.$radius-medium;
   }
 
-  & > div {
+  & > .demo-example {
```
Josh68, line 12: For handling any case where a wrapping `div` with `dir=` attribute needs to be added.

src/utilities/spacing/spacing.scss
```diff
@@ -17,11 +17,18 @@
     }
   }
 
+  /**
+    * 1. @TODO - Re-evaluate shorthand properties if they become fully supported
```
Josh68, line 21: The resolution here is to abandon shorthand properties for the moment, and substitute with `-start` and `-end` combinations, which appear to work across browsers without trying to use postcss-logical, which never gave output that worked in all cases.

Josh68:
@tylersticka and @calebeby: I am putting this out for review having completely removed `postcss-preset-env`, for the moment. I was unable to make the `logical-properties-and-values` feature work as intended, either with or without the `preserve` option.

If in the future we were to add this back, demo markup and CSS would need to be revised to add `dir` attributes to demo containers, while preserving layout (I went through these changes along the way, and they're in the commit history). Even these additions did not work to fix all the demos, with or without `postcss-preset-env`.

The logical-properties plugin did not seem to be converting all values correctly, particularly for the `Utilities` -> `Spacing` -> `Logical Properties` examples, with varied direction and writing modes. I've tried to raise the issue with the maintainer, so maybe it will be usable sometime in the future, if there are actually bugs that can be fixed. 

TBH, I get pretty confused by some of the combinations, particularly with vertical writing modes, but I think the examples in `Utilities` -> `Spacing` -> `Logical Properties` look correct as they were in #651. This update does not affect those demos, but fixes shorthand property examples (see screenshots and testing) by substituting, e.g., `padding-inline` with `padding-inline-start` + `padding-inline-end` -- within the core mixin. I added comments to this effect and hope this is considered OK for the time being.

If there's a feeling we should be keeping `postcss-preset-env` in the project for other reasons, we'd have to consider setting it to stage 3 or 4, because the default -- stage 2 -- will automatically apply logical-properties transformations and reintroduce issues.

Please let me know if you have any questions about this update. Thanks

tylersticka:
Others are welcome to review, but please don't merge until I've had a chance to. Thank you!

src/utilities/spacing/spacing.scss
tylersticka: I'd like to see this comment revised in a few ways:

- [ ] Remove `@TODO`: There's nothing actionable here until something changes in browsers, so this is just extra noise.
- [ ] Remove the commented out shorthand values, they are also just noise.
- [ ] Revise the comment to be about _why_ we're using longhand values and not about _what_ we should do later. Example: "Ideally we could use the `margin-block` shorthand, but it is not as well supported in browsers."

src/utilities/spacing/spacing.scss
```diff
@@ -41,11 +48,18 @@
...
```
tylersticka, line 52: Same feedback as the other comment:

- [ ] Remove `@TODO`: There's nothing actionable here until something changes in browsers, so this is just extra noise.
- [ ] Remove the commented out shorthand values, they are also just noise.
- [ ] Revise the comment to be about _why_ we're using longhand values and not about _what_ we should do later. Example: "Ideally we could use the `margin-inline` shorthand, but it is not as well supported in browsers."

postcss.config.js
```diff
@@ -1,15 +1,14 @@
+const plugins = [
+  require('postcss-inline-svg')({
+    paths: ['./src/assets'],
+  }),
+];
+
+// For production only, add cssnano
+if (/prod/.test(process.env.NODE_ENV)) {
+  plugins.push(require('cssnano')());
+}
```
tylersticka, line 10: Is this change necessary for this PR?

gulpfile.js/tasks/build.js
```diff
@@ -28,7 +27,7 @@ const buildSass = () => {
     )
     .pipe(rename({ basename: 'standalone' }))
     .pipe(dest(outDir))
-    .pipe(postcss([cssnano()]))
+    .pipe(postcss(postcssConfig.plugins))
```
tylersticka, line 30: Is this change necessary for this PR?

gulpfile.js/tasks/build.js
tylersticka: The PostCSS plugin does this automatically. You shouldn't need to pass along a specific config.

postcss.config.js
Josh68: None of the changes having to do with postcss configuration should be necessary for this PR. They were introduced based on initial feedback from @calebeby, before I decided to completely remove `postcss-preset-env`. I can revert all of them.

When I discussed with @calebeby, the feeling was that the gulp task should take plugins. I wasn't clear that this happened automatically (referring to https://github.com/cloudfour/cloudfour.com-patterns/pull/662#discussion_r416905806)

Josh68:
@tylersticka, ready for re-review. Thanks

gulpfile.js/tasks/build.js
Josh68: Reverted. If desired, I can add a card to move `cssnano` into the postcss config, and have it conditional to run only on build. I'm not clear whether this overrides the automatic import of other plugins, the way we currently have it. I believe that's what @calebeby was thinking.

postcss.config.js
```diff
@@ -3,13 +3,5 @@ module.exports = {
...
       paths: ['./src/assets'],
     }),
-    /**
-     * Enables Stage 2 options, by default.
-     * This includes logical properties and values.
-     * @see https://preset-env.cssdb.org/features#logical-properties-and-values
-     * @see https://github.com/csstools/postcss-preset-env#stage
-     * @see https://github.com/csstools/postcss-preset-env/blob/master/INSTALL.md#postcss-cli
-     */
-    require('postcss-preset-env')(),
```
tylersticka, line 13: Should we also `uninstall` this from our `package.json`?

src/utilities/spacing/spacing.scss
```diff
@@ -17,11 +17,17 @@
...
+    */
   &-block-#{$name} {
     @if $property == margin {
-      margin-block: $value;
+      margin-block-end: $value; /* 1 */
+      margin-block-start: $value;
     } @else {
-      padding-block: $value;
+      padding-block-end: $value; /* 1 */
+      padding-block-start: $value;
```
tylersticka, line 30: I don't think the number is necessary at all, since the comment is applying to the whole block of styles.

src/utilities/spacing/spacing.scss
```diff
@@ -41,11 +47,17 @@
...
+    */
   &-inline-#{$name} {
     @if $property == margin {
-      margin-inline: $value;
+      margin-inline-end: $value; /* 1 */
+      margin-inline-start: $value;
     } @else {
-      padding-inline: $value;
+      padding-inline-end: $value; /* 1 */
+      padding-inline-start: $value;
```
tylersticka, line 60: I don't think the number is necessary at all, since the comment is applying to the whole block of styles.

src/utilities/spacing/spacing.scss
```diff
@@ -17,11 +17,17 @@
     }
   }
 
+  /**
+    * 1. Ideally we could use the margin-inline shorthand, but it is not as well
```
tylersticka, line 21: This is `block`, not `inline`. Should probably just be `*-block` shorthand (since this covers `margin` _and_ `padding`).

src/utilities/spacing/spacing.scss
```diff
@@ -41,11 +47,17 @@
...
```
tylersticka, line 51: Should probably just be `*-inline` shorthand (since this covers `margin` _and_ `padding`).

src/utilities/spacing/spacing.scss
Josh68: Sorry, copy-paste error. Fixing now.

Josh68: Both updated.

src/utilities/spacing/spacing.scss
Josh68: https://github.com/cloudfour/cloudfour.com-patterns/pull/662#discussion_r416941077

postcss.config.js
Josh68: Thanks for catching that.

Josh68:
@tylersticka, thanks for the feedback. Updated and `postcss-preset-env` removed.

src/utilities/spacing/spacing.scss
```diff
@@ -17,11 +17,17 @@
     }
   }
 
+  /**
+    * Ideally we could use the *-block shorthand, but it is not as well
+    * supported in browsers.
+    */
```
tylersticka, line 23: Nit: We usually align the _first_ asterisks (not the second), like jsdoc.

src/utilities/spacing/spacing.scss
```diff
@@ -41,11 +47,17 @@
     }
   }
 
+  /**
+    * Ideally we could use the *-inline shorthand, but it is not as well
+    * supported in browsers.
+    */
```
tylersticka, line 53: Nit: We usually align the _first_ asterisks (not the second), like jsdoc.

Josh68: I tried that and the linter changed it. I'll try again.

Josh68: No idea why, but this time I reverted it to the way I had it and the linter didn't complain, so updated.

Josh68: @tylersticka, whenever you have a moment ^

PR #665, opened 2020-04-27
Add u-hidden-visually utility

tylersticka:
## Overview

Adds `u-hidden-visually` utility, which I wanted for a different pattern.

This is mostly the same as [what we have currently](https://cloudfour-patterns.netlify.app/patterns/utilities.html#display). I kept the `hidden` past-tense because I like the consistency with `visibility: hidden`, `aria-hidden`, the `hidden` attribute, etc.

The only differences are:

- New class name based on our latest standards.
- Very slightly updated styles based on @Paul-Hebert's [article on the subject](https://cloudfour.com/thinks/see-no-evil-hidden-content-and-accessibility/#showing-additional-content-for-screen-readers).
- Styles are applied via a mixin, in case any of our component elements would benefit from these same styles.

## Screenshots

Screen Shot 2020-04-27 at 2 55 24 PM

## Testing

1. View deploy preview.
1. Navigate to the Utilities ▸ Display ▸ Hidden Visually story.
1. Confirm that only the icon (not the text) is visible.
1. In accessibility tab, confirm that the "Ensures links have discernible text" test has passed.

### DIFF
--- a/src/mixins/_a11y.scss
+++ b/src/mixins/_a11y.scss
@@ -1,0 +1,46 @@
+@use 'sass:map';
+
+/**
+ * Visible only to screen readers.
+ *
+ * @see https://css-tricks.com/inclusively-hidden/
+ */
+
+$_sr-only-props: (
+  border: 0,
+  clip: rect(0 0 0 0),
+  clip-path: polygon(0 0, 0 0, 0 0),
+  height: 1px,
+  margin: -1px,
+  overflow: hidden,
+  padding: 0,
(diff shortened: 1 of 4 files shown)


tylersticka:
> The code changes look good to me. However, I struggled to get the deploy preview to load in Safari for VoiceOver testing, but I _assume_ that the issue is on my network or machine.

I think so, it seems to work for me:

Screen Shot 2020-04-28 at 9 29 03 AM

PR #668, opened 2020-04-28
Add Dot Leader component

tylersticka:
## Overview

Ports the existing [Dot Leader](https://cloudfour-patterns.netlify.app/patterns/components/dot-leader.html) to our new system, with the following changes:

- Now uses CSS Grid, which eliminates the need for non-breaking space character injections for margin and crazy sizing hacks for cross-browser compatibility.
- Built-in support for being a link (not reliant on utilities).

I feel strongly that the Dot Leader should be carried over to updated layouts, as its more distinctive than alternatives I've seen and feels uniquely academic in a way that feels relevant to our team.

## Screenshots

Screenshot_2020-04-28 Storybook

---

- See #492

PR #676, opened 2020-04-30
List object

### LINKED ISSUE #492: List pattern(s)
Create a list pattern. We use lists in a lot of different ways throughout the site so this may need to be multiple patterns

lists

tylersticka:
## Overview

- Adds `o-list` for unstyled lists. Replaces the former `u-listUnstyled` and the more recent `u-list-none`.
- Adds `o-list--inline`. Replaces the former `u-listInline`.
- ~Adds `o-list--ruled`. Codifies a pattern visible on [our presentation pages](https://cloudfour.com/presents/why-you-should-build-a-progressive-web-app-now/).~
- Adds design tokens for list-related spacing, and implements relevant ones in existing `c-table` styles for consistent spacing between ruled rows.
- Updates `c-dot-leader` (#668) examples to use and reference `o-list`.
- Removes `u-list-none` (#666), which had no remaining use cases.

Although I flip-flopped between this being an object or component, I feel reasonably confident in this approach after applying this pattern to the `c-dot-leader` documentation.

## Screenshots

Screenshot_2020-04-29 Storybook(1)

---

- Resolves #492


tylersticka:
I woke up _less_ sure that this isn't a component. 🤣 

I leave it to my esteemed reviewers. We can always change it later.

Josh68:
> I woke up _less_ sure that this isn't a component. 🤣

+1. I tend to agree with your nighttime thoughts.

src/objects/list/list.stories.mdx
```diff
@@ -0,0 +1,60 @@
...
+];
+const tagNames = ['ul', 'ol'];
+
+<Meta title="Objects/List" decorators={[withKnobs]} />
+
+# List
+
+The `o-list` object presents list items in a straightforward and unopinionated manner. While it can be applied to any container, it was designed specifically for `<ul>` and `<ol>` elements.
+
+**Important:** Remember to include the `role="list"` attribute on `<ul>` elements using this pattern to avoid [an issue with unstyled lists and Safari/VoiceOver](https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html).
```
spaceninja, line 20: Once again, excellent comment that answered a question I was just about to type up!

src/objects/list/list.scss
```diff
@@ -0,0 +1,56 @@
...
+.o-list--inline > * {
+  margin-left: sizes.$list-inline-gap / 2;
+  margin-right: sizes.$list-inline-gap / 2;
+}
+
+/**
+ * Modifier: Ruled
+ */
+
+.o-list--ruled > * + * {
```
spaceninja, line 48: Perhaps a silly question, but do we care about what happens if a user applies both `o-list--inline` and `o-list--ruled`?

src/objects/list/list.scss
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Modifier: Inline
+ *
+ * By default, this uses negative margins on the sides to account for extra
+ * margin between child elements. This allows child elements to reach the edges
+ * of their parent while maintaining consistent gaps between.
+ *
+ * This would be a _lot_ more elegant if it used the Flexbox-compatible `gap`
+ * property. But unfortunately there is no way to test for that without false
+ * positives as of this writing.
```
derekshirk, line 26: Is my understanding correct, that at this time there is no way to wrap a simplified version of this, using `gap` in a `@supports` media query?

src/objects/list/list.scss
tylersticka: I asked myself the same thing. Right now they are completely incompatible.

I considered adding a `:not` selector or even a combined selector that inserts vertical rules between items, but since there isn't a place where we're doing that on the current site, it felt like overkill.

I'm open to suggestions!

src/objects/list/list.twig
```diff
@@ -0,0 +1,14 @@
+{% set tag_name = tag_name|default('ul') %}
+
+{% if tag_name == 'ul' %}
+  {% set role = role|default('list') %}
```
derekshirk, line 4: Not sure if it matters, but I expected to see this pipe surrounded by spaces, like this: 

`{% set role = role | default('list') %}`

src/objects/list/list.scss
spaceninja: I think it's totally find to just add a note on the stories page saying "These two modifiers don't play nice with each other."

We don't need to cover every possible use case, but an acknowledgement that some combinations aren't supported is probably worth doing.

derekshirk: I agree 💯 with what @spaceninja said here. 😄

src/objects/list/list.twig
derekshirk: If I look at `button.twig`, It appears we _don't_ have a standard way for formatting this. 

https://github.com/cloudfour/cloudfour.com-patterns/blob/40f8e86d82037ff9b3cc78c6c5bab70b1d808526/src/components/button/button.twig#L1-L16

Is that correct?

derekshirk: Not a release blocker for this PR!

src/objects/list/list.scss
tylersticka: Yep! See the `@see` URLs for more information.

src/objects/list/list.twig
tylersticka: Either way works, but I've been writing it this way because it's how Twig's docs write it: https://twig.symfony.com/doc/2.x/templates.html

tylersticka:
You know what? After reading and re-reading a bunch of CSS namespace articles, I think I'm going to remove the `ruled` modifier.

When I think about it, that's really the only aspect of this that feels component-like to me. It has cosmetic impact, and it coordinates with the table _component_. 

Looking again at [a page where this is used](https://cloudfour.com/presents/why-you-should-build-a-progressive-web-app-now/), I think I may be prematurely abstracting the ruled list from the presentation of past events (the only place where that pattern occurs currently).

tylersticka:
This comment from @spaceninja also reenforces for me that the ruled modifier is the odd thing out: https://github.com/cloudfour/cloudfour.com-patterns/pull/676#discussion_r418137339

derekshirk:
> Although I flip-flopped between this being an object or component, I feel reasonably confident in this approach after applying this pattern to the c-dot-leader documentation.

👍

src/objects/list/list.scss
```diff
@@ -0,0 +1,41 @@
+@use '../../design-tokens/sizes.yml';
+
+/**
+ * Override browser defaults. Note that this means it's very important to
+ * include `[role="list"]` to prevent disruption of list navigation in Safari
+ * VoiceOver.
```
derekshirk, line 6: > in Safari VoiceOver.

Question: Does this only occur when using Safari + VoiceOver? I read the reference link but it was not clear to me whether this occurs when using Chrome + VoiceOver.

tylersticka: Only Safari.

PR #677, opened 2020-04-30
Add babel and typescript

### LINKED ISSUE #663: Set up babel and typescript
https://github.com/cloudfour/cloudfour.com-patterns/issues/653#issuecomment-619286889

> - Add TS config
> - Add type npm script that will check for type errors
> - Add babel-preset-typescript
> - Add babel to webpack config for storybook
> - Add babel to rollup config for output

calebeby:
## Overview

Closes #663 

Now we have type-checking enabled for .ts and .js files in `src`, excluding `src/prototypes`. Also eslint is now enabled on `.ts` files.

This PR also sets up babel-preset-env so that our code gets transpiled to modern browsers.

## Testing

1. `npm run build`, make sure it runs without errors
2. Look at the `dist` folder. Ensure there is:
   - `index.d.ts` - This file contains typings for this pattern library so that when this pattern is installed from npm, people can get correct autocomplete in vscode and typechecking if they are using it
   - `cloudfour-patterns.mjs`: This file should have the un-typed version of the button function, in the ES modules format.
   - `cloudfour-patterns.js`: This file should have the un-typed version of the button function, in the AMD/CJS format


.eslintrc.json
```diff
@@ -16,5 +16,23 @@
   "rules": {
     "padding-line-between-statements": "off",
     "react/jsx-uses-vars": "error"
-  }
+  },
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx"],
```
calebeby, line 22: Eventually I think we should integrate these eslint settings for typescript into `@cloudfour/eslint-plugin`, but for now having it here is good so that we can work out any issues we run into

package.json
```diff
@@ -17,6 +17,9 @@
     "bem",
     "typography"
   ],
+  "engines": {
+    "node": ">=12.0.0"
```
calebeby, line 21: This was needed because eslint was complaining about `fs.promises` being used and it not being supported until node 11. By specifying the minimum node version here, the rule does not trigger

src/components/button/index.ts
```diff
@@ -0,0 +1,3 @@
+export const button = (a: string) => {
```
calebeby, line 1: Remind me to remove this file after review is done. It is for testing/review/demo purposes

.storybook/main.js
```diff
@@ -92,9 +92,15 @@ module.exports = {
         // Optimize and process SVGs as React elements for use in documentation
         test: /\.svg$/,
         use: '@svgr/webpack',
+      },
+      {
+        test: /\.(ts|tsx)$/,
+        loader: require.resolve('babel-loader'),
```
tylersticka, line 98: Any reason this doesn't use the `use: 'babel-loader'` formatting of the other configurations?

gulpfile.js/tasks/build.js
```diff
@@ -33,12 +36,46 @@ const buildSass = () => {
     .pipe(dest(outDir));
 };
 
+const extensions = ['.js', '.ts', '.tsx'];
+
+/*
```
tylersticka, line 41: Nit: Missing double asterisks?

gulpfile.js/tasks/build.js
```diff
@@ -1,17 +1,20 @@
+const glob = require('tiny-glob');
+
 const { src, dest, parallel } = require('gulp');
 
 const sass = require('gulp-sass');
 const rename = require('gulp-rename');
 const postcss = require('gulp-postcss');
 const cssnano = require('cssnano');
+sass.compiler = require('sass');
```
tylersticka, line 9: This can be a separate PR, but I would find it easier to manage things if the JavaScript and CSS build tasks were in their own files rather than in one big file.

calebeby: :+1: I can do that in a separate PR

.eslintrc.json
```diff
@@ -16,5 +16,23 @@
...
+      "parserOptions": {
+        "project": "./tsconfig.json"
+      },
+      "extends": [
+        "plugin:@typescript-eslint/eslint-recommended",
+        "plugin:@typescript-eslint/recommended",
+        "plugin:@typescript-eslint/recommended-requiring-type-checking"
+      ],
+      "rules": {
+        "@typescript-eslint/explicit-function-return-type": "off", // inference is usually useful
```
Josh68, line 33: @calebeby, does this mean you never need to specify the expected return type from functions?

calebeby: It means that ESLint won't yell at you telling you that you have to. If this rule was enabled, it would require you to specify a return type for every function.

TS type inference is really helpful. I don't think it is useful to specify a return type from every function. Most of the time, TS knows the return type without you telling it. In some cases it is necessary to specify a return type (recursive function that returns a call to itself, you must specify a return type, but TS will tell you that, ESLint doesn't need to). Other times it is helpful to but not required (like when you have multiple branches that each have their own returns, or if you are returning an object that you want to be validated within the function rather than wherever the function is called).

src/components/button/index.ts
Josh68: `npm run type`

- Pass
- Remove `: string`
- error TS7006: Parameter 'a' implicitly has an 'any' type

👍

.eslintrc.json
Josh68: I amended my comment before refreshing this to see your response. That's helpful. At some point, somewhere, this should be in a guide.

src/components/button/index.ts
Josh68: I guess I will say I feel fine about you removing this now, unless someone else wants to run it, break it, and run it again.

Josh68: Obviously, it also passed here:

image

calebeby:
@Josh68 On other projects with this kind of setup, I've seen that error when I create a new .ts file, in VSCode but not if I run ESLint in the CLI. When I reload VSCode, the error goes away. Is that the case here?

gerardo-rodriguez:
> My comments appear to have been addressed. I defer approval to @cloudfour/dev. In particular, I'd love for @gerardo-rodriguez to review.

I plan on taking a peek at this PR today. 😉

gerardo-rodriguez:
> * `cloudfour-patterns.mjs`: This file should have the un-typed version of the button function, in the ES modules format.

@calebeby Is there any concern over the `.mjs` extension?

> To get modules to work correctly in a browser, you need to make sure that your server is serving them with a Content-Type header that contains a JavaScript MIME type such as text/javascript. If you don't, you'll get a strict MIME type checking error along the lines of "The server responded with a non-JavaScript MIME type" and the browser won't run your JavaScript. Most servers already set the correct type for `.js` files, but not yet for `.mjs` files. Servers that already serve `.mjs` files correctly include GitHub Pages and http-server for Node.js.

Or do we at least need to add a note about this somewhere? 🤔 

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#Aside_%E2%80%94_.mjs_versus_.js

gerardo-rodriguez:
@calebeby Question...I have the following in my `dist/` directory:
```
dist/
├── cloudfour-patterns.js
├── cloudfour-patterns.min.js
├── cloudfour-patterns.mjs
├── cloudfourPatterns.js
├── cloudfourPatterns.min.js
├── cloudfourPatterns.mjs
├── index.d.ts
├── standalone.css
└── standalone.min.css
```
Notice I have both `cloudfour-patterns.*` and `cloudfourPatterns.*`.

Should we be cleaning up the `dist/` directory as part of the `build` task? This can be a different PR/GH issue to figure out as I don't feel it's related to this PR. But figured I'd ask. 🙂

calebeby:
That is a good point, I did not think about different servers not specifying the correct MIME type for this file.

But I am not sure that that is a common use case. That would require someone to be using ES Modules in the browser directly without a build tool like rollup, and it would require them to be serving up their `node_modules` directly too, which seems really unlikely. I think it is much more common for people to do something like `import {} from "https://unpkg.com/@cloudfour/patterns?module"`, which should work because unpkg serves the right mime type.

I noticed that preact exports an identical `.mjs` and `.module.js`. I assume `.mjs` is used by node and `.module.js` is used by webpack? I know that at some point webpack had an issue with .mjs files, but I don't know if that is still a problem: https://github.com/preactjs/preact/pull/1425

I will look into how other packages handle this. I think the use case of needing `.module.js` is uncommon enough that we shouldn't need to handle it.

calebeby:
Good idea about clearing out the `dist` folder. I can do that in a follow-up PR that will also separate the js task and css task into separate files

gerardo-rodriguez:
> Confused, though, because when I try to import in another component, the typing is fine, but I get an error like this on import: typescript-eslint/typescript-eslint#967

> @Josh68 On other projects with this kind of setup, I've seen that error when I create a new .ts file, in VSCode but not if I run ESLint in the CLI. When I reload VSCode, the error goes away. Is that the case here?

@calebeby I was able to reproduce what is described above. I closed and restarted VSCode and the error went away. ¯\_(ツ)_/¯

.eslintrc.json
```diff
@@ -3,7 +3,7 @@
   "globals": {
     "Prism": true
   },
-  "plugins": ["react"],
+  "plugins": ["react", "@typescript-eslint"],
```
gerardo-rodriguez, line 6: For my own understanding, what is this `react` plugin for? 🤔

calebeby: It is used for linting .mdx files. The react plugin fixes ESLint's JSX support, by marking variables as "used", if I remember right. I think that without it something like this would fail:

```js
import {Component} from 'package'
// It would say that component is imported but never used

<Component />
```

.eslintrc.json
```diff
@@ -16,5 +16,23 @@
...
+    {
+      "files": ["*.ts", "*.tsx"],
+      "parser": "@typescript-eslint/parser",
+      "parserOptions": {
+        "project": "./tsconfig.json"
+      },
+      "extends": [
+        "plugin:@typescript-eslint/eslint-recommended",
+        "plugin:@typescript-eslint/recommended",
+        "plugin:@typescript-eslint/recommended-requiring-type-checking"
```
gerardo-rodriguez, line 30: @calebeby In the [testingjavascript.com](https://testingjavascript.com/) **Static Analysis Testing JavaScript Applications** module, Kent adds the [`eslint-config-prettier/@typescript-eslint` plugin](https://github.com/prettier/eslint-config-prettier). Does it make sense to add here?  🤔 

https://github.com/kentcdodds/static-testing-tools/blob/34d73e3251a0dd8204b2629b27b41659fd8f054d/.eslintrc#L17

I feel a bit over my element so I'm mostly curious because I don't understand. 🙂

src/components/button/index.ts
gerardo-rodriguez: Reminder to remove this file after the review is done. 🙂

.eslintrc.json
calebeby: Added :heavy_check_mark:

calebeby:
@gerardo-rodriguez The build-types task was failing after I deleted the `button.ts` file. I disabled that task for now, leaving a comment to re-enable as soon as we add .ts files.

PR #682, opened 2020-05-04
Cleanup gulp tasks

calebeby:
## Overview

Cleans up gulp tasks based on discussions in #677

- JS, Sass, and Types tasks are separated into different files, and can be run individually.
- Adds clean task
- Makes several changes that will make these files eventually more ESM-friendly
  - Removes require-dir, this is not doable with static modules in ESM
  - Changes exports to all be camel case so that they will be able to be exported in ESM. This changes the names of the gulp tasks as well

## Testing

- Make sure all pre-processing related npm tasks can still be run the same way
- Create extra files in `dist` and run `npm run build` and make sure they get removed
- Make sure that running each of the individual build tasks works via `npx gulp buildSass`, etc.


gulpfile.js/index.js
```diff
@@ -1,12 +1,16 @@
...
+module.exports = {
+  watchPreprocess: require('./tasks/watch-preprocess'),
+  theoToMDX: require('./tasks/theo-to-mdx'),
+  svgToTwig: require('./tasks/svg-to-twig'),
+  build: require('./tasks/build'),
+  buildSass: require('./tasks/build-sass'),
+  buildJS: require('./tasks/build-js'),
+  buildTypes: require('./tasks/build-types'),
+  clean: require('./tasks/clean'),
+};
```
tylersticka, line 16: Why?

calebeby: - It may make sense for a single file to export multiple tightly-related tasks
- Static require's are more explicit and less magic
- Static require's will very easily be translated into `import`s once we make that transition

tylersticka:
@calebeby There are aspects of this PR that make complete sense. There are others I have real reservations about.

After considering overnight the reasons why I have so much trepidation, I thought it might be helpful to provide some additional context.

The first bit of context is that I'm going to be far more selfish and opinionated when it comes to review of these pieces. This is because there's a very good chance that I will be called upon to troubleshoot aspects of our build tasks if they fail in the future. I apologize in advance if this is irritating, but I have to review everything build-related from the perspective of "How will I handle this when something breaks and I haven't looked at the code in 6–12 months?"

On that second point, one of my goals with this version of the project was to eliminate levels of abstraction from individual build tasks.

In the Drizzle version of this project, a huge pain point was that modifying a build task required both spelunking through giant files _and_ coordinating between multiple files in multiple repositories. Here's a summary of what was required to modify those build tasks:

File | Affected Tasks | Local Dependencies | External Dependencies
--- | --- | --- | ---
`config.js` | build<br>clean<br>copy<br>css<br>drizzle<br>js<br>serve<br>watch | – | core-gulp-tasks<br>drizzle-builder<br>Handlebars<br>PostCSS<br>Rollup
`gulpfile.js` | build<br>clean<br>copy<br>css<br>default<br>drizzle<br>frontend<br>js<br>serve<br>watch | `config.js` | core-gulp-tasks<br>core-hbs-helpers<br>drizzle-builder<br>Handlebars<br>PostCSS<br>Rollup

(I'm defining external dependencies here loosely, as in things you would have to have knowledge of or would have to modify to revise and update.)

While it _seemed_ like a good idea at the time to abstract things into their own repositories or shared configuration, in practice it meant we were taking _one chunk of logic_ (for example, PostCSS build) and spreading it out across three files in two separate repositories. We were prioritizing keeping our _configuration_ DRY instead of our build tasks self-contained.

Prior to this PR, here's how our current repository looks in comparison:

File | Affected Tasks | Local Dependencies | External Dependencies
--- | --- | --- | ---
`.storybook/*` | build<br>storybook | `.theo/*`<br>`glob-sass-importer.js`<br>`.svgo.yml` | Storybook
`.theo/*` | build<br>preprocess<br>storybook | – | Theo
`gulpfile.js/tasks/build.js` | build | `.theo/*`<br>`babel.config.js` | Babel<br>PostCSS<br>Rollup<br>Sass<br>Typescript
`gulpfile.js/tasks/svg-to-twig.js` | build<br>preprocess<br>storybook | `.svgo.yml` | –
`gulpfile.js/tasks/theo-to-mdx.js` | preprocess<br>storybook | `.theo/*` | Theo
`gulpfile.js/tasks/watch-preprocess.js` | storybook | `./svg-to-twig.js`<br>`./theo-to-mdx.js` | –
`.svgo.yml` | build<br>preprocess<br>storybook | – | SVGO
`babel.config.js` | build<br>storybook | – | Babel
`glob-sass-importer.js` | build<br>storybook | – | Sass
`postcss.config.js` | build<br>storybook | – | PostCSS

Overall, this makes a lot more sense to me. But the last column for `gulpfile.js/tasks/build.js` stood out to me, which is why I suggested splitting it up.

Here's how things look as of this PR:

&nbsp; | File | Affected Tasks | Local Dependencies | External Dependencies
--- | --- | --- | --- | ---
&nbsp; | `.storybook/*` | build<br>storybook | `.theo/*`<br>`glob-sass-importer.js`<br>`.svgo.yml` | Storybook
&nbsp; | `.theo/*` | build<br>preprocess<br>storybook | – | Theo
⚠️ | `gulpfile.js/helpers.js` | build<br>buildJS<br>buildTypes<br>buildSass | – | Rollup<br>Typescript
⚠️ | `gulpfile.js/index.js` | build<br>buildJS<br>buildSass<br>buildTypes<br>clean<br>preprocess<br>svgToTwig<br>theoToMDX<br>watchPreprocess | `gulpfile.js/tasks/*` | –
✅ | `gulpfile.js/tasks/build-js.js` | build | `../helpers.js`<br>`babel.config.js` | Babel<br>Rollup<br>Typescript
✅ | `gulpfile.js/tasks/build-sass.js` | build | `../helpers.js`<br>`postcss.config.js` | PostCSS<br>Sass
✅ | `gulpfile.js/tasks/build-types.js` | build | `../helpers.js` | Typescript
❓ | `gulpfile.js/tasks/build.js` | build | `./build-js.js`<br>`.build-sass.js`<br>`.clean.js` | –
❓ | `gulpfile.js/tasks/clean.js` | clean | `../helpers.js | –
&nbsp; | `gulpfile.js/tasks/svg-to-twig.js` | build<br>preprocess<br>storybook | `.svgo.yml` | –
&nbsp; | `gulpfile.js/tasks/theo-to-mdx.js` | preprocess<br>storybook | `.theo/*` | Theo
&nbsp; | `gulpfile.js/tasks/watch-preprocess.js` | storybook | `./svg-to-twig.js`<br>`./theo-to-mdx.js` | –
&nbsp; | `.svgo.yml` | build<br>preprocess<br>storybook | – | SVGO
&nbsp; | `babel.config.js` | build<br>storybook | – | Babel
&nbsp; | `glob-sass-importer.js` | build<br>storybook | – | Sass
&nbsp; | `postcss.config.js` | build<br>storybook | – | PostCSS

**Legend:**
- ✅ Feels self-contained, straightforward to modify later.
- ❓ Unclear why this is a Gulp task and not a `package.json` script.
- ⚠️ Seems to take a step back toward the previous version's shared configuration strategy.

Because of this, I just can't shake the feeling that this PR represents a _lateral change_ for me in terms of my ability to maintain the project moving forward. Some files are now more focused, but some feel like we've simplify _moved_ the previous complexity in a different way. It feels _different_ but not necessarily _better_.

## Suggestions

If this were my PR, I'd explore the following changes:

- I'd keep the old `require-dir` strategy for `index.js`. I understand your reasoning, but I'm not convinced that it's a helpful change for me, and I'm even less convinced it's necessary for this PR.
- I would combine `build-js` and `build-types`, since they are the primary source of _spread_ across multiple configuration files.
- I would get rid of the shared `helpers.js` file. I know, I know, we'll repeat the output directory name, but I'm okay with that.
- I would replace the `clean.js` task with a `package.json` script using Eliminate's CLI.
- I would replace the `build.js` task with a `package.json` script using `run-s` and/or `run-p`.

## Am I being a curmudgeon?

As I mentioned up top, I have reasons for being more selfish and opinionated when it comes to this project. That said, @spaceninja has done a particularly solid job of keeping me honest when it comes to this sort of thing on a recent project.  If you want to convince me that I'm wrong (in whole or in part), his 👍 would go a long way.

calebeby:
@tylersticka Thanks for taking the time to make this detailed comment, you make a lot of really good points. Here are my thoughts on some of them:

## helpers.js

I don't like this file either :joy: I agree with inlining the outDir :+1: 

## Combine build-js and build-types

I agree with combining these files, but I think that they should still be separate gulp tasks. Their implementations are closely related, but for me wanting to build just the types or just the js separately is a common use case.

## Build/Clean

I think there should still be a `build` gulp task. To me it makes a lot of sense to have a task that does everything required to prepare the `dist` folder for publishing. It feels like it would be out-of-place if that one task was an npm script, while all the other tasks are gulp tasks

tylersticka:
> I don't like this file either 😂 I agree with inlining the outDir 👍

Hooray for agreement!

> I agree with combining these files, but I think that they should still be separate gulp tasks. Their implementations are closely related, but for me wanting to build just the types or just the js separately is a common use case.

Fair enough!

> I think there should still be a `build` gulp task. To me it makes a lot of sense to have a task that does everything required to prepare the `dist` folder for publishing.

Totally, no disagreement here.

> It feels like it would be out-of-place if that one task was an npm script, while all the other tasks are gulp tasks

I really like Gulp. I'm proud to be a Gulp 4 contributor (helping design its new tree view forever ago). And the whole reason I moved to Gulp from Grunt years ago is that I really dislike traversing big ol' JSON configuration files, so I have no love for `package.json` as a task manager.

Yet on this project, I decided to try a new strategy: If it can be easily done as a `package.json` script, do it there. If not, create a Gulp task. Basically, treat Gulp as an optional tool for authoring complex `package.json` scripts. That's why there are currently four Gulp tasks and 14 `package.json` scripts (not counting the ones that expose Gulp tasks).

I decided to try this for a few reasons:

- Some dependencies ship with really robust CLIs, and it always felt a bit odd or unnecessary to wire up a Gulp task just to recreate that.
- Some `watch` commands work better on their own than via a Gulp task. For example, it'd be way worse for us to trigger a time-consuming static Storybook build on file changes when we can just run Storybook and have it watch for changes itself.
- Some resources that require a plugin to work via Gulp introduce a new dependency in our chain. We've ran into issues once or twice where a bug fix was blocked by whoever was maintaining the Gulp plugin.

All that said, we may not have struck the right balance yet, and there may be benefits to simpler Gulp tasks if they are dependencies on _other_ Gulp tasks. Of the feedback I've provided so far, I feel like we have the most wiggle room on this one.

calebeby:
@tylersticka 

- I switched build and clean to be only in package.json
- I combined buildJS and buildTypes to the same build-scripts file, so that the shared logic is in that file rather than a shared file
- I still didn't switch to require-dir because that does not allow us to export multiple related tasks from the same file, which we are doing now since the buildJS and buildTypes tasks are in the same file.

calebeby:
Thanks for your helpful feedback Tyler!

Issue #479, opened 2020-02-20
Site header pattern(s)


dromo77:
Create a site header pattern to be used throughout the site.

<img width="1423" alt="site header" src="https://user-images.githubusercontent.com/42841342/74989345-f7a2f900-53f4-11ea-8850-bfce75cbaee9.png">

tylersticka:
- We should maintain our "skip to main content" behavior.
- The animation should respect `prefers-reduced-motion`.
- The current animation throws `will-change` on the whole document, which is not ideal.
- We should aim for as smooth an animation as possible (ideally 60fps).

tylersticka:
## CSS To-dos

- [ ] Stack at small screens
- [ ] Two rows breakpoint
- [ ] Single row breakpoint
- [ ] Mouse effects
- [ ] Keyboard effects
- [ ] Improve alignment
- [ ] Reconcile with theming strategy

tylersticka:
I'm going to close this issue and open a new one for the remaining development work.

PR #687, opened 2020-05-05
Primary navigation markup

tylersticka:
## Overview

I'm extremely proud of the accessibility knowledge in this team. Although the accessibility challenges of our site pale in comparison to those of our main projects, I see that as an invitation to poke a bit at the underpinnings of what we're doing.

Case-in-point: A primary navigation PR consisting _only_ of markup! Well, markup and a pinch of Twig template tags.

I based this revised markup structure on a recent client project with a couple of small changes. Here's a summary of everything that's changed from our what's on our live site:

- Moved the "skip to main content" link _out_ of the `<header>`, since that's where [Heydon](https://heydonworks.com/latest/) seems to put it on all of his projects.
- Replaced the menu button with an actual `button` element, and updated its fallback text to be more descriptive.
- Added alternate home fallback text depending on whether or not you are currently on the home page.
- Added `focusable="false"` to the logo to avoid competition with surrounding links in old browsers.
- Replaced `role="navigation"` on the `<nav>` element with `aria-label="Main Menu"`.
- Added `aria-current="page"` attribute to current menu link (if any).

## Screenshots

Screen Shot 2020-05-05 at 2 10 21 PM

## Testing

Please kick the tires on this stuff and let me know if you have any feedback! Anything we could be doing better or differently so far to make this as inclusive as possible?

---

- See #479 

/CC @gerardo-rodriguez @Paul-Hebert

### DIFF
--- a/src/components/sky-nav/sky-nav.stories.mdx
+++ b/src/components/sky-nav/sky-nav.stories.mdx
@@ -1,0 +1,46 @@
+import { Story, Preview, Meta } from '@storybook/addon-docs/blocks';
+import template from './sky-nav.twig';
+// https://timber.github.io/docs/guides/menus/
+const menu = {
+  items: [
+    {
+      link: '/does',
+      title: 'What We Do',
+      current: true,
+    },
+    {
+      link: '/believes',
+      title: 'Our Approach',
+    },
+    {
+      link: '/made',
+      title: 'Our Work',
+    },
+    {
+      link: '/thinks',
+      title: 'Articles',
+    },
+    {
+      link: '/presents',
+      title: 'Speaking',
+    },
+    {
+      link: '/contributes',
+      title: 'Labs',
+    },
+    {
+      link: '/is',
+      title: 'Team',
+    },
+  ],
+};
+
+<Meta title="Components/Sky Nav" />
+
+# Sky Nav
+
+This will eventually be our primary navigation component. Right now, we're just critiquing some markup updates in terms of their accessibility.
+
+<Preview>
+  <Story name="Markup (WIP)">{template({ menu })}</Story>
+</Preview>
--- a/src/components/sky-nav/sky-nav.stories.mdx
+++ b/src/components/sky-nav/sky-nav.stories.mdx
@@ -1,6 +1,5 @@
 import { Story, Preview, Meta } from '@storybook/addon-docs/blocks';
 import template from './sky-nav.twig';
-// https://timber.github.io/docs/guides/menus/
 const menu = {
   items: [
     {
@@ -40,6 +39,8 @@ const menu = {
 # Sky Nav
 
 This will eventually be our primary navigation component. Right now, we're just critiquing some markup updates in terms of their accessibility.
+
+We are tentatively using the same template properties as [Timber Menus](https://timber.github.io/docs/guides/menus/).
 
 <Preview>
   <Story name="Markup (WIP)">{template({ menu })}</Story>
--- a/src/components/sky-nav/sky-nav.twig
+++ b/src/components/sky-nav/sky-nav.twig
@@ -26,7 +26,10 @@
     <span class="u-hidden-visually">Toggle Main</span>
     Menu
   </button>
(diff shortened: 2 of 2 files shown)


src/components/sky-nav/sky-nav.stories.mdx
```diff
@@ -0,0 +1,46 @@
...
+  ],
+};
+
+<Meta title="Components/Sky Nav" />
+
+# Sky Nav
+
+This will eventually be our primary navigation component. Right now, we're just critiquing some markup updates in terms of their accessibility.
+
+<Preview>
```
Paul-Hebert, line 44: I wonder if it's worth outlining the data structure expected by this menu? Especially the `is_home`/`item.current` stuff? Or is that overkill and future devs should figure it out by reviewing the code?

src/components/sky-nav/sky-nav.twig
```diff
@@ -0,0 +1,40 @@
...
+      width: '80',
+      height: '80',
+      focusable: 'false'
+    } only %}
+  </a>
+  <button type="button">
+    <span class="u-hidden-visually">Toggle Main</span>
+    Menu
+  </button>
+  <nav aria-label="Main Menu">
```
Paul-Hebert, line 29: I know I've implemented a similar markup pattern in a past project, but now that I'm reviewing again I wonder whether it would be better to use `aria-labelledby`? I'm curious what others think.

**Pros:**

- `aria-labelledby` is more reliably translated by translation tools

**Cons:**

- `aria-labelledby` would require an additional HTML element which screen readers would need to read. It might be overly verbose.

---

I'm not sure what the "right" answer is here, but figured I'd mention my thoughts in case others had more insight or felt strongly.

src/components/sky-nav/sky-nav.twig
```diff
@@ -0,0 +1,40 @@
...
+    <ul>
```
Paul-Hebert, line 30: I've been curious lately about skipping the unordered list and directly inserting links into a nav, but [this comment by expert Léonie Watson](https://css-tricks.com/navigation-in-lists-to-be-or-not-to-be/#comment-273913) convinced me we should keep the unordered list.

> A quick straw poll of around 30 screen reader users this afternoon suggests that many find lists useful in this context. Some said a list may not be nescessary if a heading were present before the navigation block, and some said they preferred to use the features of their screen reader to access links but acknowledged that if they were looking for a specific group of links that wouldn’t help.

I'm assuming that when we get to CSS we'll be adding `list-style: none;` to this. With that in mind, we should maybe add `role="list"` to [maintain list semantics in Safari](https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html).

src/components/sky-nav/sky-nav.twig
```diff
@@ -0,0 +1,40 @@
...
+{#
+  This may require `role="banner"` if nested in something other than `body`.
+  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Banner_Role
+#}
+<header>
+  <a href="/"
+    {% if is_home %}
+      aria-current="page"
+    {% else %}
+      title="Return to home"
```
Paul-Hebert, line 14: I'm curious if this is needed in addition to the visually hidden text? Would this be redundant for a screen reader?

Maybe instead of adding a title we could conditionally change the visually hidden text?

src/components/sky-nav/sky-nav.twig
```diff
@@ -0,0 +1,40 @@
+{# https://webaim.org/techniques/skipnav/ #}
+<a href="#main">
```
Paul-Hebert, line 2: I wonder if it would be worth adding content within a `<main id="main>` element for the demo? Or will we be creating a page content pattern down the line that will showcase that behavior?

src/components/sky-nav/sky-nav.stories.mdx
tylersticka: We definitely should! I left a link in the story next to the object, but we should definitely do more.

Issue #657, opened 2020-04-24
Update icons


tylersticka:
I like our current icon set, but there are a few small updates they would benefit from in the short term.

## Current Source File Location

In our shared project file location, the Illustrator files for icons are in Icons ▸ Glyphs.

## Retire some icons

There are a few icons we are never using or will likely stop using that we can remove:

- [x]  **dot**: Was used by the radio pattern, which no longer exists since we never used it.
- [x] **lanyrd**: This site has been broken [since 2019](https://en.wikipedia.org/wiki/Lanyrd).
- [x] **notification:** It's unclear what this icon was meant to do that wasn't covered by **notifications**. It isn't used in any templates.

## Organization

Currently our icons sit in one folder, but I think these might benefit from some reorganization:

- [x] Original icons should sit at the root of the icons directory. These include glyphs like arrows, link, notifications, etc.
- [x] Icons representing third-party brands should be moved to a "brands" subdirectory. These include icons for publications (A List Apart, Smashing, etc.), social media websites, etc.
- [x] Our logo should be moved _out_ of icons, possibly one level up.

This organization should also be reflected in our source files, with separate files for each type of asset.

Why? I'll get to that next…

## Sizing

The biggest design opportunity with our icons is to reconcile differences in perceived visual size.

Some icons look proportionate to one another. Notice how the smaller caret (used for dropdowns) is smaller in proportion to the check mark, the same way that the ▾ character is typically rendered compared to ✔️:

<img width="322" alt="Screen Shot 2020-04-24 at 4 30 08 PM" src="https://user-images.githubusercontent.com/69633/80264394-e2428900-8648-11ea-82e5-69ae7a800998.png">

But this is not true of all icons. Notice how small the notification bell 🔔  feels in comparison to ➕ :

<img width="307" alt="Screen Shot 2020-04-24 at 4 31 37 PM" src="https://user-images.githubusercontent.com/69633/80264469-19189f00-8649-11ea-8549-43fc42e4b35e.png">

The [Material Icon Design Principles](https://material.io/design/iconography/system-icons.html#grid-and-keyline-shapes) page does a good job of illustrating some of the concepts between icon sizing and composition to promote consistency and prevent these issues.

This is also why I feel it's important to organize our icons better. By sizing them relative to each other, maybe the Lanyrd (RIP) icon (which extends dramatically into the top-left corner) wouldn't be threatening to devour the GitHub icon (which is confined to a circle):

[...]

PR #705, opened 2020-05-15
Design: adds better, larger Cloud Four logo asset

derekshirk:
## Overview

This pull request represents the remaining task identified in #657. 

> Finally, the logo is used primarily as a branding element (not an icon). Arbitrarily squashing it into our 24×24 grid raises three issues:
> 
> - We have to scale the element up for its primary use case, which is silly.
> - Its coordinates will have more decimal points, which increases the size of the markup required to define the shape.
> - Decimal points may be rounded for performance, which could result in quality loss.
>
> So to summarize the goals here:
> 
> Resize icons as a family so they will read as visually consistent, rather than simply filling the viewBox.
>
> - [ ] Treat logo asset as separate and don't squash it down like an icon.

I opted to size the asset's viewbox off of what I observed to be the logo size on cloudfour.com when viewed at the largest breakpoint. 

## Screenshots

Screen Shot 2020-05-15 at 2 12 50 PM

To avoid confusion with between the Design > Icons > "Brand" icons and the Design > Brand logo  I opted to rename the "brand icons" section to "Third Party Icons".

Before: 
Screen Shot 2020-05-15 at 1 59 25 PM

After:
Screen Shot 2020-05-15 at 2 00 34 PM

## Testing

1. View [deploy preview](https://deploy-preview-705--cloudfour-patterns.netlify.app/?path=/docs/design-brand--logo)

2. Review updates to the logo SVG asset, viewbox dimensions, width, height and formatting

---

- see #657


src/design/icons/icons.stories.mdx
```diff
@@ -41,7 +41,7 @@ For icon usage examples, see [the icon object](/?path=/docs/components-icon--bas
 
 <IconGallery>{iconElements}</IconGallery>
 
-# Brand Icons
+# Third Party Icons
```
tylersticka, line 44: I believe "Third-Party" should be hyphenated since it is an adjective?

src/design/brand.stories.mdx
```diff
@@ -1,24 +1,15 @@
...
-  // Retains subdirectory if we want to organize with those in the future
-  const name = relative('.', key).replace(extname(key), '');
-  return (
-    <IconItem name={name}>
-      <Logo class="c-icon" />
-    </IconItem>
-  );
-});
+import { Story, Preview, Meta } from '@storybook/addon-docs/blocks';
+import brandLogo from './demo/logo.twig';
```
tylersticka, line 2: Did you try importing `'../assets/brand/logo.svg'` oor `'../assets/brand/logo.svg.twig'` or directly before creating the demo file?

derekshirk: No I did not! 

I try that out. Thanks.

derekshirk:
I just updated the deploy preview link in the testing instructions.

src/design/icons/icons.stories.mdx
derekshirk: Good point. Regarding case, my instinct is to use "Third-party Icons" instead of of "Third-Party Icons". 

Let me know if you have a preference.

src/design/brand.stories.mdx
derekshirk: My commit messages got flip-flopped. Here is what worked for me. 

d29d0b6

tylersticka: It looks like my suggestion broke linting. Which makes a bit of sense since Twig files don't get tracked.

If `.svg` (no `.twig`) did not work, then I suppose the demo strategy is fine! But we may want to add a comment like "Must track a Twig template to satisfy CI" or something.

derekshirk:
Investigating the CI / Iint errors...

src/design/brand.stories.mdx
derekshirk: Thank you. I missed your comment here, but I came to the same conclusion. 

Since, importing the `.svg` generates an error, I think you suggestion for adding a comment is a great idea. 

Screen Shot 2020-05-18 at 9 25 13 AM

Example of importing `.svg`

PR #716, opened 2020-05-18
Add Expanding Text Input

### LINKED ISSUE #683: Add JS for expanding text input
Context: https://github.com/cloudfour/cloudfour.com-patterns/issues/653#issuecomment-619286047

- Create textarea js file(s), copying the functionality of https://github.com/cloudfour/cloudfour.com-patterns/blob/master/src/assets/toolkit/scripts/lib/component/elastic-textarea.js
- ~Write tests in Jest or Cypress with testing-library (and set up these tools as necessary)~
- Make sure that built JS correctly exports the new code

spaceninja:
## Add Expanding Text Input

This PR adds the JavaScript for the Expanding Text Input option for the Input component.

## Screenshots

Screen Shot 2020-05-20 at 11 11 58 AM

## Testing

- Review the [preview environment](https://deploy-preview-716--cloudfour-patterns.netlify.app/?path=/docs/components-input--elastic-textarea) and compare to the [existing component](https://cloudfour-patterns.netlify.app/patterns/components/input.html#elastic-textarea). There should be no difference in functionality.
- Run `npm run build` and review the compiled JS for the pattern library. Ensure the code for the expanding text area is present and not malformed.

## Code Review

This is the first JS we've added to the new pattern library, so please review this code not just with an eye towards functionality, but also as a representative for future pattern library contributors of best practices for JS in the pattern library, based on [these principles](https://github.com/cloudfour/cloudfour.com-patterns/issues/683#issuecomment-629532934).

- Does this code use TypeScript properly? Is it a good representative of how we want to write TS going forward?
- This code replaces the old Class with a factory function. Does it work the way the team expects?
- There is no `toolkit.js` file, because Storybook requires applying the JS inline in the Story. Does this seem reasonable and sustainable?
- Does the JS documentation for pattern library consumers make sense? Is this how we'd like to present this information?

Note, it's okay if you have no strong thoughts on any of these questions. They're meant to be things to consider. If you look at the code and your reaction is "this is fine", that's okay! More than anything else I'm looking for red flags and bad practices.

Mostly, this is a chance to put up some code and give the team a chance to pump the brakes early if this is NOT how we want to move forward.

---

- Fixes #683


src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
...
+<p>
+  <small>* see also the JavaScript usage notes below</small>
+</p>
+
+<Preview>
+  <Story name="Elastic Textarea" height="350px">
+    {() => {
+      // Use storybook hooks to trigger JS after story renders
+      // @see https://github.com/storybookjs/storybook/issues/7786
+      useEffect(() => {
```
calebeby, line 60: I read through the issue and I'm still confused about this? Is this different from a react hook? What exactly is it doing?

src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
...
+        // Select all elastic textarea elements, run create function on each
+        [...document.querySelectorAll('.js-elastic-textarea')].map(
```
calebeby, line 62: Since we aren't using the result of the newly-created elastic textarea instances, can we change this to `.forEach` and remove the `[...`?

src/components/input/input.scss
```diff
@@ -52,11 +52,11 @@
   }
 
   /**
-   * Multi-line types should be taller
+   * Multi-line types get their height from the `rows` attribute
    */
 
   @at-root textarea#{&} {
-    height: sizes.$control-height-multiline;
+    height: auto;
```
Paul-Hebert, line 59: I wonder if it still makes sense to have a `min-height`? Otherwise a 1 row `textarea` may just look like an `input`? Though maybe that's okay?

src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
     })}
   </Story>
 </Preview>
+
+## Elastic textarea
+
+Add `js-elastic-textarea` to a `<textarea>` element\* to have it expand as the user types more than will fit.
+
+If the `rows` attribute is set, it will be used as the minimum height.
```
calebeby, line 49:
```suggestion
If the `data-min-rows` attribute is set, it will be used as the minimum height.
```

src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
     })}
   </Story>
 </Preview>
+
+## Elastic textarea
+
+Add `js-elastic-textarea` to a `<textarea>` element\* to have it expand as the user types more than will fit.
```
Paul-Hebert, line 47: Is this accurate? It seems like the class name you added would be reliant on the JavaScript you write? e.g. you might want `.js-comment-field` to be the class?

Paul-Hebert: Maybe "Call `createElasticTextArea(yourTextareaSelector)` to have your textarea expand as the user types more than will fit." or something like that?

src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
...
+
+If the `rows` attribute is set, it will be used as the minimum height.
+
+<p>
+  <small>* see also the JavaScript usage notes below</small>
+</p>
```
Paul-Hebert, line 53: I'm not sure how I feel about this footnote style. I wonder if we could add an inline link to below instead?

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @param {HTMLTextAreaElement} textarea - the target `textarea` element
+ */
+const shrinkElasticTextArea = (textarea: HTMLTextAreaElement) => {
+  const minRows = Number(textarea.dataset.minRows);
+  let rows = Number(textarea.getAttribute('rows')) || minRows;
+  let isScrolling = textarea.scrollHeight > textarea.clientHeight;
+
+  // Shrink until we hit the minimum rows or start scrolling
+  for (let i = rows; i >= minRows; i--) {
```
calebeby, line 34: Can this use a while loop like `growElasticTextArea` does?

Something along the lines of:

```js
// Shrink until we hit the minimum rows or start scrolling
while (!isScrolling && rows > minRows) {
  rows--
  ...
}
if (isScrolling) growElasticTextArea(textarea)
```

src/components/input/input.stories.mdx
Paul-Hebert, line 62: I'm not sure `map` is necessary if we're not storing the result of the `map`. Should this just be a `forEach`?

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+  textarea.dataset.lastValue = textarea.value || '';
+
+  // Set the minimum rows to the initial `rows` value or 2
+  textarea.dataset.minRows = textarea.getAttribute('rows') || '2';
+
+  // Add an event listener to run the update function after input events
+  textarea.addEventListener('input', updateElasticTextArea);
+
+  // Fire an input event to set the initial size correctly
+  textarea.dispatchEvent(new Event('input'));
```
calebeby, line 99: Can we call updateElasticTextArea directly instead of firing an event?
There may be other input listeners on this textarea and they don't need to be called

src/components/input/input.stories.mdx
Paul-Hebert: Actually, is the `map`/`forEach` necessary if the demo only has one textarea? Could we just target a single node and simplify the code?

```js
useEffect(() => { 
   createElasticTextArea(document.querySelector('js-elastic-textarea');
}, []);
```

src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
...
+          }
+        );
+      }, []);
+      return textDemo({
+        type: select(
+          'type',
+          ['text', 'email', 'search', 'date', 'textarea'],
+          'textarea'
+        ),
+        value: text('value', 'a\nb\nc\nd\ne'),
```
Paul-Hebert, line 74: Nit: Could we get some cloud four branded content here? Maybe a comment Jason or someone's written in a blog post comment thread?

Paul-Hebert: Also: is it confusing being able to change this either via knob or by typing in the textarea?

src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
...
+        id: text('id', 'demo-elastic'),
+        class: text('class', 'js-elastic-textarea'),
+      });
+    }}
+  </Story>
+</Preview>
+
+### JavaScript Usage
+
+Once you have added the `js-elastic-textarea` class to your `textarea` element, you'll also need to add some JavaScript that runs after the page loads, like so:
```
Paul-Hebert, line 86: Again, you may want to highlight that you can use whatever class or selector you want.

src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
...
+        type: select(
+          'type',
+          ['text', 'email', 'search', 'date', 'textarea'],
+          'textarea'
+        ),
+        value: text('value', 'a\nb\nc\nd\ne'),
+        placeholder: text('placeholder', 'Placeholder text…'),
+        rows: number('rows', '2'),
+        id: text('id', 'demo-elastic'),
+        class: text('class', 'js-elastic-textarea'),
```
Paul-Hebert, line 78: I wonder if the ID and class need to be knobs? Does changing those help me understand the pattern?

src/components/input/input.scss
Paul-Hebert: Does this have the correct height if `rows` is omitted or set to `1`?

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+ * the type directly, since we don't care about the rest of the `event` object.
+ * Unfortunately, that caused some TS problems. If someone finds a solution,
+ * we should update this function to take `{target}` as the only parameter.
+ * @see https://stackoverflow.com/a/28900856/4898045
+ *
+ * @param {Event} event - the input event on the target `textarea`
+ */
+const updateElasticTextArea = (event: Event) => {
+  const textarea = event.target as HTMLTextAreaElement;
+  const lastValue = textarea.dataset.lastValue || '';
```
calebeby, line 63: Instead of storing the last value in a data attribute, I'm wondering if we could store it as an "instance variable" within the `createElasticTextArea` instance and move `update` into the `createElasticTextArea` function:

```js
export const createElasticTextArea = () => {
  let lastLength = textarea.value.length
  const update = (event) => {
    if (lastLength > ...) {}
    // ...
    lastLength = element.length
  }
  // ... listeners
}
```

I think it might also make sense to move `grow` and `shrink` into the `createElasticTextArea` function, I'd be curious to hear your thoughts

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
+/**
+ * Grow Elastic TextArea
+ *
+ * Increases the `rows` attribute of a `textarea` until it no longer scrolls.
+ *
+ * @param {HTMLTextAreaElement} textarea - the target `textarea` element
+ */
+const growElasticTextArea = (textarea: HTMLTextAreaElement) => {
+  const minRows = Number(textarea.dataset.minRows);
```
Paul-Hebert, line 9: I wonder if we should make sure this never is below 1?

const minRows = Math.min(Number(textarea.dataset.minRows), 1);

---

What happens if this is not defined on the dataset? Does it fail gracefully?

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
+/**
+ * Grow Elastic TextArea
+ *
+ * Increases the `rows` attribute of a `textarea` until it no longer scrolls.
+ *
+ * @param {HTMLTextAreaElement} textarea - the target `textarea` element
```
calebeby, line 6: Can you remove the duplicated type here? When I set up TS I forgot to disable the ESLint rule that requires that the type is in the JSDoc for `.ts` files. I made that change here: https://github.com/cloudfour/cloudfour.com-patterns/pull/726/files#diff-df39304d828831c44a2b9f38cd45289cR35-R38, could you copy those changes into this PR (or I could make a separate PR)?

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Shrink Elastic TextArea
+ *
+ * Decreases the `rows` attribute of a `textarea` until it begins scrolling.
+ *
+ * @param {HTMLTextAreaElement} textarea - the target `textarea` element
+ */
+const shrinkElasticTextArea = (textarea: HTMLTextAreaElement) => {
+  const minRows = Number(textarea.dataset.minRows);
```
Paul-Hebert, line 29: Same questions as minRows above.

src/components/input/input.ts
gerardo-rodriguez: > Instead of storing the last value in a data attribute, I'm wondering if we could store it as an "instance variable" within the createElasticTextArea instance and move update into the createElasticTextArea function:

I agree with this. 👍

gerardo-rodriguez: > I think it might also make sense to move grow and shrink into the createElasticTextArea function, I'd be curious to hear your thoughts

I was wondering about this as well. It feels like they are a part of the same component. If the `grow` and/or `shrink` functions were reusable elsewhere, then they might live as their own functions and possibly be exported from their own module. But since they are only related to the `createElasticTextArea`, I think it'll be most clear for them to _literally_ live within the `createElasticTextArea` scope. 😉

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+  const textarea = event.target as HTMLTextAreaElement;
+  const lastValue = textarea.dataset.lastValue || '';
+  const newValue = textarea.value;
+
+  // Check whether to grow or shrink
+  if (newValue.length > lastValue.length) {
+    growElasticTextArea(textarea);
+  } else {
+    shrinkElasticTextArea(textarea);
+  }
```
Paul-Hebert, line 71: I'm not sure this is the best way to tell whether we need to grow or shrink. e.g. Take these two values:

**lastValue:**

```
I am a single line but I have lots of characters! I only need one row since I have no returns!
```

**newValue:**
```
I
need
more
rows!
```

In this case the new value is shorter than the old value, but the textarea needs to grow! This probably wouldn't happen if I was typing away, but it could happen if I copy/paste replaced my old text.

Could we use a similar strategy to what we did for `isScrolling`? (I realize this logic is probably carried over from the old implementation)

src/components/input/input.stories.mdx
gerardo-rodriguez: @calebeby My understanding is this is a React hook, under the hood. But I don't know enough about Storybook to confirm this.

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @param {HTMLTextAreaElement} textarea - the target `textarea` element
+ * @returns {object} object containing a `destroy()` method
+ */
+export const createElasticTextArea = (textarea: HTMLTextAreaElement) => {
+  // Save the current value for comparing
+  textarea.dataset.lastValue = textarea.value || '';
+
+  // Set the minimum rows to the initial `rows` value or 2
+  textarea.dataset.minRows = textarea.getAttribute('rows') || '2';
```
Paul-Hebert, line 93: Ahh, I see you're setting the `minRows` here... you can ignore my comments above about `minRows` being unset.

src/components/input/demo/example.twig
```diff
@@ -3,5 +3,5 @@
   {% block element %}
     {% include '../input.twig' %}
   {% endblock %}
-  {% include '../../button/button.twig' with { label: 'Button' } %}
+  {% include '../../button/button.twig' with { label: 'Button' } only %}
```
gerardo-rodriguez, line 6: @spaceninja For my own learning purposes, what does `only` do? I don't know. 🙂

src/components/input/input.ts
Paul-Hebert: I had the same thought. Then we can define all of our variables once and avoid needing to mess with the data attributes in the DOM.

src/components/input/input.stories.mdx
gerardo-rodriguez, line 47:
```suggestion
Add the `js-elastic-textarea` CSS class to a `<textarea>` element\* to have it expand as the user types more than will fit.
```

src/components/input/input.stories.mdx
gerardo-rodriguez: I agree with @calebeby's suggestion. 🙂

src/components/input/input.stories.mdx
gerardo-rodriguez: Oh, good points, @Paul-Hebert!

gerardo-rodriguez: Closing this in favor or @Paul-Hebert's [suggestions](https://github.com/cloudfour/cloudfour.com-patterns/pull/716/files#r428219468).

src/components/input/input.ts
Paul-Hebert: I realize this is an edge case and not a regression so I'd be fine filing an issue and moving on if you'd like

src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
...
+<Preview>
+  <Story name="Elastic Textarea" height="350px">
+    {() => {
+      // Use storybook hooks to trigger JS after story renders
+      // @see https://github.com/storybookjs/storybook/issues/7786
+      useEffect(() => {
+        // Select all elastic textarea elements, run create function on each
+        [...document.querySelectorAll('.js-elastic-textarea')].map(
+          (textarea) => {
+            return createElasticTextArea(textarea);
```
gerardo-rodriguez, line 64: Per @calebeby's [suggestion](https://github.com/cloudfour/cloudfour.com-patterns/pull/716/files#r428217564), we don't need `map` which means we don't need to `return` anything here. 😉

src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
...
+
+Once you have added the `js-elastic-textarea` class to your `textarea` element, you'll also need to add some JavaScript that runs after the page loads, like so:
+
+```js
+// Select all elastic textarea elements, run create function on each
+const elasticTextAreas = [
+  ...document.querySelectorAll('.js-elastic-textarea'),
+].map((textarea) => {
+  return createElasticTextArea(textarea);
+});
```
gerardo-rodriguez, line 94: Should this follow our JS prettier/ESLint guidelines for consistency? How do others feel about my question?

src/components/input/input.stories.mdx
calebeby: Just want to make a note, the createElasticTextArea function accepts an element, not a selector. But I agree with this suggestion otherwise

src/components/input/input.stories.mdx
calebeby: Prettier is being run on this file. I think ESLint is too

src/components/input/input.stories.mdx
```diff
@@ -33,3 +41,64 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
...
+```
+
+The `createElasticTextArea` function adds an event listener to the `textarea`, allowing it to respond to user input by growing or shrinking as needed. It returns an object containing a `destroy()` method, if you need to remove the event listener.
```
gerardo-rodriguez, line 97: Thoughts about explicitly calling out what event we are using? 🙂 
```suggestion
The `createElasticTextArea` function adds an [`input`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) event listener to the `textarea`, allowing it to respond to user input by growing or shrinking as needed. It returns an object containing a `destroy()` method, if you need to remove the event listener.
```

src/components/input/input.ts
gerardo-rodriguez: Good idea, @Paul-Hebert!

gerardo-rodriguez: > What happens if this is not defined on the dataset? Does it fail gracefully?

Sounds like a great candidate for a Jest test! 😄

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+};
+
+/**
+ * Shrink Elastic TextArea
+ *
+ * Decreases the `rows` attribute of a `textarea` until it begins scrolling.
+ *
+ * @param {HTMLTextAreaElement} textarea - the target `textarea` element
+ */
+const shrinkElasticTextArea = (textarea: HTMLTextAreaElement) => {
```
gerardo-rodriguez, line 28: Is the doubling up of the JSDoc `@param {HTMLTextAreaElement}` and TS `textarea: HTMLTextAreaElement` necessary? It feels like a bit much, but what do others think? 🤔

src/components/input/input.ts
calebeby: I can reproduce this issue in-browser if I paste in the first text, then select all and paste in the second text and then undo-redo. The single-line one does not make the text box resize smaller

src/components/input/input.stories.mdx
gerardo-rodriguez: @calebeby Oh, I was wondering why the `(textarea) => {}` had parathesis. Usually, we don't add parenthesis when there is only one argument.

src/components/input/input.stories.mdx
calebeby: Do we need to document this implementation detail? I think it does not affect users of this component.

I think it would be better if we only describe only what it does and how to use it, instead of how it works. If someone wants to know how it works, they can look at the source code / comments

src/components/input/input.ts
calebeby: Yeah I think we can remove it https://github.com/cloudfour/cloudfour.com-patterns/pull/716#discussion_r428228726

src/components/input/input.stories.mdx
Paul-Hebert: Oh, right. Good call!

src/components/input/input.stories.mdx
Paul-Hebert: I _think_ prettier changed this default recently... I don't remember where I read that though.

I think this had to do with it being easier to add more params if the parens are there?

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+      break;
+    }
+  }
+};
+
+/**
+ * Update Elastic TextArea
+ *
+ * When a textarea is updated, compares the new value to the previous value
+ * (from a dataset attribute) to determine whether to grow or shrink.
```
gerardo-rodriguez, line 51: Make sure to update this comment per https://github.com/cloudfour/cloudfour.com-patterns/pull/716/files#r428227091. 😉

src/components/input/input.stories.mdx
calebeby: What does it do that is different from the react useEffect (i.e. why do we import this from storybook, instead of react)?

src/components/input/input.stories.mdx
Paul-Hebert: https://prettier.io/docs/en/options.html#arrow-function-parentheses

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+  textarea.dataset.lastValue = newValue;
+};
+
+/**
+ * Create Elastic TextArea
+ *
+ * Given a textarea, adds dataset attributes for the existing value and the
+ * minimum number of rows, for use in the update function, which is added to
+ * an event listener. Returns an object containing the `destroy()` method to
+ * remove the event listener.
```
gerardo-rodriguez, line 83: Same here, make sure to update this comment per https://github.com/cloudfour/cloudfour.com-patterns/pull/716/files#r428227091. 🙂

src/components/input/input.ts
Paul-Hebert: I agree it looks like a bit much when reading the code.

That said I don't know a lot about how TS and JSDoc work. I would want both the type checking, and the param hint when using the function. Can we remove one of these and keep both of those features?

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @param {HTMLTextAreaElement} textarea - the target `textarea` element
+ * @returns {object} object containing a `destroy()` method
```
gerardo-rodriguez, line 86: @calebeby Is there a better way to document what is returned? What happens when we are returning an object that has multiple methods? It seems trying to describe every method/function that is returned would turn into a super-long sentence. 😬

src/components/input/input.ts
calebeby: @gerardo-rodriguez I set up testing with cypress here: https://github.com/cloudfour/cloudfour.com-patterns/pull/726

I don't think that Jest is the best fit for this case since JSDOM doesn't have a layout engine, so you won't really be able to test out this component. Currently working on adding tests for the textarea component, which I will push to that branch once this PR is merged

But yeah, good point, I will test with no minRows attribute

src/components/input/input.stories.mdx
calebeby: Yep prettier 2.0 has parens on all the arrow functions

https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev

calebeby: https://twitter.com/ManuelBieh/status/1181880524954050560/photo/1

src/components/input/input.ts
Paul-Hebert: In the past I've followed this strategy for defining a `param` option with multiple props. I wonder if we could do something similar for returns?

https://stackoverflow.com/a/6460748/7816145

Paul-Hebert: Though I guess the returned object is unnamed... 🤔

src/components/input/input.ts
```diff
@@ -0,0 +1,108 @@
...
+  // Fire an input event to set the initial size correctly
+  textarea.dispatchEvent(new Event('input'));
+
+  // Return an object with the destroy method
+  // so users can remove the event listener if needed
+  return {
+    destroy() {
+      textarea.removeEventListener('input', updateElasticTextArea);
+    },
+  };
```
gerardo-rodriguez, line 107: Perhaps nothing to be handled in this PR but I was curious what folks think about this:
```js
export const createElasticTextArea = () => {
  return {
    destroy() {
      // some code...
    }
  }
}
```
vs
```js
export const createElasticTextArea = () => {
  return {
    destroy: () => {
      // some code...
    }
  }
}
```
vs
```js
export const createElasticTextArea = () => {
  const destroy = () => {
    // some code
  }

  return {
    destroy
  }
}

```
I _personally_ prefer the second or third but curious how the rest of the team feels about it. 🙂

src/components/input/input.stories.mdx
gerardo-rodriguez: ¯\\\_(ツ)_/¯

src/components/input/input.ts
calebeby: If it was me, I would remove the entire `@returns` annotation. TS already knows what this returns. If you hover over the `=>`, it will show the signature, which includes the (inferred) return type.

For me, describing the return type in words does not help me understand the function.

I would add a description annotation to the `destroy` method, which would be helpful:

image

src/components/input/input.stories.mdx
gerardo-rodriguez: Cool, thanks! 🎓

src/components/input/input.stories.mdx
gerardo-rodriguez: @calebeby That's fair and a good point. 👍

src/components/input/input.ts
gerardo-rodriguez: > [I would add a description annotation to the destroy method, which would be helpful:](https://github.com/cloudfour/cloudfour.com-patterns/pull/716#discussion_r428257857)

This is one reason why I would consider the third option. It keeps the `return {}` more readable at a glance. 🙂

src/components/input/input.ts
gerardo-rodriguez: > I would add a description annotation to the destroy method, which would be helpful:

I agree! 😄

src/components/input/input.ts
calebeby: Yep! VSCode will include the type from the TS type annotation, and the description from the JSDoc

src/components/input/input.twig
```diff
@@ -13,6 +13,7 @@
     'size',
     'min',
     'max',
+    'rows',
```
calebeby, line 16: Does this need to forward data-min-rows as well?

src/components/input/input.stories.mdx
spaceninja: @calebeby because when I googled "how do I get my JS to work after a story renders in storybook" I ended up on that issue I linked in the comment which had this code sample. I don't actually understand React hooks (they came out after I finished my React training) so I'm open to being told we don't need the Storybook hook exposing this, but I'd need some help with what we _should_ use instead.

src/components/input/input.stories.mdx
spaceninja: Good catch. In an earlier iteration I was using the result to test the destroy method, but we don't need it now.

src/components/input/input.stories.mdx
spaceninja: Actually it does use the `rows` attribute — As part of the create function, we add the `data-min-rows` attribute, and assign it either the initial value of the `rows` attribute, or `2` if `rows` isn't set. So for the pattern library consumer, the part they care about is the `rows` attribute.

src/components/input/input.ts
spaceninja: That looks good — I'll try it out.

src/components/input/input.ts
spaceninja: Ah, good call. Yeah, I'll change that.

src/components/input/input.ts
spaceninja: Sounds good. I was also unhappy with having to repeat all the variables in grow and shrink, so this approach would solve that. I'll refactor.

src/components/input/input.stories.mdx
calebeby: No worries, I'm fine with leaving this as-is, I don't have a suggestion. I was just trying to understand how this is working

src/components/input/input.ts
spaceninja: @calebeby I'm a little unclear on the interaction between JSDoc and TS if you're not extracting types from JSDoc.

Normally JSDoc wants you to put the type there. We're not generating JSDoc documentation, so it kinda doesn't matter if we violate the rules, but it seems odd to me.

Is it common to leave the types out of the JSDoc blocks when you're using TS?

src/components/input/input.stories.mdx
calebeby: Got it :thumbsup:

src/components/input/demo/example.twig
spaceninja: We pass in parameters to the example file, and by default it passes those parameters down. The net effect is that I was setting `class="js-elastic-textarea"` on the Story, and it was applying the class to both the textarea and the button.

Adding `only` here says "only pass in the parameters I've specified" — in this case, only pass in the `label` parameter.

Tyler told me about this when I mentioned the problem. Details here: https://twig.symfony.com/doc/2.x/tags/include.html

src/components/input/input.scss
spaceninja: Oh that's a good question, actually. I was only thinking about the elastic textareas which will automatically get a min-height from `rows="2"`.

I'll restore the existing style here and just reset `height` for `.js-elastic-textarea`.

src/components/input/input.stories.mdx
spaceninja: You're correct with regards to the JS, but I'm actually about to add some elastic textarea-specific CSS, so they'll need this specific class. I'll add a note to that regard.

src/components/input/input.stories.mdx
spaceninja: Maybe I can rephrase the first sentence to something like "Add the `js-elastic-textarea` class to a `textarea` element (and follow the JS usage notes) to have it expand…"

src/components/input/input.stories.mdx
spaceninja: Yes, for our Story, we can do that (and I will). However, I'll leave the code sample alone, because I think most people will want to target all elements with that class, not just the first one found.

src/components/input/input.stories.mdx
spaceninja: > Could we get some cloud four branded content here?

Great idea, I'll do that.

> is it confusing being able to change this either via knob or by typing in the textarea?

Perhaps? I don't see a syntax to let me just _set_ a value without also making it a knob.

@tylersticka how do you feel about that?

src/components/input/input.twig
calebeby: Answer: no

https://github.com/cloudfour/cloudfour.com-patterns/pull/716#discussion_r428293239

src/components/input/input.ts
spaceninja: `minRows` will always be set to either the user-provided `rows` attribute or `2`.

If the user sets `rows` to 1, that's a valid value.

If the user sets `rows` to 0 or a negative value, the browser ignores it and the effective result is `2`.

(unfortunately, changing the `rows` value in the Story knob seems to break our JS, so this is difficult to test, but I just verified it)

src/components/input/input.ts
calebeby: Yep, it is common! https://github.com/microsoft/tsdoc#what-is-tsdoc

src/components/input/input.ts
spaceninja: No need to set a minimum or default value here, it's set during creation to either the value of `rows` or `2`.

src/components/input/input.ts
spaceninja: Heh. Tricky edge case!

The shrink method is clever enough to handle this — it'll shrink until it either hits the minimum rows value or starts scrolling, and then it'll grow until it stops scrolling.

But the grow method is dumb, and just adds rows as long as it's scrolling. It doesn't the same sanity check as shrink does.

I don't know how to fix it, though — grow would need to check whether the content is currently _shorter_ than the height. And I don't know how to check that. Does anyone have any ideas?

(Also, as Paul points out, this edge case is present in the original component as well, so if this is hairy to fix, perhaps this should be a new issue, rather than a part of this PR)

src/components/input/input.ts
spaceninja: This is one of those things where because I came to JS after ES6 was already a thing, I think the ES6 shorthand method looks simpler, but I don't feel strongly about it.

If we were picking a style for our JS guide, my personal preference would be 1, 3, 2. I don't think 2 adds anything over 3 besides more verbosity, unless I'm missing something. But I could see a case being made for 3 if you had a lot of stuff in the return object, so you don't clutter it up.

src/components/input/input.ts
calebeby: How about this:

Currently it works like this:
```
If the contents.length is longer then before, grow()
Else shrink()
```

How about we change it to:
```
If there is a scrollbar, grow()
Else shrink()
```

I think that would fix this edge case

src/components/input/input.stories.mdx
spaceninja: I reconsidered, and updated the documentation to differentiate between the classname needed for styling and the one for JS selection, and indicated that you could use any classname for selection.

src/components/input/input.scss
spaceninja: I actually moved the style to a `c-input--elastic` modifier to better follow our BEM rules. JS-specific classnames don't belong in CSS files, by definition.

src/components/input/input.ts
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @param textarea - the target `textarea` element
+ */
+const growElasticTextArea = (textarea: HTMLTextAreaElement) => {
+  const minRows = Number(textarea.dataset.minRows);
+  let rows = Number(textarea.getAttribute('rows')) || minRows;
+  let isScrolling = textarea.scrollHeight > textarea.clientHeight;
+
+  // Grow until we stop scrolling
+  while (isScrolling) {
```
calebeby, line 14: If someone forgets to have the `c-input--elastic` class, then this causes an infinite loop that keeps increasing `rows` with no effect on the scrollbar. This also happens if it is used somewhere where there is conflicting CSS that sets `max-height` or `height`

Can we prevent the infinite loop? Here are some possibilities as to how:

- Check that rows < 500 or 1000
- Check that the height of the element has increased as a result of the `rows` attribute increasing

Paul-Hebert: Could we check for the `c-input--elastic` class when initializing?

src/components/input/input.ts
Paul-Hebert: I agree with Scott. In general I prefer #1 but sometimes #3 is necessary if you need to have internal access to the method you're returning.

src/components/input/input.ts
calebeby: There is still the possibility that it could cause an infinite loop, like if that CSS did not load for some reason, or was accidently ordered with the wrong priority compared to other CSS that has fixed height

src/components/input/input.ts
spaceninja: Updated with option 3

src/components/input/input.ts
gerardo-rodriguez: > Could we check for the `c-input--elastic` class when initializing?

Wouldn't we be mixing styling concerns with business logic concerns if we did this? Or perhaps I'm misunderstanding. 🤔 

If the JS logic needs to depend on a CSS class, then it should be prefixed with `js-*`.

src/components/input/input.ts
gerardo-rodriguez: This comment has nothing actionable for this PR, so I'm going to resolve this conversation.

Thank you for your input! After reading, I can be convinced any are fine with a lean toward the third. But that is not a blocker to this PR. 😉

src/components/input/input.ts
spaceninja: @calebeby To be clear — you're just proposing a maximum rows value as a sanity check so the user won't cause an infinite loop. They'll still know something is wrong because the textarea will be pegged at the max value, right?

calebeby: The infinite loop happens when changing the `rows` attribute does not cause any visual change, so the scrollbar will never go away (i.e. the element has a fixed height or max-height). If there was a maximum number of rows, then if that infinite loop case occurred, then the user would see whatever the fixed height or max height is. They would not know that anything is wrong other than that the text area does not resize

src/components/input/input.stories.mdx
gerardo-rodriguez: > Just want to make a note, the createElasticTextArea function accepts an element, not a selector. 

How did the previous JS Class version work?

I would expect to pass a CSS selector to the `createElasticTextArea()` function. But I also don't recall how the previous version was working. 🤷‍♀️

gerardo-rodriguez: Or, wait...I'm getting myself all confused. Ignore my comment for now. I think you'll have it handled. LOL

gerardo-rodriguez:
@spaceninja I'm going to bow out of the PR review for now. @calebeby and @Paul-Hebert are doing an excellent job reviewing. 

If I am needed for another review later, though, ping me again and I can take a look then. Thanks! 😉

src/components/input/input.ts
spaceninja: Clever, and it totally works!

src/components/input/input.ts
spaceninja: Dope. I added a maxRows check to the grow method

spaceninja:
Okay! I've addressed all the feedback, and I'm personally very happy with the changes. Alltogether they dramatically simplified the code, increased readability, fixed a bug that was in the original version, and staved off at least one pretty easy to trigger infinite loop condition!

I have one outstanding question for @tylersticka about whether there's a way to pass parameters into the Story that are NOT knobs (eg, the ID and value). That's not a blocker, though.

I've re-requested reviews from y'all. Thanks for helping to create a better example for upcoming JS contributors!

src/components/input/input.stories.mdx
Paul-Hebert: > Perhaps? I don't see a syntax to let me just set a value without also making it a knob.

I think that you would just omit the `text()` function: `value: 'a\nb\nc\nd\ne'`

gerardo-rodriguez:
> Okay! I've addressed all the feedback, and I'm personally very happy with the changes. Altogether they dramatically simplified the code, increased readability, fixed a bug that was in the original version, and staved off at least one pretty easy to trigger infinite loop condition!

Yay, refactoring an initial effort leading to more simplified and readable code always feels great! Excellent work, @spaceninja! 😄 

I'm going to take another peek now at the most recent changes. 🙂

src/components/input/input.stories.mdx
tylersticka: @spaceninja You should be able to just set a value. Just pass it along without the `text(...)` function.

src/components/input/input.stories.mdx
```diff
@@ -1,8 +1,16 @@
...
+  text,
+  number,
+  boolean,
+  select,
+} from '@storybook/addon-knobs';
+import { useEffect } from '@storybook/client-api';
 import textDemo from './demo/example.twig';
 import selectDemo from './demo/select.twig';
 import './demo/styles.scss';
+import { createElasticTextArea } from './input.ts';
```
gerardo-rodriguez, line 13: I was expecting the filename to more closely be related to the module itself. Might we consider something like:
```js
import { createElasticTextArea } from './elastic-textarea.ts`;
```
🤔

src/components/input/input.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * Create Elastic TextArea
+ *
+ * Adds an event listener to a `textarea` elment that responds to input events
+ * by either increasing or decreasing the `rows` attribute based on whether the
+ * `textarea` is scrolling or not. Returns an object containing a `destroy()`
+ * method to remove the event listener.
+ *
+ * 1. Sanity check to prevent an infinite loop if the textarea has a fixed height
+ *    @see https://github.com/cloudfour/cloudfour.com-patterns/pull/716#discussion_r428349375
```
gerardo-rodriguez, line 10: Oh, have we introduced a new pattern for documenting our JS? I would normally expect this comment to be inline as opposed to being included up here. 🤔 

It seems this pattern follows our CSS guidelines for adding comments. Sorry if I missed that we were changing our JS documentation style to follow the CSS documentation style. 😬 

FWIW, I have a preference lean toward having these types of comments inline within the body so I don't have to keep scrolling up/down to read and understand what this comment is referring to. 🙂

src/components/input/input.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    if (isScrolling) {
+      grow();
+    } else {
+      shrink();
+    }
+  };
+
+  /** Remove the event listener from the textarea */
+  const destroy = () => textarea.removeEventListener('input', update);
```
gerardo-rodriguez, line 58: Might we be able to change the angle on this comment to answer the _why_ behind the `destroy()`? At the moment, it feels as though the comment isn't adding much because it is restating what the code already tells me. 

Why are we adding a `destroy()` function? Is it because we want to be responsible and considerate in allowing users of this API to manage event listeners as needed? 🤔

src/components/input/input.ts
```diff
@@ -0,0 +1,68 @@
...
+      if (isScrolling) {
+        grow();
+        break;
+      }
+    }
+  };
+
+  /** Decide whether to grow or shrink the textarea */
+  const update = () => {
+    isScrolling = checkScrolling();
```
gerardo-rodriguez, line 48: Feel free to ignore this suggestion if you disagree, but I wondered if we might rename this to something like:
```js
isScrolling = getIsScrolling();
```
At a glance, I gather the following from `getIsScrolling()` without needing to look at the source function:
- `get` => something will be returned
- `getIsScrolling` => I expect `IsScrolling` to be a `boolean`, therefore, `getIsScrolling` must return a `boolean` letting me know if the textarea element is scrolling

Right now, `checkScrolling()` doesn't really help me understand what `checkScrolling()` does. It checks scrolling and...removes event listeners? Calls other functions? Sets the BG color to a different color? It's not very clear to me what `checkScrolling()` might do. 🙂 

Again, feel free to ignore this suggestion if you strongly disagree. I won't be offended. 😉

src/components/input/input.ts
```diff
@@ -0,0 +1,68 @@
...
+  /** Remove the event listener from the textarea */
+  const destroy = () => textarea.removeEventListener('input', update);
+
+  // 1. Add an event listener to run the update function after input events
+  textarea.addEventListener('input', update);
+
+  // 2. Fire an input event to set the initial size correctly
+  textarea.dispatchEvent(new Event('input'));
+
+  // 3. Return an object with the destroy method to remove the event listener
```
gerardo-rodriguez, line 66: The `1.`, `2.`, `3.` confuse me. Are they needed? What do they mean? 🤔

src/components/input/input.ts
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  /** Remove the event listener from the textarea */
+  const destroy = () => textarea.removeEventListener('input', update);
+
+  // 1. Add an event listener to run the update function after input events
+  textarea.addEventListener('input', update);
+
+  // 2. Fire an input event to set the initial size correctly
+  textarea.dispatchEvent(new Event('input'));
```
gerardo-rodriguez, line 64: > Can we call updateElasticTextArea directly instead of firing an event?
> There may be other input listeners on this textarea and they don't need to be called

This suggestion from @calebeby is still outstanding. 🙂

src/components/input/input.ts
gerardo-rodriguez, line 66: Similar to [this suggestion](https://github.com/cloudfour/cloudfour.com-patterns/pull/716/files#r428779629), I can see from the code that we are returning an object with the destroy method. Duplicating what the code already informs me in the comment doesn't add too much value, I feel. 

Maybe something like:
```js
// Return a public API for consumers of this component
```
Or something along those lines. 🙂

src/components/input/input.ts
```diff
@@ -0,0 +1,68 @@
...
+    } else {
+      shrink();
+    }
+  };
+
+  /** Remove the event listener from the textarea */
+  const destroy = () => textarea.removeEventListener('input', update);
+
+  // 1. Add an event listener to run the update function after input events
+  textarea.addEventListener('input', update);
```
gerardo-rodriguez, line 61: Same here, what does this comment provide that I can't already gather from the line of code itself? 🤔 

If we try to answer "_Why_ are we adding an `'input'` event listener that runs an `update` function callback?", the answer might be, "To initialize the textarea component."
```js
// Initiailize the textarea component
textarea.addEventListener('input', update);
```
This is only one idea, maybe there are better ones. 🙂

gerardo-rodriguez:
Great work, @spaceninja! I left some minor feedback and suggestions. I _love_ how this effort is turning out! 😄

src/components/input/input.stories.mdx
spaceninja: Interesting question!

I was operating on the assumption that "Input" is the component, and this is the component-specific JS, one function of which is `createElasticTextArea`.

If we added another feature to the Input component, would you expect it to live in a separate file or the same?

src/components/input/input.ts
spaceninja: That's fine — like you said, I'm used to this pattern from CSS-land when a comment is too long to go inline. I'll move it.

src/components/input/input.ts
spaceninja: How about: `/* Allow users to remove event listener from the textarea */`?

src/components/input/input.ts
spaceninja: Good call, I'll update.

src/components/input/input.ts
spaceninja: Oops, those were helping me figure out what was actually _happening_ as opposed to defined methods and variables. I'll remove the numbers.

src/components/input/input.ts
spaceninja: Whoops! Sorry I tried to address that earlier and it didn't work because I still needed to pass the event. But now that I've refactored, I should be able to make this change.

src/components/input/input.ts
spaceninja: that makes sense, I'll update

src/components/input/input.ts
spaceninja: Good call

src/components/input/input.stories.mdx
spaceninja: Ah, beauty! Thanks Tyler!

src/components/input/input.stories.mdx
Paul-Hebert: I'd prefer these to be split up into separate files. That should make it easier to only import the functionality you actually need on your page.

Side question: would this work on a naked `<textarea>` or does it rely on the component CSS?

gerardo-rodriguez: Excellent question, @spaceninja. 🙂 

> If we added another feature to the Input component, would you expect it to live in a separate file or the same?

I would imagine each module corresponds to one feature. I would like to avoid massive TS files if possible.

I'm not sure how this would be packaged up, though, for consumer use. Would there be one `input/index.ts` file that re-exports the `elastic-textarea` module? Something like:
- `input/elastic-textarea.ts`
- `input/another-input-feature.ts`
- `input/the-perfect-select.ts`
- `input/index.ts`

Or is it something like?
- `elastic-textarea/index.ts`
- `another-input-feature/index.ts`
- `the-perfect-select/index.ts`

I suppose this is more an architectural question. I'm curious what the thoughts of @Paul-Hebert and @calebeby are. I'm also curious what open-source projects might be doing. 🔍

spaceninja: That's fine, knowing we prefer smaller, feature-specific files makes sense.

To answer your other questions, there's no need for an index file (I dislike index files anyway because it makes it difficult to tell which file you're editing when you have multiple `index.js` files open)

Any `*.ts` files in the component directory will be added to the compiled package library JS file, and we'll just need to import each file as needed into the stories that need them (which I think is fine)

Assuming @Paul-Hebert and @calebeby have no objections, I'll rename this file.

TAKEAWAY: If a component has multiple JS features, each should live in a separate, clearly-named `.ts` file.

spaceninja: > Side question: would this work on a naked <textarea> or does it rely on the component CSS?

It relies on the component CSS to set `height:auto`, which I've added to the documentation.

And we added a `maxRows` last night to avoid triggering an infinite loop if the users forgets that class or if the textarea has a fixed height for any reason.

Paul-Hebert: I'd also prefer to leave out the index file unless strictly necessary. I can imagine a really complex feature having its own directory with multiple modules and an index file, but for this example I'd rather have `input/elastic-textarea.ts`

spaceninja: File renamed! I'm putting this back up for review now.

spaceninja:
Okay, y'all, I've addressed the latest feedback and I think this is done-done now! Can I get some approvals? :D

PR #726, opened 2020-05-20
Setup testing

### LINKED ISSUE #701: Add Testing
Pulling the testing-related tasks out of #683:

- Write tests for existing JS files in [Jest](https://jestjs.io/) and/or [Cypress](https://www.cypress.io/) with [testing-library](https://testing-library.com/)
- set up these tools as necessary

calebeby:
## Overview

Closes #701 

- Adds Cypress for component testing (experimental feature) in order to test individual patterns
- Uses @testing-library/cypress to encourage better testing practices and more accessible components

## Why not Jest
Because JSDOM does not emulate the visual layout part of the browser, which means that any tests which depend on actual layout will not be able to be tested.

## Testing

1. `npm i`
2. `npm run test`/`npm run test:open`


spaceninja:
#716 is merged! — I'll check this out now…

src/components/input/elastic-textarea.ts
```diff
@@ -12,6 +12,7 @@ export const createElasticTextArea = (textarea: HTMLTextAreaElement) => {
   const maxRows = 500; // Used to prevent infinite loop if textarea has fixed height
   const minRows = Number(textarea.getAttribute('rows')) || 2;
   let rows = Number(textarea.getAttribute('rows')) || minRows;
+  textarea.setAttribute('rows', String(rows));
```
calebeby, line 15: @spaceninja I wanted to flag this change, one of the tests fails without it (the test makes sure that when you initialize the JS, if there is no rows attribute, it sets it to 2). I'm fine with a different solution to that if you have one

spaceninja: 👍 I think that's fine — there's no difference for the user (because the update function runs a second later and will set the rows value), so this is just a sanity check for the tests.

calebeby: This does make a difference for the user. When the update function runs, in this case it does not set the rows value, if it did the test would pass without this addition.

`update` -> `isScrolling` is false -> `shrink()` -> `rows < minRows` so while loop never runs

spaceninja: I see what you're saying that the initial update function doesn't add a rows value if one isn't there and the contents don't exceed the min value. But I don't understand how that affects the user. Will they see any changes in behavior as a result of the rows value being added during the create function?

src/components/input/elastic-textarea.cypress.ts
```diff
@@ -0,0 +1,67 @@
...
+  it('Resizes correctly with no rows attribute', () => {
+    render(
+      textInputHTML({
+        class: 'c-input--elastic js-elastic-textarea',
+        type: 'textarea',
+      })
+    );
+    const textarea = cy.get('.js-elastic-textarea');
+
+    textarea.invoke('css', 'max-width', '500px');
```
spaceninja, line 16: What's the `max-width` for?

src/components/input/elastic-textarea.ts
calebeby: Yes. Since the default `minRows` value is `2`, with this addition, now when the JS is initialized it will update the textarea to 2 lines tall. Before this change, it would leave it at just 1 line

src/components/input/elastic-textarea.cypress.ts
spaceninja: Oh wait, I see — it's to make a predictable width so you know when your sample text will wrap?

src/components/input/elastic-textarea.ts
spaceninja: I'm unable to reproduce that behavior, but what you're describing is fine — the textarea will be initialized with the `minRows` value if no `rows` attribute is provided.

src/components/input/elastic-textarea.cypress.ts
calebeby: Yes. Before, it was full-screen-width (which is known and can be configured in cypress) but that would mean that I would need to put more words in my sentence in order for it to wrap 😂

calebeby:
Sorry @spaceninja there was another conflict cuz of other PR's getting merged

PR #727, opened 2020-05-20
Disable renovating node version

calebeby:
Tries to disable PR's like this: https://github.com/cloudfour/cloudfour.com-patterns/pull/690

I think this is the config we want: https://github.com/renovatebot/config-help/issues/81#issuecomment-414104278

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,8 +2,10 @@
   "extends": ["group:recommended", "group:monorepos"],
   "reviewers": ["team:dev"],
   "rangeStrategy": "pin",
-  "node": {
-    "supportPolicy": ["active"],
-    "rangeStrategy": "bump"
-  }
+  "packageRules": [
+    {
+      "packageNames": ["node"],
+      "enabled": false
+    }
+  ]
 }


renovate.json
```diff
@@ -2,8 +2,10 @@
...
   "rangeStrategy": "pin",
-  "node": {
-    "supportPolicy": ["active"],
-    "rangeStrategy": "bump"
-  }
+  "packageRules": [
+    {
+      "packageNames": ["node"],
+      "enabled": false
+    }
```
Paul-Hebert, line 9: So this is saying:

1. Find any package named "node"
2. Apply the following special rules to it:
3. The only special rule is to entirely disable renovate for it.

Is that right?

calebeby: Yes that is my understanding

calebeby:
No we don't have a way to test this until we merge it.

Here was my last shot at this: https://github.com/cloudfour/cloudfour.com-patterns/pull/689

_narrator: it did not work_

Paul-Hebert:
Cool, well the code looks good to me based on the linked issue and Renovate docs ¯\\\_(ツ)_/¯

I assume after we merge we just wanna keep an eye on Renovate to make sure it keeps acting correctly?

calebeby:
Yeah, should I merge or should I wait for @spaceninja to review?

calebeby:
After this is merged, I am expecting/hoping that this will auto-close: https://github.com/cloudfour/cloudfour.com-patterns/pull/690

Paul-Hebert:
Your call. I'm tempted to wait for @spaceninja 's review, but worst case this breaks renovate and not our actual codebase, so if you'd prefer to merge now that's fine with me

Issue #710, opened 2020-05-18
Style critical Gutenberg block markup


tylersticka:
The new version of our WordPress theme will support [WordPress's Gutenberg editing experience](https://wordpress.org/gutenberg/).

We've already added markup examples for these blocks to Storybook (see #590). But some of these would benefit from some additional styles.

## How should we do this?

- [x] 1. Create a new `wordpress` directory in `src/third-party` or `src/vendor` (depending on the outcome of #709).

- [x] 1. Create one or more CSS files within that directory that style the WordPress Gutenberg blocks using the classes they provide.
    - If we have an equivalent pattern, we should use [Sass mixins](https://sass-lang.com/documentation/at-rules/mixin) to extend that pattern to the WordPress equivalent. For example, we could create a mixin that lets us apply the styles for `c-button` to `wp-block-button__link`.
    - If there is no equivalent pattern, we can author original styles.

## Why not modify the markup instead?

Since our patterns already include classes and markup structures for some Gutenberg patterns, it's reasonable to ask if we can customize their output. But I'd like to avoid that strategy for a few reasons:

- We tried [modifying markup of shortcodes](https://wordpress.org/support/topic/customizing-the-html-output-by-the-shortcode/) before, and those modifications ended up causing regressions when new features were added later on.
- Our incentive at the time for modifying markup was that existing markup did not use semantic HTML5 elements. This is no longer the case.
- Having this customization live mostly in CSS land gives more of the team an opportunity to contribute to their maintenance than if they were in PHP and React.
- Repeating CSS via mixins [has very low performance impact](https://csswizardry.com/2016/02/mixins-better-for-performance/) and is easier than ever [now that we're using Sass again](https://cloudfour.com/thinks/feeling-sassy-again/).

## Shouldn't this be in our WordPress theme?

A solid argument could be made that we should be exposing necessary mixins via the npm package and applying those as needed within the WordPress theme instead of doing that work here.

But our pattern work is much farther along than our theme work, and we already have markup examples, so it seems silly to needlessly complicate our short-term process.

## Priorities

There are currently 19 Gutenberg blocks we've documented [in our Storybook](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/third-party-gutenberg--core-audio). Of those, I think we should prioritize the following:

- [x] import `@wordpress/block-library` styles (#778)
- [x] core/button (#729)
- [x] core/table (#743)
- [x] core/file (#773)
- [x] core/image (#754)
[...]

PR #729, opened 2020-05-21
Chore: style gutenberg button same as c-button 

derekshirk:
## Overview

See #710 

Creates a button mixin and applies our `c-button` styles to buttons output via Gutenberg button blocks. Selectors account for filled and outlined buttons as well as maintaining our desired colors and border radii.

I failed to arrive a solution for styling `:focus` states due to the lack of buttons and the following HTML output: 

```html
<div> // The (button) container
  <a> // the (button) link
</div>
```

## Screenshots

I've added an outlined markup example to the vendor → WordPress → Gutenberg → core/button example: 

Screen Shot 2020-05-22 at 10 31 15 AM

Testing output in a local version of our site:

Screen Shot 2020-05-22 at 11 18 57 AM

Example of the available options within the button block editor: 

Screen Shot 2020-05-21 at 5 04 22 PM

## Testing

1. View [deploy preview](https://deploy-preview-729--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-gutenberg--core-button)
1. Test `:hover`
1. Test `:focus`
1. Bonus testing: 
  - Set up a local version of cloudfour-dot-com@v-next
  - Enable Gutenberg editor / uncheck setting in profile to disable visual editor
  - Test output of buttons via block editor

## Notes

- See: https://github.com/cloudfour/cloudfour.com-wp/issues/491


src/vendor/wordpress/_wordpress.scss
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Gutenberg block: Button
+ * Applies our pattern library styles to buttons generated via Gutenberg blocks.
+ */
+
+.wp-block-button {
+  @include button.default;
+}
+
+.wp-block-button.is-style-fill {
```
spaceninja, line 13: Can you nest `&.is-style-fill` and `&.is-style-outline` inside `.wp-block-button`?

derekshirk: Good suggestion. Thanks @spaceninja! 

> Overall this seems reasonable to me. I'll defer to @tylersticka, though.

During my 1:1 with Tyler earlier this morning, he mentioned he had some general high-level feedback about the overall approach here, so I'm going to wait and see how his feedback affects this suggestion before pushing any updates.

src/vendor/wordpress/_wordpress.scss
```diff
@@ -0,0 +1,24 @@
...
+ * Gutenberg block: Button
+ * Applies our pattern library styles to buttons generated via Gutenberg blocks.
+ *
+ * 1. The block editor allows editors to adjust button border radius. This style
+ *    is applied via an inline style on the `wp-block-button__link` element. In
+ *    order to prevent unwanted border-radius we apply an `!important` tag.
+ */
+
+.wp-block-button {
+  display: inline-block;
```
tylersticka, line 14: I'm assuming this is so multiple buttons will not break to new lines.

Did you consider using `display: flex` on `wp-block-buttons` with padding for nicer gaps between multiple buttons?

src/vendor/wordpress/_wordpress.scss
```diff
@@ -0,0 +1,24 @@
+@use "../../mixins/button";
+@use "../../design-tokens/sizes.yml";
+
+/**
+ * Gutenberg block: Button
+ * Applies our pattern library styles to buttons generated via Gutenberg blocks.
+ *
+ * 1. The block editor allows editors to adjust button border radius. This style
+ *    is applied via an inline style on the `wp-block-button__link` element. In
+ *    order to prevent unwanted border-radius we apply an `!important` tag.
```
tylersticka, line 10: Could you move this comment so that it is above the block of CSS where the number exists rather than above its parent?

src/vendor/wordpress/_wordpress.scss
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+.wp-block-button {
+  display: inline-block;
+
+  .wp-block-button__link {
+    @include button.default;
+    border-radius: sizes.$radius-medium !important; /* 1 */
```
tylersticka, line 18: If this style is omitted, what happens? Is it always overridden or is it only overridden if the author chooses a different radius?

derekshirk: Great question @tylersticka, 

If the author does not customize the border-radius section in the block editor this style is not needed. 

If the author does does modify the border-radius settings, this style is needed to prevent pill shaped or other unwanted border radius values. 

Does that answer your question?

tylersticka: It does.

In that case, I'm not sure I agree that this style is necessary. It feels odd to me that we would purposely break CMS functionality from our stylesheet instead of disabling the CMS feature somehow.

derekshirk: Excellent point. 

I had considered this as well. However, my current understanding is disabling the border-radius option is either not possible, or rather difficult. 

https://github.com/WordPress/gutenberg/issues/19796

tylersticka: I am comfortable with the worse-case scenario of "the Cloud Four team may optionally select an off-brand border radius for buttons."

src/vendor/wordpress/_wordpress.scss
derekshirk: > I'm assuming this is so multiple buttons will not break to new lines.

Correct. My intent was to match the presentation and layout of our existing `c-button` pattern as much as possible. 

> Did you consider using `display: flex `on `wp-block-buttons` with padding for nicer gaps between multiple buttons?

I had considered using `display: flex` with a gap, but did not consider using padding. 🤔 

If you feel like adding `display: flex` with padding is the preferred approach here, I need to think though how the padding would work.

tylersticka: Sorry, I should have said "whitespace" (did not mean to suggest the solution actually must use `padding`).

src/vendor/wordpress/_wordpress.scss
```diff
@@ -0,0 +1,27 @@
...
+
+.wp-block-buttons {
+  display: flex;
+  flex-wrap: wrap;
+
+  & * {
+    margin: sizes.$edge_large / -4;
+    padding: sizes.$edge_large;
+  }
+}
```
tylersticka, line 17: The gap seems very, very small and there is some extra space around the outer edge.

Here's what I would do: https://codepen.io/tylersticka/pen/f51fd10bbaa45abc9238ee833392db71?editors=1100

src/vendor/wordpress/_wordpress.scss
```diff
@@ -0,0 +1,55 @@
+@use "../../mixins/button";
+@use "../../design-tokens/sizes.yml";
+
+/**
+ * Gutenberg block: Button
+ * Applies our pattern library styles to buttons generated via Gutenberg blocks.
+ */
+
+$rhythm-space-generous: sizes.$rhythm_space_generous;
+$list-inline-gap: sizes.$list_inline_gap;
```
tylersticka, line 10: Are these extra variables necessary here? I had them in my CodePen because I didn't have the actual design tokens to reference.

Also, the design tokens should be referenced with dashes rather than underscores.

derekshirk: 👋 @tylersticka 

> Are these extra variables necessary here?

I don't know that they are _necessary_, but I thought they were a convenience, since the same tokens are used on multiple lines in the same file and I thought it would offer developer convenience to be able to change values in a single location. 

Perhaps I should work to gain a better understanding of when it is appropriate to to create custom variables from tokens and when not to. 

For example, I observed this in `table.scss`:

https://github.com/cloudfour/cloudfour.com-patterns/blob/31e00c2cce7b3b6a42d9884dac0256a6a772dc18/src/components/table/table.scss#L6-L8

...and this in `calendar-date.scss`:

https://github.com/cloudfour/cloudfour.com-patterns/blob/31e00c2cce7b3b6a42d9884dac0256a6a772dc18/src/components/calendar-date/calendar-date.scss#L7-L9

tylersticka: I think what's giving me pause is that the variables are the same name in both places. In the examples you provided, there's some abstraction happening... the `$table-` variables being a good example.

If these variables were named `$wp-button-gap: sizes.$list-inline-gap`, I think the intent would be a lot more clear.

PR #748, opened 2020-05-27
Feature/add sky nav behavior

spaceninja:
## Overview

Please include a summary of the change and which issue is fixed.
Please also include relevant motivation and context.
List any dependencies that are required for this change.

## Screenshots

Visual changes should include a screenshot.

## Testing

1. instructions for reviewers
1. to test your changes

---

- Fixes # (issue)

/CC @person


spaceninja:
Note: This PR was originally intended to handle the animation as well as the functionality, but since it's been sitting idle for nearly a year, I decided to split the animation out into a new ticket #1137 so we can get this over the finish line finally.

src/components/sky-nav/sky-nav.scss
```diff
@@ -208,11 +208,15 @@ $_max-width: #{unit.strip(layout.$max-width-spread) +
 /**
  * The element that toggles the menu visibility at narrow widths.
  *
- * 1. This prevents us having to set and then override padding on the masthead
+ * 1. Vertical positioning is handled by JS because it's easier to grab the
+ *    height of the header bar from there.
```
tylersticka, line 212: I'm gonna poke at this a bit, because there's something about overtly setting a specific property via JS that feels like it's overstepping the separation of concerns between CSS and JavaScript a tad?

Some alternative ideas to consider:

1. Could we save `header.clientHeight` to a custom property, and use that and the existing `$control-height` token in a `calc` function to position the element?
2. Could we set the nav bar height to a specific value in CSS so that we know exactly how to position the button? This would likely mean some new restrictions in terms of nav item wrapping but I think it's fair to say "this is optimized for X number of rows maximum" if we need to.

tylersticka (commented):
Great start! One piece of critical feedback. Happy to discuss with voices sometime this afternoon if that's helpful.

src/components/sky-nav/sky-nav.scss
spaceninja: Now that I'm looking at this again, I don't see why we can't set a fixed height on the navbar for small screens, which would allow calculating the offset using CSS. Not sure why I didn't see that to begin with, but I'll give that a shot now.

spaceninja: Updated!

tylersticka:
Adding @Paul-Hebert as a reviewer because I may not be able to prioritize reviewing this as quickly as I'd like to. Feel free to proceed with his approval.

src/components/sky-nav/sky-nav.cypress.ts
```diff
@@ -0,0 +1,35 @@
+import { render } from '../../../cypress/cypress-render-component';
+import skyNavMarkup from './sky-nav.twig';
+import { createSkyNav } from './sky-nav';
+import menu from './demo/menu.json';
+
+describe('Sky Nav', () => {
+  it('can be opened on small screens', () => {
+    render(skyNavMarkup({ includeMainDemo: true, menu }));
+    cy.viewport('iphone-6');
```
Paul-Hebert, line 9: This is neat!

src/components/sky-nav/sky-nav.ts
```diff
@@ -0,0 +1,65 @@
+import { m } from '../../design-tokens/generated/breakpoint.js';
```
Paul-Hebert, line 1: Nit: This import name isn't particularly descriptive. (I saw it below and had to scroll back up.)

I wonder if we could do something like this?

```js
import { m as mediumBreakpoint } from  '../../design-tokens/generated/breakpoint.js'
```

src/components/sky-nav/sky-nav.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Adds an event listener to the toggle button of a Sky Nav component for click
+ * events, which runs the toggle command to show or hide the menu with animation.
+ * Responds to breakpoint changes to show or hide the nav for large and small
+ * screens. Returns an object containing a `destroy()` method to remove the
+ * event listener.
+ *
+ * @param navButton - the toggle button for the navigation
+ */
+export const createSkyNav = (navButton: HTMLButtonElement) => {
```
Paul-Hebert, line 14: Non-blocking nit: The verb `create` makes it seem like this should add new markup instead  of just adding event listeners. I wonder if `init` would make more sense? Though create makes sense in context with destroy... I don't have super strong feelings on this but figured I'd mention it.

src/components/sky-nav/sky-nav.ts
```diff
@@ -0,0 +1,65 @@
...
+   * Update Menu Layout
+   * Sets visibility of menu & navButton for small vs large screen layouts.
+   */
+  const update = (isLargeScreen: boolean) => {
+    if (isLargeScreen) {
+      navButton.removeAttribute('aria-expanded');
+      menu.removeAttribute('hidden');
+    } else {
+      navButton.setAttribute('aria-expanded', 'false');
+      menu.hidden = true;
```
Paul-Hebert, line 28: I didn't realize you could access `hidden` directly. Neat!

Paul-Hebert (dismissed):
Looks good @spaceninja ! I left some nits and comments inline but none are blocking.

Paul-Hebert (approved):
Looks good! Nice job!

PR #751, opened 2020-05-28
Change how twig import paths work

### LINKED ISSUE #740: Reevaluate Twig include path strategies
We use [Twig](https://twig.symfony.com/) templates for this project because they're well supported across various platforms, and specifically in [Timber](https://www.upstatement.com/timber/), which we've enjoyed using on several WordPress projects.

Twig templates can include _other_ Twig templates in a handful of ways:

```twig
{# Output a template right here #}
{% include 'template.twig' %}

{# As a function #}
{{include('template.twig')}}
{# or #}
{% set content = include('template.twig') %}
{{content}}

{# Override layout blocks within #}
{% embed 'template.twig' %}
  {% block title %}
    <h2>Hello world!</h2>
  {% endblock %}
{% endembed %}
```

What all of these techniques have in common is that they require _a path to a template_. If Twig can't find `template.twig`, it can't render its contents.

And what _this_ project requires are paths that will work here not just in our Storybook examples, but _elsewhere_... specifically within our WordPress theme.

## What we do now
Our strategy so far has been to use relative paths.

So given this simplified version of our pattern repo's structure:

```
src
├── components
│   └── button
│       └── button.twig
├── objects
│   └── container
│       └── container.twig
└── prototypes
    └── example
        └── hello-world.twig
```

The paths  `hello-world.twig` would use to access the other templates look like this:

```twig
{% embed '../../objects/container/container.twig' %}
  {% block content %}
    {% include '../../components/button/button.twig' %}
  {% endblock %}
{% endembed %}
```

Because the paths are relative, they should work as long as the templates don't change location relative to each other. So in a WordPress project structured like this:

```
wp-content
└── themes
    └── example-theme
        ├── node_modules
        │   └── @cloudfour
        │       └── patterns
        └── templates
            └── page.twig
```

`page.twig` could reference `hello-world.twig` without issue:

```twig
{% include '../node_modules/@cloudfour/patterns/src/prototypes/example/hello-world.twig' %}
```

### Benefits of this approach
- It requires no configuration (it is the default behavior), which means it works consistently across platforms.
- It's easy to understand.
- In JavaScript, only templates that are actually referenced are loaded.

### Issues with this approach
Unfortunately, I've noticed some issues in actual use I believe we'll have to address soon.

#### Long paths
I don't know about you, but seeing paths like `../node_modules/@cloudfour/patterns/src/prototypes/example/hello-world.twig` makes me kind of sad.

#### Harder to move
Because the paths are relative, they'll change slightly depending on the depth of the file you're working in. If you decide to move a template from `templates/a.twig` to `templates/a/default.twig`, you will have to add another `../` segment to each and every include path therein.

They'll also change depending on the platform, which is likely to have a different directory structure. We've talked about leveraging future Storybook source improvements to expose more code samples, but those will be of limited utility if you the paths won't work in your project.

#### Lack of dynamic paths in JavaScript
When our projects have an absolute directory of includes or partials, it lets us write components that can extend different paths based on a property.

Consider the social media links in our website footer:

Screen Shot 2020-05-22 at 4 24 44 PM

Ideally, those links would be defined in our CMS. So our template for outputting them might look like this:

```twig
<ul>
  {% for item in menu.items %}
    <a href="{{item.link}}">
      {% include ['../path/to/' ~ item.slug ~ '.svg.twig', '../path/to/fallback.svg.twig'] with { class: 'c-icon' } only %}
      <span class="u-hidden-visually">{{item.name}}</span>
    </a>
  {% endfor %}
</ul>
```

This would work fine in PHP: If the item's slug matches an icon key, it will display, otherwise the fallback will display.

But in client-side JavaScript the slug will _never_ match. That's because relative include paths are fetched prior to compilation, _not_ prior to render. Since Twig can't know the path until the template is rendered with data, it won't have fetched that file.

#### Misc. "inline template" errors in JavaScript
While writing the complex Sky Nav component, I encountered issues where attempting to keep my code DRY with separate templates resulted in seemingly random errors complaining that certain template types were not allowed to be inline. The issues I found related to this pointed either to a missing configuration option (which our project already sets) and to relative paths.

## Alternative 1: Use a different Twig
Our project currently uses [Twig.js](https://github.com/twigjs/twig.js/wiki) and [twig-loader](https://github.com/zimmo-be/twig-loader). There is a competing Twig implementation called [Twing](https://www.npmjs.com/package/twing) and a [twing-loader](https://github.com/NightlyCommit/twing-loader#readme). Twing is more complex to setup, but boasts greater parity with the PHP version of Twig. Its loader also includes a pre-compilation strategy that may solve some of the issues we've been seeing.

## Alternative 2: Register templates ahead of time
It's possible in Twig.js to [register templates at runtime](https://github.com/zimmo-be/twig-loader#dynamic-templates-and-registering-at-runtime), letting the engine know up-front what templates are available.

### Benefits of this approach
- Dynamic include paths can be evaluated as long as they match a template that was registered at runtime.
- Less chance of an "inline template" error since templates are evaluated ahead of time (rather than being injected at the time of request).
- Ability to simplify paths so they can be more concise.

### Issues with this approach
- Unlike relative paths, this approach will require configuration.
- As near as I can tell, partial paths that are preregistered will no longer be relative. So if you register a path at `components/button/button.twig`, the path will be the same no matter what folder depth you are in.
- Because this solution requires configuration, it will also require additional steps to integrate those templates in other projects. We wouldn't want to use a path format that we couldn't recreate in the PHP version of Twig, for example.

## Next steps
- Explore Alternative 1 to see if it can solve the issues we're seeing.
	- If so, proceed with a PR.
	- If not, move on to exploring Alternative 2, specifically checking if relative paths are actually supported or not.
		- If  relative paths_are_ actually supported with this strategy, proceed straight to PR.
		- If relative paths are _not_ supported, research pathing strategies that will work across JS and PHP implementations and proceed with PRs for both projects.

calebeby:
## Overview

Closes #740

- Switches to [twing](https://github.com/NightlyCommit/twing) and [twing-loader](https://github.com/NightlyCommit/twing-loader)
- Pre-loads all twig paths
- Since twing more strictly follows twig-php's syntax, I changed some deprecated syntax into syntax it accepted 

Now `src/components/button/button.twig` can be imported as `@cloudfour/patterns/button/button.twig`. I can change this if we want to something shorter, maybe like `@patterns/button/button.twig` if that is better.

## Testing

- Visit the [preview](https://deploy-preview-751--cloudfour-patterns.netlify.app/), check that everything still works since all the import paths changed
- Test that importing dynamic paths (i.e. when looping through an array) works. @tylersticka I haven't actually tested this yet. Which file would be easiest to test this with?


calebeby:
@tylersticka one thing I noticed is that in storybook the "docs" tab is selected, when switching between different components, it takes longer for the preview iframes to appear. I assume this is because all of the twig templates are being loaded for each one in an iframe, but I don't know of any alternative 🤔 

For some reason when "canvas" is selected it loads right away, I'm not sure why that is

tylersticka:
> I can change this if we want to something shorter, maybe like `@patterns/button/button.twig` if that is better.

Could we remove the `patterns` portion of the paths?

So for example `@cloudfour/components/button/button.twig` or `@cloudfour/objects/rhythm/rhythm.twig`?

tylersticka:
> For some reason when "canvas" is selected it loads right away, I'm not sure why that is

My understanding (which may be incorrect) is that in Canvas view, you have a single instance of a canvas with a single iframe that gets updated when you switch stories. Whereas in Docs, it's not just multiple iframes, it's also multiple canvases that are initializing all at once with their own options.

This was already pretty slow before. I've been meaning to create an issue to provide our own static HTML inline provider logic so we can embed some _simple_ patterns as inline elements rather than requiring an iframe and canvas, which should help load times.

tylersticka:
> * Test that importing dynamic paths (i.e. when looping through an array) works. @tylersticka I haven't actually tested this yet. Which file would be easiest to test this with?

I just pushed a branch I've been working on with an example, an array of menu items with icon names. The icon is displayed by concatenating a path based on that name: https://github.com/cloudfour/cloudfour.com-patterns/blob/c67caa6a86c8f6f6c891ee63ce6fc20de2cc9009/src/components/ground-nav/ground-nav.twig#L52-L56

tylersticka:
I haven't dived into reviewing the code yet but I'm very happy with the functionality. I just tried three things that didn't work before:

- A dynamic path.
- An array of paths (first one to resolve is displayed).
- The same template imported and displayed _and_ included via a Twig template within a single `.stories.mdx` file (would often error out after navigation occurred before).

The only question I have so far: What is the expected behavior for new `.twig` files? I can't tell if they are detected automatically or if a restart of Storybook is required first... it may be that my testing is so sporadic that I'm _perceiving_ inconsistent behavior, or it may be that a restart may sometimes be required.

src/objects/embed/embed.twig
```diff
@@ -5,7 +5,7 @@
 <div class="o-embed{% if class %} {{class}}{% endif %}"
   {% if style %}
     style="
-      {%- for rule in style|split(';') if rule|length -%}
+      {%- for rule in style|split(';')|filter(rule => rule|length) -%}
```
tylersticka, line 8: Yay, the ES6-style Twig filters work now! (They didn't in Twig.js)

.storybook/main.js
```diff
@@ -1,5 +1,9 @@
 const { resolve } = require('path');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const {
+  twingEnvironmentPlugin,
+  twingOptions,
+} = require('../twing-environment-webpack-plugin');
```
tylersticka, line 6: Since there are two files prefixed with `twing-`, do you think we should create a directory for these files?

calebeby:
I am 90% sure that after adding a new twig file you will need to restart webpack. I don't see a trivial way to solve this, I guess one option would be in the webpack config to watch for files being created using a separate watcher from webpack, and re-build the virtual module after that happens.

It might also be possible to hook into webpack's watcher and use that to detect new twig files, but I am not sure.

Do you want me to dig into that?

tylersticka:
> It might also be possible to hook into webpack's watcher and use that to detect new twig files, but I am not sure.
> 
> Do you want me to dig into that?

That sounds really helpful, but if it's more than an hour of effort, it might make more sense to tackle as a separate issue?

src/objects/embed/embed.twig
calebeby: Yeah, I love this syntax! 🎉

.storybook/main.js
calebeby: Sure

calebeby:
I have a solution that feels very close to working. The problem I have is that when the file change is triggered, https://github.com/sysgears/webpack-virtual-modules isn't acatually forwarding the change into webpack (webpack still sees the old version)

Tomorrow morning I'm going to spend a few more minutes debugging that, if it doesn't work out, then I'll open an issue to tackle this sepraately

calebeby:
@tylersticka Can you take another look and test out handling new files without restarting?

calebeby:
@Paul-Hebert Those suggestions make a lot of sense 👍 

I haven't dug into how much of that is possible via PHP twig, once we get that figured out, we can go ahead and implement some or all of those changes

calebeby:
@tylersticka @Paul-Hebert I resolved a merge conflict in this file: https://github.com/cloudfour/cloudfour.com-patterns/pull/751/files#diff-55e20ee15440dd0343356ac5752e62d2

tylersticka:
> * Remove the `twig` extension: `@cloudfour/components/button/button`
> 
> * If the last two path chunks are identical, remove the last chunk: ``@cloudfour/components/button`

Yeah, I thought of this, too, but didn't suggest it because I _think_ we get the namespace prefix for PHP-land for free (but not full rewriting of paths). Definitely something to explore in the future, though!

> * Shorten `cloudfour` to `c4`: `@c4/components/button`

I have mixed feelings about this suggestion.

Issue #720, opened 2020-05-19
Cloud Divider


haysjoey:
I'd really like to focus on a revised cloud cover design that takes into account the following:

- Avoids repetition if possible to decrease possibility of horizontal/vertical artifacts in some browsers at different zoom levels.
- Takes greater advantage of space at wider viewports.
- Does not cover entire bottom of container to avoid points of tension with rectangular content below.
- Does not immediately appear to be a butt.

haysjoey:
https://github.com/cloudfour/cloudfour.com-patterns/pull/717

tylersticka:
@aileenjeffries Here are the thoughts I promised to provide earlier this morning…

At larger sizes, it seems like there's still an opportunity to improve the contrast in size between the different cloud areas and to really take advantage of the space:

![cloud-sketch](https://user-images.githubusercontent.com/69633/82367149-329ed380-99c8-11ea-9ded-8c5c1492f06b.jpg)

I also wonder if in actual use we'll want this to be symmetrical or if it would be better left to the right-hand side, where there's naturally a bit of an awkward gap currently:

<img width="1679" alt="Screen Shot 2020-05-19 at 11 58 08 AM" src="https://user-images.githubusercontent.com/69633/82367258-5e21be00-99c8-11ea-95ea-78a5b6b7708a.png">

PR #752, opened 2020-05-28
hero section clouds

aileenjeffries:
This is a bit of an evolution of the original clouds, which are intended to act as a divider between longer sections of content. Tyler pointed out we need one for "hero" sections, which tend to have a very short chunk of text on the left and nothing on the right, or an illustration. So, it makes more sense to have just clouds on the right for hero sections. 

This is still very WIP / prototyping. The CSS is just a glorious stew of new selectors and old flex and SUIT CSS utils. :) I'm happy with how the clouds look visually, but I think the implementation is clunky. I have a feeling there's so much fun stuff that we could do to move individual clouds around as screen size increases. So, I'd like to get this merged in so others can look at it and either play with it themselves or make implementation suggestions. 

hero-clouds-lg

Screen Shot 2020-05-28 at 2 17 46 PM


tylersticka:
See #720

PR #754, opened 2020-05-28
Chore: Adds embed object mixin

derekshirk:
## Overview

This PR was originally intended to address the needs of styling the Gutenberg core/image block. After further effort I realize that the work represented here thus far, no longer meets the needs of that goal. 

However, the embed object mixin may still be a meaningful addition. This pull request accomplices the following: 

- adds mixin for embed object
- updates `o-embed` to make use of available mixin

I am opening this up for review as is. What do you think? Are these changes worth introducing even though they aren't aligned with the original intent of styling the Gutenberg core/image block?

Gutenberg core/image block styles will be addressed and added via #802

## Testing

1. View [deploy preview](https://deploy-preview-754--cloudfour-patterns.netlify.app/?path=/docs/objects-embed--image)
1. Review embed object output
1. Confirm no regressions have been introduced

---

See #710


tylersticka:
@derekshirk I tried it out locally and it seems fine to me.

The way I tested was that I changed the `@supports` query to some nonsense (I believe I used `floob: whut`), confirmed the fallback styles were applied, then reverted the query to what it was before and confirmed the enhanced styles were applied.

derekshirk:
I will update/add the Changeset once this is ready for review.

PR #755, opened 2020-05-29
Add support for inlining stories in docs

tylersticka:
## Overview

[Storybook Docs](https://github.com/storybookjs/storybook/tree/master/addons/docs) support two methods of rendering stories:

- Within an iframe.
- Inline within the document.

Both have pros and cons. With an iframe, media queries will behave as expected, and state changes will be nicely sandboxed (no conflicts between `id` attributes, for example). But inlined components render much faster and don't need a `height` defined.

Prior to this PR, only the iframe option was available to us. This is because Storybook's HTML preset does not ship with [the necessary `prepareForInline` function](https://github.com/storybookjs/storybook/blob/v5.3.19/addons/docs/docs/docspage.md#inline-stories-vs-iframe-stories), which converts a story function to a React component that Storybook will understand how to render.

On a lark, I tried writing my own `prepareForInline` function using [the html-to-react package](https://github.com/aknuds1/html-to-react). To my surprise, it seemed to work without much finagling.

The majority of our stories seem to be a good fit for inlining, so I made that the new default. I then went through our existing stories to remove unnecessary `height` attributes and to opt _out_ of inlining for those few that benefited from sandboxing.

### TL;DR

- We can now inline stories.
- Stories are inline by default.
- Individual stories can opt out of inlining by setting `inline` to `false`.
- Entire docs pages can opt out of inlining by setting the `docs.inlineStories` parameter to `false`.

## Testing

Compare [deploy preview](https://deploy-preview-755--cloudfour-patterns.netlify.app/) to [v-next](https://v-next--cloudfour-patterns.netlify.app/?path=/story/getting-started--page). Inlined stories should render faster, have automatic heights, and should not introduce regressions.


calebeby:
@tylersticka This is great! Could you add comments on the individual stories which are opted out of `inline`, describing why they benefit from having the iframe?

tylersticka:
@calebeby Done!

PR #788, opened 2020-06-10
Add Theo support to Rollup configuration

### LINKED ISSUE #787: Figure out how to consume design tokens in client-side JS
Related to: #718 

This issue was briefly discussed [in Slack](https://cloudfour.slack.com/archives/C04AEEVK2/p1591735057108500) in connection with implementing different behavior for the Sky Nav at different breakpoints... 

**Background**
Our design tokens live in centralized yaml files that are consumed by The Patterns. At the time of writing this, we are importing them into `scss` files. But we do not (yet) reference them directly in client-facing JavaScripts.

The tool we are currently using for abstracting tokens is [Theo](https://github.com/salesforce-ux/theo). However, Tyler expressed interest in this alternative as well: https://amzn.github.io/style-dictionary/#/

**Def of Done**
- [ ] Ability to reference a design token variable in a client-facing JavaScript file
-  ~a solution for how to _use_ `em` values in a JavascriptContext*~

\* This may be trickier than it sounds, because we're currently using fluid font sizes that scale up/down proportionally to the viewport width. So it's not as simple as multiplying by the font-size (integer) of the relevant breakpoint

tylersticka:
## Overview

Adds support for importing tokens from Theo configuration files in YAML format. (There is [a Rollup plugin for Theo](https://www.npmjs.com/package/rollup-plugin-theo), but it appears to only transform to CSS.)

Tokens use [the `module.js` format](https://github.com/salesforce-ux/theo#modulejs), which converts tokens into ES exports.

So given a token file like this:

```yaml
aliases:
  black_transparent_85: 'rgba(0, 0, 0, 0.85)'
props:
  - name: text_dark
    value: '{!black_transparent_85}'
    category: text-color
    comment: Accessible on white and lightest gray.
```

And JavaScript source like this:

```js
import * as colors from 'path/to/design-tokens/colors.yml';

console.log(colors.textDark);
```

The compiled JavaScript will look like this:

```js
/* Accessible on white and lightest gray. */
const textDark = "rgba(0, 0, 0, 0.85)";

console.log(textDark);
```

This also adds usage examples in JavaScript to the design token pages.

## Screenshots

Screen Shot 2020-06-10 at 12 11 35 PM

## Testing

1. Check out branch locally.
1. Attempt to import a design token of your choosing using one of those documented in Storybook. (To avoid having to create a new file, try adding it to the Elastic Textarea script.)
1. Run `npm run build`
1. See if the token made it to `dist/cloudfour-patterns.js`.

---

- Resolves #787

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -33,6 +33,7 @@
       "rules": {
         "@typescript-eslint/explicit-module-boundary-types": "off", // inference is usually useful, even for module boundaries
         "@typescript-eslint/no-unused-vars": "off", // TypeScript itself checks this via the "noUnusedLocals"/"noUnusedParameters" options
+        "@typescript-eslint/no-unsafe-member-access": "off", // Makes importing from Theo token files troublesome
         "@typescript-eslint/ban-ts-comment": [
           "error",
           { "ts-expect-error": false }
--- a/.theo/formats/stories.mdx.js
+++ b/.theo/formats/stories.mdx.js
@@ -1,4 +1,5 @@
 const groupBy = require('lodash/groupBy');
+const camelCase = require('lodash/camelCase');
 const kebabCase = require('lodash/kebabCase');
 const startCase = require('lodash/startCase');
 const { basename, extname } = require('path');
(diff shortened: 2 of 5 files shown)


tylersticka:
@calebeby Looking at this again, I think I can generate modules and declarations pretty easily. Trying it out instead. Might be nicer to use overall.

tylersticka:
This is no longer necessary now that #790 has been merged.

PR #790, opened 2020-06-11
Generate JavaScript tokens

### LINKED ISSUE #787: Figure out how to consume design tokens in client-side JS

tylersticka:
## Overview

This is an alternative approach to #788. Instead of exposing Theo tokens directly via Rollup, the tokens are preprocessed and JavaScript versions are generated.

The drawback to this approach are that the import paths are a little longer and recompiling token changes is a bit slower.

But there are a few benefits:

1. It appears to play nicely with TypeScript, and does not seem to require any separate declaration files or additions to our existing `types.ds` file.
1. It allows you to `import` values directly instead of as a glob.
1. In the future, this may invalidate our need for the Theo Webpack loader, which would further simplify our configuration.

## Screenshots

Screen Shot 2020-06-11 at 1 48 48 PM

## Testing

1. Check out branch locally.
1. Attempt to import a design token of your choosing using one of those documented in Storybook. (To avoid having to create a new file, try adding it to the Elastic Textarea script.)
1. Run `npm run build`
1. See if the token made it to `dist/cloudfour-patterns.js`.
1. Run `npm run lint` and `npm run type` to make sure there are no errors after that change was made.

---

- Resolves #787

/CC @calebeby


.theo/formats/stories.mdx.js
```diff
@@ -141,8 +146,14 @@ import { Meta, ColorPalette, ColorItem } from '@storybook/addon-docs/blocks';
 # ${title}
 
 \`\`\`scss
-@use 'path/to/${filename}';
-$example: ${slug}.$${kebabCase(props[0].name)}; // => ${props[0].value}
+@use 'path/to/design-tokens/${filename}';
```
spaceninja, line 149: I found the `path/to/design-tokens` bit confusing. Could we either change this to `../../design-tokens` which will be the case for most components, or add a bit of documentation explaining that this is the path to the generated design tokens file relative to the file you're editing?

tylersticka:
@cloudfour/dev @calebeby My heart sank as I added `preprocess` to our build script and encountered a new error:

```
error TS5055: Cannot write file '/Users/tylersticka/Repos/cloudfour.com-patterns/src/design-tokens/generated/colors.js' because it would overwrite input file.
```

I tried following the recommendations [in this Stack Overflow thread](https://stackoverflow.com/questions/40460790/getting-a-storm-of-error-ts5055-messages-on-every-typescript-compile-in-vs), but they don't seem to work.

I'm beyond frustrated. So far, 100% of my experience with TypeScript has been resolving unhelpful errors. I don't know what else to try.

.theo/formats/stories.mdx.js
tylersticka: Good idea 👍

calebeby:
This looks like it is an error coming from `dts-bundle-generator`, if `npm run type` passes but the gulp buildScripts fails.

If that is the case, then I think we should find an alternative to `dts-bundle-generator`, since it has been causing a lot of problems

spaceninja:
@calebeby as a way to keep Tyler moving forward with this PR, could we temporarily comment out the `buildTypes` command?

If not, is there something we could do to help Tyler get this unblocked? I'm unfamiliar with what `dts-bundle-generator` is doing, so I'm not confident in finding a replacement.

Basically, I think we need to prioritize getting this PR unblocked. Whether that means we fix the root problem, or add a short-term workaround and file an issue to fix the problem the right way.

calebeby:
Since we have continuous publishing to npm, I think we should not merge this PR to v-next with buildTypes commented out, because that means that a release could be triggered to npm with types missing, which should be a breaking change.

@tylersticka If you want I can make the changes in another PR today to switch out dts-bundle-generator

tylersticka:
@calebeby That would be great! Thank you for your help.

tylersticka:
Feedback addressed and updated with @calebeby's TypeScript improvements. _Appears_ to work on my end once testing steps are followed.

PR #796, opened 2020-06-15
Switch to rollup-plugin-dts

calebeby:
## Overview

Context: https://github.com/cloudfour/cloudfour.com-patterns/pull/790#issuecomment-644280325

Switches from `dts-bundle-generator` to https://github.com/Swatinem/rollup-plugin-dts

I am hopeful that `rollup-plugin-dts` has fewer bugs than `dts-bundle-generator`, which has frequently been frustrating for us.

`rollup-plugin-dts` generates a bundled `.d.ts` file by using rollup's tree-shaking and scope merging abilities. Admittedly, this is kind of a hack, but I think this is better than `dts-bundle-generator`. `dts-bundle-generator` builds on TypeScript's compiler API which not really stable.

`rollup-plugin-dts` leaves most of the work to `tsc`, and it only performs the last step of stitching together the exported type files (and more importantly leaving out unwanted type definitions like for twig, cypress, etc.). Because of this I trust this tool more.

## Testing

1. Manually delete the `dist` directory
1. `npm run build:types`
1. Now you should see a `ts-dist` folder that has all the type definitions for everything in it. (Note: this will also get created when you run `npm run type`
1. You should also see a `ts-dist/generated-ts-root-file.d.ts` file. This is created by the gulp types task
1. You should also see `dist/cloudfour-patterns.d.ts` file. This is the only type file to published to npm. It should have _only_ the type definitions for the actual authored JS in this project, none of the types from twig, cypress, etc.


calebeby:
I am leaving out a changeset because this produces functionally equivalent output to the previous version. Compare the output in `dist/cloudfour-patterns.d.ts` to [the previous version on unpkg](https://unpkg.com/browse/@cloudfour/patterns@0.2.0/dist/index.d.ts)

calebeby:
@spaceninja you need to delete the file you added for testing Tyler's PR. This PR does not include the changes from Tyler's PR

Issue #789, opened 2020-06-10
Resolve differences in Storybook vs. Wordpress visual presentation


derekshirk:
I was a bit confused by this and thought opening an issue would be the easiest way to communicate what I am experiencing and ask for help. I hoped that the process of me writing this down would help me gain some clarity into the issue and sure enough it did!

## TL:DR
I believe I need to modify the changes that were introduced in #778 to update the order in which CSS is loaded for our Gutenberg previews. 

If anyone is interested in following along with the thought processes that lead me to this conclusion I am going to keep most of what I had originally intended to include in this issue below. 

---

## The issue

#778 introduced the `@wordpress/block-library` styles into our Storybook previews. 

The part that has me scratching my head a bit is, I am experiencing visual discrepancies between the visual output of what I see in Storybook vs. what I see in a local version of cloudfour.com with the same CSS. 

Example: 

When testing on a local.cloudfour.com environment I see this:

![Screen Shot 2020-06-10 at 3 40 49 PM](https://user-images.githubusercontent.com/1427548/84325995-e0695200-ab30-11ea-85f7-1e6331fee7b1.png)

That same CSS within a Storybook preview looks like this: 

![Screen Shot 2020-06-10 at 3 42 41 PM](https://user-images.githubusercontent.com/1427548/84326286-6daca680-ab31-11ea-950b-14592a79ffd6.png)

I _can_ get around this by increasing the specificity of some selectors, which also fails to adhere to our CSS guidelines and best practices, by doing this. 

Changing this: 

```css
  .wp-block-file__button,
  :not(.wp-block-file__button) {
    @include button.default;
    opacity: inherit !important; /* 1 */
  }
```

To this: 

```css
  a.wp-block-file__button,
  :not(.wp-block-file__button) {
    @include button.default;
    opacity: inherit !important; /* 1 */
  }
```

I don't want to style a selector with a tag and I don't think I should have to do that. Which leads me to believe that perhaps there is a "source order" discrepancy and the CSS is not being loaded in the same order in both environments. ~However, I have yet to be able to confirm or deny this.~

When I view source of my local WordPress install, I see the following CSS being loaded in this order: 

```html
<link rel='stylesheet' id='wp-block-library-css'  href='https://local.cloudfour.com/wp-includes/css/dist/block-library/style.min.css?ver=5.4.1' type='text/css' media='all' />
[...]

PR #778, opened 2020-06-04
Chore: imports WordPress block-library styles

derekshirk:
## Overview

So far, we have not been including the WordPress default styles for Gutenberg block elements in Storybook. This PR imports the missing stylesheet into our Gutenberg story documentation. 

## Screenshots

Observe the before and after presentation of various block elements

**Before:** 
Screen Shot 2020-06-04 at 11 43 25 AM

**After:** 
Screen Shot 2020-06-04 at 3 45 04 PM

**Before:**
Screen Shot 2020-06-04 at 3 48 35 PM

**After:** 
Screen Shot 2020-06-04 at 3 45 18 PM

## Notes

After installing `@wordpress/block-library` via NPM I observed these security audits in my terminal:

Screen Shot 2020-06-04 at 3 39 37 PM

I assumed the Renovate bot would address these so no action was taken in this PR. 

## Testing

1. Compare the Gutenberg stories in the [master branch](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/vendor-wordpress-gutenberg--core-audio) to the same blocks in the [deploy preview](https://deploy-preview-778--cloudfour-patterns.netlify.app/) for this PR. 

---

See: https://developer.wordpress.org/block-editor/packages/packages-block-library/
See: #710

PR #798, opened 2020-06-16
Chore: Update index.scss to include gutenberg blocks library CSS

derekshirk:
## Overview

See #789, and Tyler's suggestion [here](https://github.com/cloudfour/cloudfour.com-patterns/issues/789#issuecomment-642322457). 

In order to remedy a CSS source order loading issue that was causing the WordPress Gutenberg block library CSS from trumping our patterns CSS in Storybook previews, this PR does the following: 

• Creates a new `index-with-wp-block-library` file that imports WP styles before our pattern styles (index.scss)
• Updates `.storybook/preview.js` to include the new `index-with-wp-block-library` instead of `index`
• Removes `@wordpress/block-library/build-style/style.css` from `src/vendor/wordpress/gutenberg.stories.mdx`, previously added in #778 

## Consideration 

With #778 (WP) Gutenberg block styles were only loaded on the Gutenberg stories. With this update, as I understand it, the Gutenberg CSS will now be loaded on every Storybook page. If that is a concern we should explore other alternative approaches. 

### Before
Notice the secondary button is outlined with a black border and black text. 

Screen Shot 2020-06-16 at 3 08 54 PM

### After
Here, the secondary button is styled as expected with our `c-button c-button--secondary` styles.

Screen Shot 2020-06-16 at 2 22 30 PM

## Testing

1. View [deploy preview](https://deploy-preview-798--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-gutenberg--core-button)
1. Confirm the secondary button is outlined with a blue border and has blue text. 

---

See #789


derekshirk:
> If these changes should be published to npm, you need to add a changeset.

I am unsure if this PR requires a changeset. 🤔

.storybook/preview.js
```diff
@@ -8,7 +8,7 @@ import * as breakpoints from '../src/design-tokens/breakpoint.yml';
 import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport';
 import { ratio } from '../src/design-tokens/modular-scale.yml';
 import 'focus-visible';
-import '../src/index.scss';
+import '../src/index-with-wp-block-library.scss';
```
derekshirk, line 11: I am unsure if this file name is laughable or not. I am open to suggestions here.

tylersticka: Maybe `index-with-dependencies` in case we want to add something else later?

derekshirk: I like that suggestion. Thank you. Updated.

PR #802, opened 2020-06-18
Chore: Styles Gutenberg core image block

derekshirk:
## Overview

This pull request adds a few small style adjustments to Gutenberg image blocks to improve layout and alignment for WP image blocks. 

- Removes margin-bottom from image blocks
- Fixes issue with smaller/resized images with center alignment

Additionally, This PR now adds CSS support for Gutenberg wide and full images. 

## Screenshots

Screen Shot 2020-08-10 at 1 33 53 PM

Screen Shot 2020-08-10 at 11 24 58 AM

### Before alignment fix

Screen Shot 2020-06-22 at 11 38 40 AM

### After alignment fix

Screen Shot 2020-06-22 at 11 39 33 AM

## Testing

1. Review [changes](https://github.com/cloudfour/cloudfour.com-patterns/pull/802/files)
1. Check code comments, make sure they are clear, concise, explain the "why" of what they are annotating and above all that the changes they are introducing are sensible. 

---

See #710


derekshirk:
After I opened this PR up for review, I realized that this PR did not include the changes I intended to add to the core/image documentation about how `u-pull-inline` utility classes can be used to extend the left and right edges of an image outside of the parent prose container. 

I plan to update that before merging.

src/vendor/wordpress/gutenberg.stories.mdx
```diff
@@ -365,15 +365,19 @@ The image block is wrapped in a `figure` element with the `wp-block-image` class
...
 
-Adding caption text to the image generates a `figcaption` element before the closing
-wrapper tag.
+Adding caption text to the image generates a `figcaption` element before the
+closing wrapper tag.
+
+`u-pull-inline-{amount}` class names from our [negative
+margin](/?path=/docs/utilities-spacing--negative-margin) utility classes can
+be added to the additional css section of the image block to extend the left and
+right edges of images beyond their parent prose container.
```
derekshirk, line 380: I've updated this section of documentation and re-requesting review.

tylersticka: I'm surprised to see the direction to use utility classes. Is it possible for us to add opinionated defaults for the breakout alignment options I see in the Gutenberg demo ("wide width," "full width")? I'd love to get these common things out of our post drafts and into our code.

Screen Shot 2020-06-23 at 10 24 51 AM

derekshirk: > Is it possible for us to add opinionated defaults for the breakout alignment options I see in the Gutenberg demo

It's possible! I will update this PR to account for the "wide width" and "full width" settings.

derekshirk: @tylersticka I know you had a busy day, so no need to respond to this right away. 

Part of _my_ confusion was that my image block editor options looked different than your screenshot above from your Gutenberg _demo_. In my local development environment, I'm using the version of Gutenberg that ships with WordPress and my 'alignment'  options only include; left, center and right options. 

Screen Shot 2020-08-06 at 3 36 40 PM

As far as I can tell, this option needs to be enabled via `add_theme_support` 
https://gogutenberg.com/blocks/image/

Which appears to work...

Screen Shot 2020-08-06 at 4 24 46 PM

Unless I'm missing something, my plan is to submit a PR to the WP repo to allow for these "advanced alignment" options as well.

src/vendor/wordpress/_wordpress.scss
```diff
@@ -111,6 +106,44 @@ $wp-button-gap: sizes.$list-inline-gap;
...
+      width: 100vw;
+    }
+  }
+
+  &.alignwide {
+    @media (min-width: breakpoint.$m) {
+      left: -6vw; /* 3 */
+      width: calc(100% + 12vw); /* 3 */
+    }
+  }
```
derekshirk, line 144: The resulting width applied via this selector may seem a bit magical and opinionated. I arrived at this after surveying cloudfour.com and trying to establish a reasonable width based on what I saw being used most often in articles.

derekshirk:
Hi @tylersticka, 

I've added some image styles for `alignwide` and `alignfull` based on earlier [feedback](https://github.com/cloudfour/cloudfour.com-patterns/pull/802#discussion_r444388140). 

My intent for the following classes: 

`.alignfull`

This results in the image block always extending to both viewport edges at all breakpoints

image-alignfull

`.alignwide`

This results in the image block extending to both viewport edges on smaller screens and then restricted to an opinionated width when the `(min-width: breakpoint.$l)` media query is true. 

image-alignwide


_Note: In order to test this, I was running a local version of the `v-next` cloudfour.com-wp repo, which does not have a max-width prose container yet (AFAICT), so I used the existing padding and width styles from the existing cloudfour.com article container to test._

src/vendor/wordpress/_wordpress.scss
```diff
@@ -111,6 +106,50 @@ $wp-button-gap: sizes.$list-inline-gap;
...
+  &.aligncenter {
+    width: 100%; /* 2 */
+  }
+
+  &.alignfull {
+    left: 50%;
+    margin-left: -50vw;
+    margin-right: -50vw;
+    right: 50%;
+    width: 100vw;
```
tylersticka, line 134: I assume this will cause horizontal scrollbars in browsers/platforms where they are always shown and the page has a scrollbar?

It's okay not to account for this now, but we should probably add a `@todo` comment or something to that effect.

tylersticka, line 134: These styles seem more verbose than the ones I wrote about and that we use on our current site: https://cloudfour.com/thinks/breaking-out-with-viewport-units-and-calc/#flipping-the-script

src/vendor/wordpress/_wordpress.scss
```diff
@@ -111,6 +106,50 @@ $wp-button-gap: sizes.$list-inline-gap;
...
+    @media (min-width: breakpoint.$l) {
+      left: calc(12vw / -2); /* 3 */
+      margin-left: auto;
+      margin-right: auto;
+      right: auto;
+      width: calc(100% + 12vw); /* 3 */
```
tylersticka, line 148: I think if you used `calc` you could simplify? (See earlier comment.)

src/vendor/wordpress/_wordpress.scss
```diff
@@ -111,6 +106,50 @@ $wp-button-gap: sizes.$list-inline-gap;
...
+ *    layout when placed after a right-aligned image in a post.
+ * 3. Applies an opinionated wide-image setting that allows an image element to
+ *    spill slightly outsides its container at larger viewports.
+ */
+.wp-block-image {
+  margin-bottom: 0; /* 1 */
+  position: relative;
+
+  &.aligncenter {
+    width: 100%; /* 2 */
```
tylersticka, line 126: Can you move the relevant comment so it precedes the relevant block, instead of having the large comment block above everything?

src/vendor/wordpress/_wordpress.scss
derekshirk: I attempted to simplify this, by incorporating the additional feedback below. See 6664d90

derekshirk: I agree. Thank you. I attempted to address this feedback in 6664d90

src/vendor/wordpress/_wordpress.scss
derekshirk: Agreed. I attempted to address this feedback in 6664d90

src/vendor/wordpress/_wordpress.scss
derekshirk: Thank you. I attempted to address this feedback in d6e9fe3

derekshirk:
@tylersticka 

> You might consider adding knobs for this to toggle the various classes, which might make testing easier?

Excellent idea!

**A few notes:** 

1. I'm still a little "green" when it comes to setting up knobs in Storybook, so I imagine my changes will elicit some additional feedback. 

The way the story was configured prior looked like this: 

```mdx
<Preview>
  <Story name="core/image">
    {`<figure class="wp-block-image size-large is-style-default">
        <img
          src="/media/MtBlanc1.jpg"
          alt="Mont Blanc"
          class="wp-image-130"
          srcset="
            /media/MtBlanc1.jpg         500w,
            /media/MtBlanc1-300x225.jpg 300w
          "
          sizes="(max-width: 500px) 100vw, 500px"
        />
        <figcaption>Mont Blanc appears—still, snowy, and serene.</figcaption>
      </figure>`}
  </Story>
</Preview>
```

I wasn't sure how to add knobs to this format, so I followed some examples I observed elsewhere and reformatted the story like this: 

```mdx
<Preview>
  <Story name="core/image">
    {blockImageDemo({
      class: select('alignment class', alignmentClasses, null),
    })}
  </Story>
</Preview>
```

I am curious about the legitimacy of this approach and possibly learning better/alternate ways to configure this. 

2. Regarding the select knob...I ended up adding the options like this: 

```mdx
const alignmentClasses = {
  None: null,
  Left: 'alignleft',
  Center: 'aligncenter',
  Right: 'alignright',
  Full: 'alignfull',
  Wide: 'alignwide',
};
```

I wonder if their is a more efficient way to accomplish this. 


3. I assumed it would help demonstrate the effects of alignment classes, if I added surrounding placeholder text to the block/image story but wanted to confirm if this was desirable or not.

4. I added some conditional logic to `src/vendor/wordpress/demo/image.twig ` that _might_ seem strange (6200fed). I did this because WP actually modifies the markup of the image block it outputs depending the select alignment classes. Feedback welcome!

src/vendor/wordpress/_wordpress.scss
```diff
@@ -111,6 +106,47 @@ $wp-button-gap: sizes.$list-inline-gap;
...
+ * 1. Since we are applying `margin-top` to CMS content via our staggerItems
+ *    utility, the default `margin-bottom` value on Gutenberg image blocks is
+ *    unnecessary, so we remove it here.
+ */
+.wp-block-image {
+  margin-bottom: 0; /* 1 */
+  position: relative;
+
+  /**
+   * 1. Setting a width of `100%` prevents resized and centered images from
```
tylersticka, line 122: Nit: Comment numbering isn't necessary when there is only one line in the block.

src/vendor/wordpress/_wordpress.scss
```diff
@@ -111,6 +106,47 @@ $wp-button-gap: sizes.$list-inline-gap;
...
+   *    breaking layout when placed after a right-aligned image in a post.
+   */
+  &.aligncenter {
+    width: 100%; /* 1 */
+  }
+
+  &.alignfull {
```
tylersticka, line 129: Consider adding a preceding comment linking to the article this was pulled from for context.

src/vendor/wordpress/_wordpress.scss
```diff
@@ -111,6 +106,47 @@ $wp-button-gap: sizes.$list-inline-gap;
...
+   *    spill slightly outsides its container at larger viewports.
+   */
+  &.alignwide {
+    margin-left: calc(-50vw + 50%);
+    margin-right: calc(-50vw + 50%);
+    @media (min-width: breakpoint.$l) {
+      left: calc(12vw / -2); /* 1 */
+      margin-left: auto;
+      margin-right: auto;
+      width: calc(100% + 12vw); /* 1 */
```
tylersticka, line 145: Would this accomplish the same thing?

```suggestion
      margin-left: -6vw; /* 1 */
      margin-right: -6vw; /* 1 */
```

src/vendor/wordpress/gutenberg.stories.mdx
```diff
@@ -1,7 +1,17 @@
 import { Story, Preview, Meta } from '@storybook/addon-docs/blocks';
+import { withKnobs, select } from '@storybook/addon-knobs';
 import codeBlockDemo from '../../design/typography/demo/code-block.twig';
-
-<Meta title="Vendor/WordPress/Gutenberg" />
+import blockImageDemo from './demo/image.twig';
+const alignmentClasses = {
+  None: null,
```
tylersticka, line 6: Storybook select values are converted to strings no matter what they are, so (annoyingly) `null` is the same as passing in a string called `'null'`. You _can_ pass in an empty string, though, (`''`), which should have the desired effect.

(It's times like these I wish knobs were better documented... but they've been replaced by a more elegant solution in the latest Storybook release, so that's nice!)

src/vendor/wordpress/gutenberg.stories.mdx
```diff
@@ -365,31 +375,20 @@ The image block is wrapped in a `figure` element with the `wp-block-image` class
...
-          srcset="
-            /media/MtBlanc1.jpg         500w,
-            /media/MtBlanc1-300x225.jpg 300w
-          "
-          sizes="(max-width: 500px) 100vw, 500px"
-        />
-        <figcaption>Mont Blanc appears—still, snowy, and serene.</figcaption>
-      </figure>`}
+    {blockImageDemo({
+      class: select('alignment class', alignmentClasses, null),
```
tylersticka, line 390: I would either shorten this to "alignment," or show the full class names (`alignright` instead of `Right`). As is, it says "alignment class" but it's not actually showing the class.

src/vendor/wordpress/demo/image.twig
```diff
@@ -0,0 +1,30 @@
+<div
+  class="demo-container"
```
tylersticka, line 2: Consider using [the Rhythm object pattern](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/objects-rhythm--default-story) for this, especially since your styles account for it.

src/vendor/wordpress/gutenberg.stories.mdx
derekshirk: Good point! Excellent feedback.

src/vendor/wordpress/gutenberg.stories.mdx
derekshirk: Thank you, that was extremely helpful. I had originally started with an empty string, but ended up mimicking what I saw in the [documentation I was reading](https://github.com/storybookjs/storybook/tree/master/addons/knobs#select), which used `null`. 🙃

src/vendor/wordpress/_wordpress.scss
derekshirk: It does! I am introducing these changes manually.

src/vendor/wordpress/_wordpress.scss
derekshirk: The GitHub UI isn't showing this comment as "outdated" but it was updated in a190909

derekshirk:
Hi @tylersticka, 

> I notice if I change from the "None" alignment option to "Center," the amount of whitespace below the example increases but nothing else does.

I want to make sure I understand expectations here correctly. 

### 1. Whitespace question

I can't determine if you are suggesting that the amount of whitespace should change or not. I had originally assumed you were suggesting that it _should not_, so I pushed one possible "solution" (3ed4aae) to "remedy" that. However, now I'm not so sure!

**Example of update introducing _no_ change in whitespace**

knobs-alignment-update

If I compare behavior of c4staging.wpengine.com/ master branch I notice the white space _does_ change when switching between these two alignment options: 

knobs-alignment-comparison

_Note the classic editor will apply a class of `alignnone` while the block editor does not. `alignnone` does not appear to have any styles associated with it._

### 2. Alignment question
I also want to make sure I understand the second part of your comment: 

> but nothing else does. 

I interpret this as, "when 'None' is selected and the image size is reduced it should _not_ maintain centered alignment".  It appears that "None" and "Center" currently behave similarly when reducing the image size due to some styles defined in our base CSS.

`/src/base/_defaults.scss`

```css
/*
 * ...
 * 2. Center content
 * ...
 */

figure {
  ...

  & > * {
    margin-left: auto; /* 2 */
    margin-right: auto; /* 2 */
  }

  ...
```

If my interpretation is correct, is it more desirable to override the existing `<figure>` styles defined in `_defaults.scss` within our vendor `_wordpress.scss` (only for `.wp-block-image` elements), so that when no alignment class is selected the resulting behavior is like this:

knobs-alignment-no-styles

An alternate approach might require rethinking how the base `<figure>` styles are written but centering `<figure>` contents by default seemed smart to me. 

Thanks!

tylersticka:
> I can't determine if you are suggesting that the amount of whitespace should change or not. I had originally assumed you were suggesting that it _should not_, so I pushed one possible "solution" ([3ed4aae](https://github.com/cloudfour/cloudfour.com-patterns/commit/3ed4aae9fdc28bfec748bee48fbaec93544846d8)) to "remedy" that. However, now I'm not so sure!

I couldn't see a reason why one would have whitespace and the other would not. The issue is the (seemingly unintentional) inconsistency, not the presence of or lack of whitespace.

Was the whitespace intentional, or an oversight? If it was not intentional, and there isn't a design justification for it, then we should remove it. If it _was_ intentional but was omitted from the default, then we should add it.

Does that make sense?

> I also want to make sure I understand the second part of your comment:
> > but nothing else does.

This isn't a second part, I just meant that only `aligncenter` adds whitespace. It's just the inconsistency.

If you _intended_ for `aligncenter` to have more whitespace below it, then we should have _added_ whitespace consistency. If you _did not_ intend for `aligncenter` to have more whitespace below it, then you were correct in removing it.

tylersticka:
To summarize:

- Did you intend for `aligncenter` to have more whitespace below it?
  - **Yes:** I'd like to know why. We can't just have inconsistency without a justification.
  - **No:** Okay, is it an improvement (or "happy accident")?
    - **Yes:** Then we should add it to the other options as well.
    - **No:** Then it was good idea to remove it.

derekshirk:
> Did you intend for `aligncenter` to have more whitespace below it?

Based on my observations on cloudfour.com I had _thought_ the additional whitespace was intentional. So my only "intent" at the time was to mimic what I thought was expected, due to the `display: block;` property included with WordPress's `aligncenter` class. This is by no means a justification. 

cloudfour-dot-com-aligncenter-whitespace

> No: Okay, is it an improvement (or "happy accident")?
>  → Yes: Then we should add it to the other options as well.
>  →  No: Then it was a good idea to remove it.

No, I don't think so. I think the additional whitespace throws off the vertical rhythm applied by our Rhythm object pattern. 

Happy to discuss further.

derekshirk:
Hi @tylersticka, 

I added a changeset (69c1af7) and it dismissed your approval.

PR #848, opened 2020-07-15
Improve GitHub Actions Performance

spaceninja:
## Overview

This PR makes several changes to improve the performance of our GitHub actions, with the goal of reducing the number of minutes used.

- Only run jobs on `pull_request` events to the `main` branch.
- Removes trigger to run jobs on `push` events.
- Consolidates lint, test, build, and type checking into one job, saving time running duplicate steps in each job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,67 +1,11 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches:
+      - v-next
 
 jobs:
-  lint:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - name: Use Node.js 12
-        uses: actions/setup-node@v1
-        with:
-          node-version: 12
-      - name: Cache node modules
-        uses: actions/cache@v1
-        with:
-          path: ~/.npm
-          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: ${{ runner.os }}-node-
-      - name: Install Dependencies
-        run: CYPRESS_INSTALL_BINARY=0 npm ci
-      - name: Run Lint
-        run: npm run check-lint
-
-  type:
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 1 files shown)


spaceninja:
> GitHub Actions usage is free for public repositories.

There's no minute limit for this repo, but since this repo exists as an example of our current best practices, it should have the same performance improvements I just made for a client project.

calebeby:
@spaceninja I like to be able to see the status checks on branches through the commits view. Personally I think it would be better to keep GH actions running on branches

calebeby:
Also since separate actions run concurrently, it is usually faster to have multiple separate actions. For example in [this PR](https://github.com/cloudfour/cloudfour.com-patterns/pull/802), each of the individual actions were faster.

Also it is nice to be able to see which check failed directly from the checks view rather than having to click on it (often for example if it says "lint failed" then I just go directly to running `npm run lint` locally to fix the problem rather than clicking through to the CI view)

spaceninja:
That's fair. It's true that it's a bit slower overall to run all the checks in a single job, though it does save on total build minutes. Given that we're not charged by the build minutes here, there's definitely a case to be made for leaving things as-is.

However, I think the point about this repo being our "best practices" repo remains valid.

I'm tagging in the rest of the dev team on this for further consideration.

Paul-Hebert:
I don't feel super strongly about this, but I think @calebeby makes some compelling points for keeping things as-is.

I agree that this site should showcase our best practices, but I think what the best practices are varies depending on the constraints we're working under. If there's no constraint on build minutes I don't know that it's a best practice to optimize for reducing build minutes.

gerardo-rodriguez:
> I don't feel super strongly about this, but I think @calebeby makes some compelling points for keeping things as-is.
> 
> I agree that this site should showcase our best practices, but I think what the best practices are varies depending on the constraints we're working under. If there's no constraint on build minutes I don't know that it's a best practice to optimize for reducing build minutes.

➕1️⃣

spaceninja:
Thanks y'all! I'll close this PR and leave things as-is. :)

PR #865, opened 2020-07-27
Add Storybook Build Step to CI

spaceninja:
This commit adds a new job to the GitHub Actions CI to run a Storybook
build. Previously, this wasn't happening, which led to CI passing even
though Storybook build failed on Netlify.

This also removes the preprocess step from the build job because the
build script already runs the preprocess step.


calebeby:
@spaceninja Since Netlify already runs the storybook build and reports the status to GitHub, what benefit is there to also having GH Actions run it?

spaceninja:
> Since Netlify already runs the storybook build and reports the status to GitHub, what benefit is there to also having GH Actions run it?

That's a fair question. I just thought it was odd to see a green check by CI and a fail by Netlify. But I suppose if we think of Netlify as another CI check, rather than an environment, it's reasonable.

Maybe it's not needed after all?

calebeby:
> But I suppose if we think of Netlify as another CI check

This is the way that I think of it.

I'm fine with whatever, but for me personally a green check next to GH Actions is no different from a green check next to Netlify

spaceninja:
> for me personally a green check next to GH Actions is no different from a green check next to Netlify

Yep, you've convinced me. Closing this PR, and thanks for the insight!

PR #893, opened 2020-08-12
Add ground nav styles

### LINKED ISSUE #480: Site footer pattern
Create a footer pattern to be used throughout the site.

site footer

dromo77:
## Overview

This PR adds CSS to existing markup for the `ground-nav` component. The `ground-nav` uses CSS grid for the layout with fallback styles for browsers that don't support grid.

It also introduces a new utility, `u-hidden`, which can be used to hide content at different screen sizes.

### cloudfour.com
c4-sm
c4-lg

### PWA stats
pwa-sm
pwa-lg

### Fallback
c4-fallback-lg
pwa-fallback-lg

---

- Fixes #480


src/components/ground-nav/ground-nav.twig
```diff
@@ -10,101 +10,120 @@
...
-            {% if item.current %}
-              aria-current="page"
-            {% endif %}>
-            {{item.title}}
-          </a>
-        </li>
-      {% endfor %}
-    </ul>
-  </nav>
+      <hr class="c-ground-nav__divider u-hidden@m">
```
tylersticka, line 94: I learned recently that horizontal rules have more semantic meaning and accessibility impact than I thought they did:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr
- https://www.w3.org/WAI/GL/WD-WCAG10-TECHS-20000615/Overview.html#structure

We should avoid using them and instead use CSS where possible.

I'd also encourage you to build the responsive hide/show functionality into the component itself where it's necessary to achieve the desired layout.

src/components/ground-nav/ground-nav.twig
```diff
@@ -10,101 +10,120 @@
...
+          {% endfor %}
+        </ul>
+      </nav>
 
-  {% if legal %}
-    <div>
-      {{legal}}
+      <hr class="c-ground-nav__divider u-hidden@l">
+
+      <div class="c-ground-nav__copyright">
```
tylersticka, line 118: This may contain more than just copyright, so can we use `byline` instead?

src/utilities/display/display.scss
```diff
@@ -7,3 +9,33 @@
...
+  @include display.hidden(xl);
+}
+
+@media (min-width: breakpoint.$xxl) {
+  @include display.hidden(xxl);
+}
+
+@media (min-width: breakpoint.$xxxl) {
+  @include display.hidden(xxxl);
+}
```
tylersticka, line 41: See earlier comment about whether or not this is necessary for this component.

static/portland.svg
```diff
@@ -0,0 +1,37 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="140" height="70" viewBox="0 0 140 70">
+  <style type="text/css">
+    .st0{fill:#f2f5f7;}
+    .st1{fill:#738da0;}
+    .st2{fill:#158466;}
+    .st3{fill:#2cdd90;}
+    .st4{fill: none; stroke: #f2f5f7; stroke-width: 2.4; stroke-miterlimit: 3.2;}
+    .st5{fill: none; stroke: #f2f5f7; stroke-width: 2.4; stroke-miterlimit: 2.4;}
```
tylersticka, line 8: We should be exporting with these set to attributes so these classes don't inadvertently litter the DOM or complicate the CSS include.

static/portland.svg
```diff
@@ -0,0 +1,37 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="140" height="70" viewBox="0 0 140 70">
```
tylersticka, line 1: Can we move this file to somewhere in `src/assets/`?

dromo77:
Thanks @tylersticka! I'll take a look at the inline feedback and address it.

> Could we remove the top border from this component (since PWA Stats won't use it)?

👍 

> Is there anything we can do to improve the appearance of the trees and buildings?

Sure! I updated the colors so they're consistent with our new palette, but I can revisit the illustration and see where I can make changes.

derekshirk:
Thanks for including me as a reviewer on this PR @dromo77. I will wait to review this until Tyler's feedback has been addressed.

dromo77:
@tylersticka I had fun exploring some different variations of the footer illustration. I wanted to know what you think about these two. The composition is largely the same in the first one, but with more details. In the second I tried adding more depth.

pdx-1-sm
pdx-1-lg

pdx-2-sm
pdx-2-lg

tylersticka:
@dromo77 I like varying the colors and the addition of another background "plane" to add a bit more depth. I'm also open to varying the details on the buildings, though currently the varying window sizes are drawing my eye a bit too much.

I'm not as big a fan of the changes to the tree shapes and the removal of the branches. The originals were abstracted as a sort of hidden tribute to UPA, a progressive mid-century American animation studio that injected cubism and modernism into their work:

e94bff850314c793367e8fe9d543817b

More info: : https://www.youtube.com/watch?v=A8-09DbxHBU

I'd love to see a revision that retains the slightly improved depth but keeps that glimmer of UPA inspiration intact... it's still there in the buildings, less so in the trees.

Happy to chat if you'd like to discuss further!

dromo77:
> The originals were abstracted as a sort of hidden tribute to UPA, a progressive mid-century animation studio that injected cubism and modernism into their work

Cool! Going to check out the video and see if I can find some other images of their work for inspiration 🙂

dromo77:
@tylersticka Made a few tweaks to the illustration.

pdx-v2-sm
pdx-v2-lg

tylersticka:
@dromo77 This is an improvement.

The lighter branches are maybe a _touch_ too thin now to my eye... they're looking a little less like branches and more like twigs now.

I'm not super sure about the colors still. The trees feel really dark, and the buildings feel a bit dissonant with each other.

If you look at a drawing [like this one](https://www.instagram.com/p/B83RqJbh1ao/), notice how the buildings vary quite a bit in hue but not necessarily in value and saturation. Where the value comes in is in the details and the depth.

In images like [this movie poster](http://www.fubiz.net/2018/12/05/the-beautiful-graphics-of-the-100-poster-project/melissa-ya-100-posters-06/), [this magazine](https://www.instagram.com/p/Bp9Mf_Sh5Qz/) or [this illustration](www.fubiz.net/2017/03/28/minimalist-architectural-illustrations-of-californian-buildings/minimalistarchitecturalillustrations3/) it's really the details (linework, etc.) that convey depth. Could those details be used to offset buildings from the trees in a similar way?

Photo Aug 14, 9 27 12 AM

Is there a way for the trees to look more vibrant than they would in real life, more like they're in the sun (even tough there are clouds)? See how greens are paired with architecture [in this illustration](http://www.fubiz.net/2018/01/22/geometric-and-aesthetic-landscapes-by-mike-ellis/mikeellis4/), the landscape of [a lush video game environment](https://www.sonicthehedgeblog.com/post/188680919182/a-panel-from-the-first-issue-of-archie-comics), etc.

(Let me know if this feedback gets frustrating... we can always wrap up everything else and revisit.)

dromo77:
@tylersticka Thank you for the feedback, I'd like to take another pass at the illustration. I've been working with the new color palette, should I explore colors outside of it?

tylersticka:
> I've been working with the new color palette, should I explore colors outside of it?

Sure! As long as the colors feel _harmonious_ with the palette, I don't think we should restrict ourselves when the illustrations call for some unique twists. The illustrations in the home page animation are a good example of where we went outside of our palette a bit when it made sense.

dromo77:
@tylersticka I made some more changes to the illustration. It uses lighter colors that are closer in value. Some of the colors are from the new palette, but I did play around with different shades. I also tried to give the foreground buildings more details than those in the background.

pdx-v3-sm
pdx-v3-lg

tylersticka:
@dromo77 Danielle!! That looks great!!! 🙌 

My only feedback is that there are a couple of small details I think could still be simplified. The lighter stroke around the one building's windows and the little partial overlap lines on another feel like they could be streamlined a bit:

Screen Shot 2020-08-18 at 8 51 48 AM

But overall, this is a lovely evolution!

dromo77:
Thanks @tylersticka, I simplified these details. Here are some updated screenshots:

gound-nav-sm
ground-nav-lg

tylersticka:
@dromo77 I like that!

Can I see how it looks without the inner shadow in the left blue building's windows?

Also, is it just my eyes or is the trunk/branches of the leftmost tree _slightly_ off-center?

dromo77:
@tylersticka I made a few tweaks based on your feedback. I removed the inner shadow from the left blue building windows. I also centered both tree trunks/branches. 

ground-nav-sm
ground-nav-lg

tylersticka:
@dromo77 I like it!

dromo77:
@tylersticka Woohoo! Can you take another look? I pushed up several changes, here's a refresher:

- Added the new illustration to the assets folder
- Replaced the `hr` elements 
- Removed the top border
- Renamed the "copyright" section to "byline"
- Removed the "u-hidden" utility/mixin

derekshirk:
@dromo77 I _really_ like the evolution of the cityscape illustration!

dromo77:
@derekshirk Thank you!

dromo77:
@tylersticka I think the changes I just pushed up should address the design feedback, can you take another look?

src/components/ground-nav/ground-nav.scss
```diff
@@ -0,0 +1,262 @@
...
+@use '../../design-tokens/font.yml';
+@use '../../mixins/ms';
+@use '../../mixins/unit';
+
+// Whitespace used for gaps, padding and margin
+// (It uses `rem` so space will be consistent throughout the component)
+$_ground-nav-space: ms.step(1, 1rem);
+// Negative value used to reduce the space around elements as well as position elements.
+$_ground-nav-negative-space: #{unit.strip($_ground-nav-space) * -1}rem;
+$_ground-nav-border-width: 3px;
```
tylersticka, line 12: Should this correspond to one of our [edge size tokens](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/design-tokens-sizes--page#sizing)?

src/components/ground-nav/ground-nav.scss
```diff
@@ -0,0 +1,262 @@
...
+@use '../../design-tokens/colors.yml';
+@use '../../design-tokens/font.yml';
+@use '../../mixins/ms';
+@use '../../mixins/unit';
+
+// Whitespace used for gaps, padding and margin
+// (It uses `rem` so space will be consistent throughout the component)
+$_ground-nav-space: ms.step(1, 1rem);
+// Negative value used to reduce the space around elements as well as position elements.
+$_ground-nav-negative-space: #{unit.strip($_ground-nav-space) * -1}rem;
```
tylersticka, line 11: Did `($_ground-nav-space * -1)` not work?

src/components/ground-nav/ground-nav.scss
```diff
@@ -0,0 +1,262 @@
+@use '../../design-tokens/breakpoint.yml';
+@use '../../design-tokens/colors.yml';
+@use '../../design-tokens/font.yml';
+@use '../../mixins/ms';
+@use '../../mixins/unit';
+
+// Whitespace used for gaps, padding and margin
+// (It uses `rem` so space will be consistent throughout the component)
+$_ground-nav-space: ms.step(1, 1rem);
```
tylersticka, line 9: FYI: The way Sass variables work, if you prefix them with a leading underscore, they won't leak out into other files as long as those files import the variable with `@use`.

So if you want to be a bit more succinct with them in the context of a component file, you can. No specific action item required.

dromo77:
Thanks for the demo @tylersticka! The absolute positioning and adjacent margins are a fallback for browsers that don't support grid. I made a mistake with my last commit, so it didn't look like I was using CSS grid 🤦‍♀️ . The solution I arrived at is a bit different than the one you shared. Would you mind taking another look?

dromo77:
> Is there a version of this where browsers without grid support see a much simpler version of the nav?

@tylersticka That's a good question! I erred on the side of caution providing a similar design, but I can simplify it if we're ok with the fallback experience looking different.

dromo77:
@tylersticka I simplified the fallback style to stay stacked when grid isn't supported. Here's an example of what that looks like on larger screens.

ground-nav-fallback

src/components/ground-nav/ground-nav.scss
```diff
@@ -0,0 +1,197 @@
...
+.c-ground-nav__content-inner > * + * {
+  margin-top: $_ground-nav-space;
+}
+
+/**
+ * 1. Removes `margin-top` so the organization information is
+ *    aligned with the other footer elements
+ */
+.c-ground-nav__organization {
+  font-style: normal;
```
tylersticka, line 81: Is this here because of the default styles for `address`?

Do you think we should set this for `address` by default? I find it had to imagine we'll want an address that is italicized by default, but I'd love a second opinion. 🤔

dromo77: Yes, I can update the default style for `address` instead. We can always change it in the future if we need to.

dromo77:
@tylersticka Mind taking another look?

Issue #746, opened 2020-05-27
Update favicons


tylersticka:
Our color palette has been updated, so our icons should be updated as well. There are also new best practices for favicons we can follow.

These should also be documented in Storybook (possibly related: #725).

More info: https://css-tricks.com/svg-favicons-and-all-the-fun-things-we-can-do-with-them/

derekshirk:
Note to self: Review favicon additions introduced in #535

aileenjeffries:
@tylersticka I've pushed my branch for this, `aj-favicon` and had a few questions, primarily about documentation. I kept the magenta development / blue production trick since I find it handy, so that means 4 total favicon files. I added a little section under Brand about favicons, explaining why we're stroking it in white. But, I'd love to be able to actually demonstrate the need for this by coloring the Story the icon is shown in, in something similar to Firefox's dark gray. 

My other question is more general, about how to document icons such as the MS Tile. That tile is a two parter - the png, plus a browserconfig.xml file specifying the tile's background color. I added both to the static directory, and updated `manager-head.html` but I'm not sure if this is really sufficient documentation? Same deal with the masked Android/PWA icon. I found this write up: https://css-tricks.com/maskable-icons-android-adaptive-icons-for-your-pwa/ made a new png with the icon being inside the safe zone, and then added a manifest.json because you need to add `"purpose": "maskable"`  to that icon's listing. But again, is this really the ideal documentation? Do we have other pattern libraries for clients that document this sort of stuff that isn't really a pattern but still affects the look and function of the site?

tylersticka:
Relevant: https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs

tylersticka:
The longer we wait on this, the less we need to provide (it seems): https://css-tricks.com/how-to-favicon-in-2021/

PR #908, opened 2020-08-17
Chore: add SVG favicon with dark mode support

derekshirk:
## Overview

This PR represents a single task as part of #746, which adds a favicon in SVG format and experiments with adding support for dark mode via `prefers-color-scheme` media queries. 

## Screenshots

After reading [this article](https://css-tricks.com/svg-favicons-and-all-the-fun-things-we-can-do-with-them/#keep-it-crisp), I thought I would make an attempt at decreasing antialiasing. This resulted in the the size of the favicon being slightly reduced. I'm not sure if this was the correct decision and therefore am requesting feedback. 

Screen Shot 2020-08-17 at 3 12 28 PM

Screen Shot 2020-08-17 at 1 44 06 PM

I also thought I would experiment with adding support for dark mode via `prefers-color-scheme`:

Screen Shot 2020-08-17 at 12 43 34 PM

Screen Shot 2020-08-17 at 12 43 42 PM

Since the "dark mode" favicon's fill color is white, and is presented on a white backdrop in Storybook, I added a `<style>` block with prefers-color-scheme media queries to the twig story file that will also toggle its color, which makes the white favicon visible when dark mode is selected.

## Testing

1. View [deploy preview](https://deploy-preview-908--cloudfour-patterns.netlify.app/?path=/docs/design-brand--favicon)
1. Provide feedback on size of favicon
1. Toggle dark mode on your system and confirm prefers-color-scheme works as expected
1. Provide feedback on the decision to add dark mode support and the color selections presented in this PR
1. If appropriate, critique and provide feedback for how I modified the story backdrop color to demonstrate the dark mode favicon
1. Provide code review

---

see: issue #746


src/design/demo/favicon-svg.twig
```diff
@@ -0,0 +1,15 @@
...
+  @media (prefers-color-scheme: light) {
+    #favicon-demoWrapper {
+      background-color: white;
+    }
+  }
+  @media (prefers-color-scheme: dark) {
+    #favicon-demoWrapper {
+      background-color: black;
+    }
+  }
```
tylersticka, line 11: Can we make the icon preview smaller so it's easier to tell how it will look in a tab?

src/design/demo/favicon-svg.twig
```diff
@@ -0,0 +1,15 @@
+<style>
```
tylersticka, line 1: Instead of including these with an inline style, can we include a `demo/styles.scss` file similar to other demo-specific styles?

src/design/demo/favicon-svg.twig
```diff
@@ -0,0 +1,15 @@
...
+  @media (prefers-color-scheme: light) {
+    #favicon-demoWrapper {
+      background-color: white;
+    }
+  }
+  @media (prefers-color-scheme: dark) {
+    #favicon-demoWrapper {
+      background-color: black;
```
tylersticka, line 9: Can we use the same dark BG as our theme add-on configuration does?

src/design/brand.stories.mdx
```diff
@@ -14,3 +15,11 @@ Cloud Four logo in SVG format. The logo fill color can be modified via the CSS
 <Preview>
   <Story name="Logo">{brandLogo}</Story>
 </Preview>
+
+# Favicons
+
+Cloud Four favicon in SVG format with dark mode support via `prefers-color-scheme`. Toggle dark mode in system settings to view alternate color mode.
```
tylersticka, line 21: Is there an opportunity to credit the CSS-Tricks article you found helpful somewhere in this documentation?

src/assets/brand/favicon.svg
```diff
@@ -0,0 +1,22 @@
...
+      @media (prefers-color-scheme: light) {
+        .c-favicon {
+          fill: #215cca;
+        }
+      }
+      @media (prefers-color-scheme: dark) {
+        .c-favicon {
+          fill: #fff;
+        }
+      }
```
tylersticka, line 16: This isn't a component, so we probably prefix its class with `c-`?

derekshirk: @tylersticka 

> Do you think there's anything we can do to make the icon work better in browsers where light mode is selected but the inactive tabs are darker?
> 
> Screen Shot 2020-08-17 at 4 28 01 PM

That is an excellent question. The first thing that comes to mind is to explore using `mix-blend-mode`. I will test this out and as well as review other possible solutions and their outcomes.

Here's a non-favicon example of `mix-blend-mode: luminosity;` 

mix-blend-mode

derekshirk:
> Do you think there's anything we can do to make the icon work better in browsers where light mode is selected but the inactive tabs are darker?
> 
> Screen Shot 2020-08-17 at 4 28 01 PM

That is an excellent question. The first thing that comes to mind is to explore using `mix-blend-mode`. I will test this out and as well as review other possible solutions (filters perhaps 🤔) and their outcomes.

Here's a non-favicon example of `mix-blend-mode: luminosity;` 

mix-blend-mode

derekshirk:
@tylersticka, 

My experiments with mix-blend-modes did not prove to be successful. 😞 Presumably because the way favicons are implemented in the browser is different from how stacked HTML elements are rendered, which allows mix-blend-modes to work. 🤔 

This makes me think other possible solutions might involve: 

- Moving away from our primary brand color to a color that works better in both (light mode) active and inactive tabs. 🤔 
- Consider placing a script inside the SVG (or elsewhere) that could detect if the tab/window is active or inactive. Upon a few quick searches it looks like there might be a Page Visibility API capable of doing just that: https://stackoverflow.com/questions/1060008/is-there-a-way-to-detect-if-a-browser-window-is-not-currently-active/1060034#1060034 . This seems like something that likely has performance considerations. 

Do _you_ have any thoughts or ideas?

This is a problem I would really love to solve! However, if a "simple" solution does not present itself, I wonder if this is in scope for this issue/PR and perhaps creating a separate issue might be more appropriate. Let me know what you think.

derekshirk:
I like that idea. I will work on implementing that and will communicate if I think of anything else along these lines.

PR #966, opened 2020-09-29
Switch from storybook knobs to controls

### LINKED ISSUE #947: Migrate from storybook knobs to controls
https://medium.com/storybookjs/storybook-controls-ce82af93e430

https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#how-do-i-migrate-from-addon-knobs

After making this change, we should re-evaluate this issue: https://github.com/cloudfour/cloudfour.com-patterns/issues/556

calebeby:
## Overview

Closes #947 

It is worth testing the behavior of all the stories I updated. I updated them by hand and there are likely places where the behavior is not exactly the way we want.

## Testing

1. Check the deploy preview, check each story that changed


tylersticka (commented):
Reviewing to stop reminders until Netlify deploy failure has been resolved. Please request a re-review once all the status checks pass. Thanks!

tylersticka:
Some options that were using `select` look a little off with radios. Do we have any other options for this sort of thing?

Example before:

Screen Shot 2020-09-29 at 4 24 47 PM

And after:

Screen Shot 2020-09-29 at 4 24 22 PM

src/objects/media/media.stories.mdx
```diff
@@ -41,13 +50,23 @@ This pattern is composed of three elements:
...
+        defaultValue: meganProfileImage,
+      },
+      text: { type: { name: 'string' }, defaultValue: exampleText },
+      imgSize: {
+        type: { name: 'number' },
+        control: imgSizeOptions,
+        defaultValue: 64,
+      },
+      reverse: { type: { name: 'boolean' }, defaultValue: true },
+    }}
```
tylersticka, line 67: Is there a way using Controls that we can cut down on this repetition and just show a different version of the first story with `reverse` set to `true`?

Or alternatively, could we expose the list of controls in the Docs view so we don't need to manually show that example, but have a UI for toggling that option?

tylersticka (commented):
Overall, Caleb, this is awesome. I have some high-level questions in here, but just wanted to leave a note letting you know I'm very pleased with the progress overall.

calebeby:
@tylersticka I can switch radio buttons back to select. Should I switch just the ones where the empty value makes the radio button off center, or should I switch _all_ of the radios back to selects?

src/objects/media/media.stories.mdx
calebeby: Is this solution fine? [`2de2f0f` (#966)](https://github.com/cloudfour/cloudfour.com-patterns/pull/966/commits/2de2f0f3d2ee6c030a71ffca27aa08bd8e9d5edd)

tylersticka:
> I can switch radio buttons back to select. Should I switch just the ones where the empty value makes the radio button off center, or should I switch _all_ of the radios back to selects?

I haven't reviewed every control, but the ones with empty values or more than 3 or 4 values should probably use a `select`.

src/objects/media/media.stories.mdx
tylersticka: @calebeby Yeah, that's way better!

calebeby:
@tylersticka I went through all the radios/selects and set them based on the # of items and whether they had empty values ✅

tylersticka (commented):
@calebeby I think we may have some options for further simplifying the definition of controls.

You can avoid associating `argTypes` with every individual story by assigning it to the `Meta` element at the beginning of the file:

```jsx
<Meta title="Components/Button" argTypes={argTypes} />
```

You can then limit usage of `argTypes` on `Story` elements to cases where you want to _override_ those for some reason.

We can also simplify variations by using the `args` property rather than modifying and passing along a unique `argTypes` value. Example:

```jsx
// ...

<Meta title="Objects/Media" argTypes={argTypes} />

// ...

<Canvas>
  <Story name="Image">{(args) => imageDemo(args)}</Story>
</Canvas>

// ...

<Canvas>
  <Story name="Image Reversed" args={{ reverse: true }}>
    {(args) => imageDemo(args)}
  </Story>
</Canvas>
```

calebeby:
@tylersticka This is ready for another review! I switched many of the stories to use `argTypes` in `<Meta>`. Unfortunately, it seems like an individual story cannot _exclude_ any args from the `argTypes` defined in `<Meta>`, so in the cases where I needed to do that, I left `argTypes` on the individual stories.

src/components/message/message.stories.mdx
```diff
@@ -1,34 +1,30 @@
 import { Story, Canvas, Meta } from '@storybook/addon-docs/blocks';
-import { withKnobs, text, boolean } from '@storybook/addon-knobs';
 import messageDemo from './demo/default-message.twig';
-import messageWithCloseDemo from './demo/message-with-close.twig';
```
calebeby, line 4: @tylersticka I removed this since I could just use the default demo and pass in the `include_close` arg

tylersticka (approved):
This is awesome. Thanks, @calebeby!

PR #967, opened 2020-09-30
Replace Theo with Style Dictionary

### LINKED ISSUE #888: Display hex color values
On the [colors page in the pattern library](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/design-colors--page), it's only displaying rgb values. It would be great if it showed hex values.

spaceninja:
## Overview

This PR changes the tool we use to manage our design tokens from Theo to Style Dictionary. We're making this change because we've used it on several other projects and been impressed with the flexibility it gives us.

## Testing

Review the [preview environment](https://deploy-preview-967--cloudfour-patterns.netlify.app/?path=/story/getting-started--page), and confirm there are no visual changes other than the color values changing from RGB to hex values.

1. Check out the branch locally, run `npm ci` and `npm run storybook`
1. Try editing a design token and confirm that it updates in Storybook.

## Code Review

Apologies for the size of this changeset. You'll probably want to break your code review up by file type:

- SCSS files should only contain variable name updates
- YML files have all been deleted (those are the old token files)
- JSON files are mostly new design token files
- JS files are all either design token files or tooling to help build them
- MDX files are either new (replacing some previously auto-generated design token stories) or updates to how they consume tokens.

---

- Fixes #888


spaceninja:
Note there is one big of code that I'd like to DRY up — at the top of most of the design token story files (see `src/tokens/*.stories.mdx`) there's a repeated bit of code to generate a table of properties. This should ideally be pulled out into a separate file to be included or converted to a React component (like `ColorPalette`). Unfortunately, when I did so, I got errors about JSX that I wasn't able to quickly resolve. Since I'm being pulled back to client work tomorrow and this PR is already pretty big, I decided to leave the code as-is for now. Assuming @tylersticka is okay with it, I'll file a followup ticket to come back and DRY that up.

(Note if someone else is interested, the "definitive" version of that code can be found in `font.stories.mdx`.)

spaceninja:
@calebeby I tagged you in this as an FYI since we were talking about the JSX code not working. Don't feel obligated to review, though.

build-tokens.js
```diff
@@ -0,0 +1,78 @@
+const StyleDictionary = require('style-dictionary');
+
+const styleDictionaryConfig = {
+  log: 'warn',
+  source: ['src/tokens/**/*.+(js|json)'],
+  platforms: {
+    scss: {
+      transformGroup: 'custom/transform-group/css',
+      buildPath: 'src/compiled/scss/',
```
tylersticka, line 9: I don't care about the change from "generated" to "compiled," but why remove tokens from the path?

build-tokens.js
```diff
@@ -0,0 +1,78 @@
+const StyleDictionary = require('style-dictionary');
```
tylersticka, line 1: This is a bit of a nitpick, but it feels odd to me to have this just hanging out in the root. Maybe it's because I remember our Theo config starting out that way and then eventually needing its own directory.

Could we potentially follow [Style Dictionary's example](https://github.com/amzn/style-dictionary/tree/main/examples/advanced/custom-transforms) and create a shared directory that includes a `config.json` file for the simple/default options and a `build.js` script for the rest?

package.json
```diff
@@ -62,6 +62,7 @@
     "@wordpress/block-library": "2.24.0",
     "babel-loader": "8.1.0",
     "browserify-transform-tools": "1.7.0",
+    "chokidar-cli": "^2.1.0",
```
tylersticka, line 65: It feels off to me that we have Gulp for managing and watching tasks, but instead we're using a bare `.js` file that we're running via `node`, and now we're adding the Gulp watch dependency so we can configure that separately.

Can we not manage the Style Dictionary build script as a Gulp task and then the watch command from the same place as everything else?

package.json
```diff
@@ -72,11 +73,9 @@
     "eslint-plugin-react": "7.20.6",
     "gulp": "4.0.2",
     "gulp-postcss": "8.0.0",
-    "gulp-prettier": "3.0.0",
```
tylersticka, line 75: Was this not being used or was it only used by Theo before?

src/base/_defaults.scss
```diff
@@ -53,15 +51,15 @@ ul {
 
 body {
   background-color: #fff; /* 1 */
-  color: colors.$text-dark; /* 2 */
+  color: tokens.$color-text-dark; /* 2 */
```
tylersticka, line 54: My least favorite thing about Style Dictionary in a recent client project is how every single token group is housed under a `tokens.` object. If there isn't a way to recreate this with Style Dictionary, I'm honestly tempted to say this transition isn't worth it right now. (That may seem dramatic, but it's really a pain to type dozens of times!)

tylersticka (changes requested):
It's nice that Style Dictionary doesn't require as much up-front logic as Theo did. But I'll be honest, this doesn't feel like an upgrade in its current state:

- As mentioned in the PR description, we're now requiring repetitive code across the story documentation that we weren't before. This will make it more difficult for contributors to make new token groups.
- Previously, you would `@import` a category once and then refer to it by that name. Now, every token is output to a single monolithic `tokens` object, increasing the number of characters you need to type to refer to them. This will also make looping over tokens more difficult than it would be otherwise, something we've had to hack around in a different project.
- Although we already have a dependency for defining JS build scripts (Gulp) and triggering it on file changes (also Gulp), this PR uses a bare Node script and a new Chokidar dependency. This feels like an artifact of a different project and confuses the organization of this specific project.

While I support this change long-term, my feelings haven't changed since Monday: If this will take more than a couple of hours to resolve, it's probably best to create an issue and return to it after more of our user-facing tasks have been addressed.

build-tokens.js
spaceninja: Good idea!

build-tokens.js
spaceninja: Sure, I can add tokens back to the path

package.json
spaceninja: It was only used by Theo

src/base/_defaults.scss
spaceninja: That makes sense, and it's absolutely something we can address by generating some useful subsets of tokens. I'll talk to you about what subsets would be most useful.

package.json
spaceninja: Sorry about that, I was under the impression that we were trying to move away from Gulp toward npm scripts when possible. I'll look into how we can use Gulp to manage this.

spaceninja:
The danger of walking away from a PR of this scope for six months is that the merge conflicts when you come back will take a full day to reconcile. I've now got this PR updated and I'll tackle the changes @tylersticka requested shortly.

spaceninja:
@tylersticka I've addressed your feedback, and I think you'll be much happier with the current state of this PR, especially the new token groupings, which allows for using the Sass tokens in almost exactly the same way as before.

The Sass `@use` groups are manually configured, but adding new ones is pretty easy if needed. Also, the previous `tokens` file is still generated, so if a designer ever feels slowed by the process, they can still access new tokens that way, and file a ticket for a dev to create new groups if needed.

The one thing I'm unhappy with in this PR is the Story files for the tokens. Previously, these were automatically generated by Theo. Now they're manually created, and still have the duplicate code issue.

However, I think the rest of the changes are beneficial enough to warrant merging this as-is and filing a new issue to address that. It should be possible to generate the Story files with Style Dictionary, but this PR is already quite large. Unless you see any reason we shouldn't move forward with Style Dictionary, I'd like to merge this and start working on that ticket as a follow-up.

Let me know if you'd like to talk about this PR with voices.

CC'ing @Paul-Hebert as a reviewer since he's had hands on Style Dictionary recently for a client project and may have insight or feedback.

tylersticka (commented):
This is awesome. I love the changes to naming conventions, the config necessary seems intuitive, I dig the use of the modular functions instead of custom transform types, and overall the outputs seem like the best of both worlds now.

The only thing stopping me from marking "Approve" is that I have not had time to do a pass for regressions in the components. If someone else is able to do so, that would be swell!

Paul-Hebert:
I should be able to do a regression pass this afternoon :+1:

Paul-Hebert:
Doing a regression pass now (not planning to review the code since it sounds like Tyler has already but if you'd like code review as well lemme know)

Paul-Hebert (commented):
@spaceninja I did a regression test and I spotted no differences on components, utilities, etc.

I did notice that for some reason on the preview deploy the Prism docs page is missing. I also see an error in the console:

>Unexpected error while loading ./vendor/prism/prism.stories.mdx: ReferenceError: Prism is not defined

I'm not sure what the deal is... I don't see any obvious changes that would have broken this but I also didn't do a deep dive on the code. Any idea what's going on there?

spaceninja:
Okay, I fixed the Prism error. No idea why that broke on my branch, but the fix was simple enough.

Paul-Hebert (approved):
🚢

PR #988, opened 2020-10-23
Only use svgr in specific contexts

tylersticka:
## Overview

I started making an unrelated change to a pattern, but on every save I saw errors about SVG loading in the Terminal. These appear to be uncaught issues with changes from #862, which caused the icon listing to fail (which explains its disappearance from [our deploy](https://v-next--cloudfour-patterns.netlify.app/)).

I tried a few different solutions, but there were three points of Webpack configuration around `.svg` files and whenever I fixed _one_ a different one would have issues.

I looked back at the history of _why_ there was so much SVG configuration, and realized it was all due to #503, specifically the need to convert SVG files to React components for the sole purpose of documentation.

Since that's the _only_ context where we'd want to do that, I decided to use inline loaders within that file instead. While that's not a great practice for an entire project, it keeps this complexity confined to the single file that needs it. 

## Screenshots

Screen Shot 2020-10-23 at 10 54 52 AM

## Testing

On [the deploy preview](https://5f931c8716bbef00073f5e90--cloudfour-patterns.netlify.app/)...

1. Observe that [the icon listing](https://5f931c8716bbef00073f5e90--cloudfour-patterns.netlify.app/?path=/story/design-icons--page) is displayed correctly.
1. Confirm that Twig inlined SVGs haven't been affected in [examples like this](https://5f931c8716bbef00073f5e90--cloudfour-patterns.netlify.app/?path=/story/design-brand--logo).
1. Confirm that CSS inlined SVGs haven't been affected by inspecting [elements like this](https://5f931c8716bbef00073f5e90--cloudfour-patterns.netlify.app/?path=/story/components-input--select-element).
1. Confirm that SVGs in [prototypes like this](https://5f931c8716bbef00073f5e90--cloudfour-patterns.netlify.app/?path=/story/prototypes-clouds--example) haven't been affected.

---

/CC @spaceninja @calebeby

### DIFF
--- a/.changeset/cuddly-kids-taste.md
+++ b/.changeset/cuddly-kids-taste.md
@@ -1,5 +1,0 @@
----
-'@cloudfour/patterns': patch
----
-
-Fix for issues related to SVG loading
--- a/.changeset/cuddly-kids-taste.md
+++ b/.changeset/cuddly-kids-taste.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Fix for issues related to SVG loading
--- a/src/design/icons/icons.stories.mdx
+++ b/src/design/icons/icons.stories.mdx
@@ -2,7 +2,11 @@ import { Meta, IconGallery, IconItem } from '@storybook/addon-docs/blocks';
 import { extname, relative } from 'path';
 // Import whole directory of icons together
 // @see https://webpack.js.org/guides/dependency-management/#context-module-api
-const iconDir = require.context('../../assets/icons', false, /\.svg$/);
+const iconDir = require.context(
+  '!!@svgr/webpack!../../assets/icons',
+  false,
+  /\.svg$/
+);
 const iconElements = iconDir.keys().map((key) => {
   const Icon = iconDir(key).default;
   // Retains subdirectory if we want to organize with those in the future
@@ -14,7 +18,7 @@ const iconElements = iconDir.keys().map((key) => {
   );
 });
 const brandIconDir = require.context(
-  '../../assets/icons/brands',
+  '!!@svgr/webpack!../../assets/icons/brands',
   true,
   /\.svg$/
 );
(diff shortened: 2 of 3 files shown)


src/design/icons/icons.stories.mdx
```diff
@@ -2,7 +2,11 @@ import { Meta, IconGallery, IconItem } from '@storybook/addon-docs/blocks';
 import { extname, relative } from 'path';
 // Import whole directory of icons together
 // @see https://webpack.js.org/guides/dependency-management/#context-module-api
-const iconDir = require.context('../../assets/icons', false, /\.svg$/);
+const iconDir = require.context(
+  '!!@svgr/webpack!../../assets/icons',
```
spaceninja, line 6: Could you add a comment explaining this svgr syntax, or linking to some docs?

spaceninja (commented):
Sounds like a great idea to me, and I'm very much in favor of reducing the complexity of our app. I'll leave final approval to @calebeby since this touches webpack, but it LGTM.

tylersticka:
FYI, no need for anyone to review till I've marked as "ready for review" and assigned a reviewer. Was waiting till the deploy preview finished so I could make the testing instructions easier.

src/design/icons/icons.stories.mdx
tylersticka: Sure, this is more webpack syntax than svgr but I've linked to docs for both now.

.changeset/cuddly-kids-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Fix for issues related to SVG loading
```
calebeby, line 5: I _think_ we don't need a changelog entry for this because it only affects the storybook site, not the npm package

tylersticka: All righty, removed the changeset!

calebeby (approved):
🎉 Thanks for catching that!

PR #989, opened 2020-10-23
Run PostCSS in buildSass

tylersticka:
## Overview

No inlined SVGs in CSS files were included in the output CSS. This is because we weren't running PostCSS outside of Webpack except for minification.

This adds PostCSS as a step in our `buildSass` task.

## Testing

1. Observe [the latest version of `standalone.css`](https://unpkg.com/@cloudfour/patterns@0.3.0/dist/standalone.css), confirm that it includes uncompiled `svg-load()` functions in its output (which it should not).
1. Check out branch locally.
1. Run `npm run build:sass`
1. Confirm that `dist/standalone.css` and `dist/standalone.min.css` include `url("data:image/svg+xml;...")` strings instead of `svg-load()` functions.

### DIFF
--- a/.changeset/green-grapes-unite.md
+++ b/.changeset/green-grapes-unite.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Fix for missing SVGs in compiled CSS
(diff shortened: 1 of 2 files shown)


calebeby (commented):
🎉 

FWIW this will trigger the release action for the first time in a while. I am curious to see whether the change in https://github.com/cloudfour/cloudfour.com-patterns/pull/949 will have fixed it (my guess is that it will still fail, but it will be good to confirm that)

calebeby (approved):

PR #990, opened 2020-10-23
Change attribute for next/prev pagination check

tylersticka:
## Overview

We based the pagination component template's properties on [Timber's pagination object](https://timber.github.io/docs/guides/pagination/), but mistakenly assumed `pagination.prev` and `pagination.next` mimicked individual `pagination.pages` entries exactly. They do not, and unfortunately they lack the `name` key we were using to detect when the current page matched the previous or next.

This PR swaps the `name` property for `link`, which _does_ exist in both. I also tested this in a WordPress theme to make sure it worked.

## Testing

Compare [deploy preview](https://deploy-preview-990--cloudfour-patterns.netlify.app/?path=/story/components-pagination--example) with [the previous iteration](https://v-next--cloudfour-patterns.netlify.app/?path=/story/components-pagination--example), they should behave the same when controls are changed.


calebeby (approved):
LGTM. I think it would be best to merge https://github.com/cloudfour/cloudfour.com-patterns/pull/992 first so that we can make sure the new release flow is working correctly

PR #1005, opened 2020-10-30
Support non-link pagination items

tylersticka:
## Overview

It's possible in WordPress or Timber through certain configurations to output pagination items without any links, for example a "…" item preceding a jump to the very last page number. This PR prevents those actions from outputting empty links anyway.

## Testing

Review [the pagination component in the deploy preview](https://deploy-preview-1005--cloudfour-patterns.netlify.app/?path=/story/components-pagination--example). It should work the same as before.


Paul-Hebert (approved):
Looks good!

PR #1016, opened 2020-11-06
Update Elastic Textarea JS

spaceninja:
## Overview

We made some updates to this code in another project that's using it,
so this commit is updating this version to match.

## Testing

Review the [elastic textarea pattern](https://deploy-preview-1016--cloudfour-patterns.netlify.app/?path=/docs/components-input--elastic-textarea). There should be no visible or functional changes.

### DIFF
--- a/.changeset/silent-planes-boil.md
+++ b/.changeset/silent-planes-boil.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Code refactor of the elastic textarea javascript results in a slight performance improvement for IE11.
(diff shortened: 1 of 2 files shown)


tylersticka (commented):
@spaceninja Should this include a changeset since it includes a user-facing performance improvement for old browsers?

Paul-Hebert:
Code looks good besides Tyler's comment. 

Obviously this is not required and you're already going above and beyond by making these change, but this could make a nice npm package for easy reuse in the future! I searched npm and I like this better than any of the results I found

tylersticka (approved):

PR #1043, opened 2020-11-17
Page object

### LINKED ISSUE #742: Overall page layout with stretchy main area
On our current site, the default page color is silver. This means that if the page is very short, the footer will appear to stretch to meet the bottom of the viewport:

Screenshot_2020-05-26 Page not found - Cloud Four

But ideally, I would prefer for the background color to stay white and for the main content area to be the thing that stretches:

stretchy

Not only would this make our CSS easier to use in prototypes and such (since it wouldn't the background to silver), but I think it looks more balanced to not increase the footer area.

Here's a good place to start in terms of different techniques: https://css-tricks.com/couple-takes-sticky-footer/

And whoever takes on this issue might want to chat with @spaceninja about some challenges we had maintaining this behavior in a recent Vue project.

tylersticka:
## Overview

Adds an `o-page` object for defining the main page content versus the footer content, which allows the footer to "stick" to the bottom of the viewport on shorter pages.

Storybook injects some HTML that works against the demo, but I think I figured out how to override it for this example via `useEffect`.

I chose to use CSS Grid because it required less CSS. I think it's fine if this doesn't work in unsupported browsers, it just means the footer will directly follow the content (as it does today).

## Screenshots

Screen Shot 2020-11-17 at 3 52 51 PM

---

- Fixes #742

### DIFF
--- a/src/objects/page/page.scss
+++ b/src/objects/page/page.scss
@@ -1,0 +1,29 @@
+/**
+ * Page container
+ *
+ * Attempts to fill the available area (assuming `height: 100%` is correctly
+ * applied to all parent elements) and keeps the footer at the bottom of that
+ * area.
+ *
+ * @see https://css-tricks.com/couple-takes-sticky-footer/#there-is-grid
+ */
+
+.o-page {
+  display: grid;
+  grid-template-rows: 1fr auto;
+  min-height: 100%;
+}
+
+/**
+ * These classes aren't strictly required, but they keep the HTML source a
+ * little easier to read while also supporting swapping of source order if
+ * that's ever something that content calls for.
+ */
+
+.o-page__content {
+  grid-row: 1;
+}
+
+.o-page__footer {
+  grid-row: 2;
+}
--- a/.changeset/wicked-experts-listen.md
+++ b/.changeset/wicked-experts-listen.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add Page object for "sticky" page footers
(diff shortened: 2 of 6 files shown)


src/objects/page/page.scss
```diff
@@ -0,0 +1,29 @@
...
+ * Attempts to fill the available area (assuming `height: 100%` is correctly
+ * applied to all parent elements) and keeps the footer at the bottom of that
+ * area.
+ *
+ * @see https://css-tricks.com/couple-takes-sticky-footer/#there-is-grid
+ */
+
+.o-page {
+  display: grid;
+  grid-template-rows: 1fr auto;
```
spaceninja, line 13: Should we consider wrapping this layout in a min-height media query here? That would prevent the footer from obliterating the page content on devices with short screens, such as an iPhone in horizontal mode.

spaceninja: Wait, no, I misunderstood. you set a min-height here, which means it's allowed to grow beyond the viewport. Sorry, my bad!

tylersticka: I don't understand. Can you share a screenshot of the behavior you're describing?

When there isn't enough room for the body content and footer, the content will scroll as it normally would.

tylersticka: o-page-behavior

spaceninja (approved):
Tested and works as described.

Does our naming convention use `l-` for layout components, or just `o-`? I forget which flavor of BEM we're using.

tylersticka:
> Does our naming convention use `l-` for layout components, or just `o-`? I forget which flavor of BEM we're using.

We're trying out [BEMIT](https://csswizardry.com/2015/08/bemit-taking-the-bem-naming-convention-a-step-further/), which uses `o-` for "layout objects."

PR #1056, opened 2020-11-25
Revise theme strategy to use custom properties

tylersticka:
## Overview

Before this PR, our themes were powered entirely by the CSS cascade:

```css
.example {
  color: black;
}

.t-dark .example {
  color: white;
}
```

This works pretty well, _until_ you want to nest one themed pattern in another... something we'd probably want to do in a case like [PWA Stats' cards](https://www.pwastats.com/). In order to do that, you end up needing to double up your default theme's selectors so there's a mechanism for opting back into them:

```css
/* for every single themed class... */

.example,
.t-light .example {
  color: black;
}

.t-dark .example {
  color: white;
}
```

Which may not seem like a big deal for one class, but it can get pretty unwieldy across a system.

CSS properties make this a lot simpler:

```css
/* theme properties defined once */

:root,
.t-light {
  --color-example: black;
}

.t-dark {
  --color-example: white;
}

/* then referenced once in classes */

.example {
  color: var(--color-example);
}
```

This is easier to maintain, and the eventual CSS bundles are smaller since it takes less code to write.

## Screenshots

Screenshot_2020-11-24 Storybook

## Known issues

- I suspect these aren't 100% the right theme properties yet. I sort of gave them a [Style Dictionary](https://amzn.github.io/style-dictionary/)-esque structure, which isn't the token library we're using, and there's a lot of redundancy because I kept getting hung up on finding the best _shared_ names for things. Eventually I decided I'd wasted too much time deliberating and I should just create properties based on what we were using so far and think about simplifying, consolidating or harmonizing later on.
- It's possible that these theme values should be stored as tokens, but since we have some outstanding discussions about the token library we're using, I didn't want to get hung up on that yet, either.
- There were a handful of places where I couldn't make things work with custom properties alone or felt a little stymied about whether they should support themes at all. I tried to comment these clearly.
- There aren't any fallbacks yet for browsers that don't support custom properties. There's a comment about that in the docs, but if we think that's a blocker for moving forward, we can stop and address it now.
- I wanted the new theme demo to include a `Card` for the nested theme example since that's what originally inspired it, but there aren't great modifiers for that yet. There were already comments about that in the Card docs, but I added one to the Twig template for the theme demo as well.

## Testing

Compare [the deploy preview](https://deploy-preview-1056--cloudfour-patterns.netlify.app/) to [the `v-next` preview](https://v-next--cloudfour-patterns.netlify.app/) for regressions. The themes have [their own documentation](https://deploy-preview-1056--cloudfour-patterns.netlify.app/?path=/docs/design-themes--light), but there is also a toolbar button to toggle the canvas view's theme.

Components with theme-specific styles include:
- Button
- Calendar Date
- Checkbox
- Dot Leader
- Input
- Pagination
- Sky Nav


src/components/checkbox/checkbox.scss
```diff
@@ -125,21 +121,22 @@
...
-       */
-
-      .t-dark & {
-        color: colors.$text-light;
-      }
-
       &::after {
-        background-image: svg-load('icons/check.svg', stroke=colors.$gray-dark);
+        /**
+         * Inline SVGs aren't aware of custom properties, so we have to do a bit
```
Paul-Hebert, line 129: Nit: I was a little confused by this comment since I thought you meant inlining the SVG code in HTML, not using an SVG as a background image.

src/components/pagination/pagination.scss
```diff
@@ -167,12 +172,21 @@ $indicator-width-current: $indicator-height-current;
...
+   */
+
+  &,
+  .t-light & {
+    background-image: svg-load(
+      'icons/arrow-right.svg',
+      stroke=colors.$primary-brand
+    );
+  }
+
```
Paul-Hebert, line 189: Nit: this pattern has been used three times and will likely be used more. I wonder if it would be worth wrapping in a mixin for easier use? (Otherwise I'll be searching these out and copy/pasting them whenever I need to do something similar)

src/mixins/_theme.scss
```diff
@@ -0,0 +1,86 @@
+@use "../design-tokens/colors.yml";
+@use "../design-tokens/opacity.yml";
+@use "../design-tokens/sizes.yml";
+
+/// Establish the minimal rules needed for a theme's root element, specifically
+/// a `background-color` and default text color. Without this, the theme root
+/// won't be visually offset from its parent and its contents may not be very
+/// readable.
+///
+/// @output `background-color` and `color` rules
```
Paul-Hebert, line 10: Nit: I'm not used to this comment style. Is this an SCSS convention or something?

Paul-Hebert (commented):
This looks good @tylersticka! I'm definitely a fan of using custom props for this, and I'm cool with most of the known issues. My only hesitation around this setup is where we are defining component-specific theme properties. 

Right now all of our theme properties are defined in `_theme.scss`. I think this makes sense for global theme properties like text color, but when it comes to component-specific properties it seems like it could be inconvenient. You'd need to be hopping back and forth between `_theme.scss` and your component SCSS a lot. I wonder if these properties like these would be easier to maintain in component CSS?

- `--theme-color-background-pagination-control`
- `--theme-color-border-button-default`
- `--theme-color-border-button-secondary`
- `--theme-color-border-calendar-date`
- `--theme-color-border-input-disabled`

So each component would look like the example you put in your PR description:

```css
:root,
.t-light {
  --color-example: black;
}

.t-dark {
  --color-example: white;
}

/* then referenced once in classes */

.example {
  color: var(--color-example);
}
```

This is a little more verbose, but everything specific to a component's styles would be in one file, instead of spread across a few. You also mentioned this:

> and there's a lot of redundancy because I kept getting hung up on finding the best shared names for things. Eventually I decided I'd wasted too much time deliberating and I should just create properties based on what we were using so far and think about simplifying, consolidating or harmonizing later on.

If we're planning on replacing these component specific properties with more consolidated theme properties in the future this may not be a big deal. That said, I tried implementing a similar theming strategy on my portfolio site and found it difficult to strike the right balance with the consolidated properties. I ended up wanting to use 1-off theme properties for lots of components and found it awkward that all my properties lived in a `_theme.scss` file.

---

Anyways, I'm curious what you and others think! This isn't a huge deal, and I could be convinced the current setup is the best way to go, but I figured it would be worth discussing a bit before we move forward. Thanks!

src/mixins/_theme.scss
tylersticka: 'Tis SassDoc! http://sassdoc.com/

src/components/pagination/pagination.scss
tylersticka: Hmm, I'm torn on this.

On one hand, I see what you mean. We're using this pattern a few different times.

On the other hand, I sort of _want_ to discourage this method of theming over the user of custom properties. And the only part we'd really be keeping DRY is `&, .t-light &`, which feels leaner than any mixin name might be.

Paul-Hebert: That's fair. I don't feel strongly about this.

src/mixins/_theme.scss
Paul-Hebert: Gotcha, thanks!

tylersticka:
> My only hesitation around this setup is where we are defining component-specific theme properties.
> 
> Right now all of our theme properties are defined in `_theme.scss`. I think this makes sense for global theme properties like text color, but when it comes to component-specific properties it seems like it could be inconvenient. You'd need to be hopping back and forth between `_theme.scss` and your component SCSS a lot. I wonder if these properties like these would be easier to maintain in component CSS?
> 
> [...]
> 
> Anyways, I'm curious what you and others think! This isn't a huge deal, and I could be convinced the current setup is the best way to go, but I figured it would be worth discussing a bit before we move forward. Thanks!

That makes sense, and I appreciate the suggestion.

I'm torn on this suggestion and I'm having trouble articulating why. It makes sense for component-specific properties to be with the component, but will that always be the case if we're able to use more generic property names? Or maybe it will eventually be a mix of both, with generic properties in one file and component-specific properties in another? Would this all be moot if these values were maintained as tokens instead of in an SCSS file? Which of these questions should we answer now versus later?

I'm also interested in what others think, especially @spaceninja and @gerardo-rodriguez.

Paul-Hebert:
>Or maybe it will eventually be a mix of both, with generic properties in one file and component-specific properties in another?

I think this would be my dream setup but am open to other opinions. 

>Would this all be moot if these values were maintained as tokens instead of in an SCSS file? Is there value in separating to reinforce the idea that certain properties are deferring to the theme, effectively out of the component's hands? But then again, keeping those properties with the component means outputting less CSS if we ever decided to modularize bundles... but would we ever do that? Which of these questions should we answer now versus later?

These are all good questions! This PR is definitely a step in the right direction, so I'm fine if we punt on some of these questions for now and circle back as we're further along. I'm also curious what others think!

spaceninja:
> I'm torn on this question and I'm having trouble articulating why. It makes sense for component-specific properties to be with the component, but will that always be the case if we're able to use more generic property names? Or maybe it will eventually be a mix of both, with generic properties in one file and component-specific properties in another? Would this all be moot if these values were maintained as tokens instead of in an SCSS file? Is there value in separating to reinforce the idea that certain properties are deferring to the theme, effectively out of the component's hands? But then again, keeping those properties with the component means outputting less CSS if we ever decided to modularize bundles... but would we ever do that? Which of these questions should we answer now versus later?

Before I chime in, I want to stress that I have no really strong feelings on this topic, and will happily roll with whatever the group decides.

Back in the day, we used to keep every Sass variable in a single file called `_vars.scss`. This made it really easy to find variables, but the file eventually became overwhelming as more and more things became variables.

The next step was to keep "global" variables in the `_vars.scss` file and move component-specific variables into the component's stylesheet. This had the advantage of trimming down the `_vars.scss` file, and "scoping" our component variables, in that they literally weren't available outside the component, which eliminated a lot of "component A secretly relies on variables from component B so you can never delete component B".

When switching from Sass variables to custom properties, I maintained this approach and have found it continues to work well.

Mixing Sass variables and custom properties also became a thing, and in hindsight, what we were doing was a low-budget design token system, where sass variables were for the "atomic" values, like specific color hexcodes, common sizes, etc., and custom properties were used for "semantic" values, eg: `--button-primary-color: #{$red};`

Now, with design tokens, we've got to consider not only how our sass vars/custom properties are organized and kept in the filesystem, but also how to group our design tokens and when/if to break those files apart.

Not to mention what belongs in a design token vs a custom property, or whether there should be any custom properties that arent' design tokens.

Long story short, I don't think there's a settled best practice for this. I see Tyler's concerns and understand them. I tend to lean towards Paul's opinion, advocating for component-related things to live in component-related files, but I don't think that's a hard and fast rule. There may be exceptions or edge cases I haven't considered.

So, I think my takeaway is, all things being equal:

- "global" or "atomic" or whatever you want to call them values should be stored in design tokens or a global Sass variables file.
- "semantic" values should be stored as close to where they will be consumed as possible, ideally as custom properties. Ideally, your semantic variables roughly align with your components, in which case they should probably live with components.
- If you have multiple components that share the same value, they should both have a scoped semantic custom properties that consume the same global design token or sass value.

Those are kind of my rules of thumb, but I think that at Cloud Four we tend to push the boundaries of what design systems can do, so I'm 100% open to the idea that there are situations that those rules of thumb don't cover or don't make sense for.

Hopefully that was helpful rather than just muddying the water!

spaceninja:
Just realized I didn't actually answer these questions:

> It makes sense for component-specific properties to be with the component, but will that always be the case if we're able to use more generic property names?

I think you'll likely end up needing both generic and specific values.

> Or maybe it will eventually be a mix of both, with generic properties in one file and component-specific properties in another?

Yeah, that's my preference. It's like design tokens without design tokens.

> Would this all be moot if these values were maintained as tokens instead of in an SCSS file?

Not moot, but design tokens _definitely_ push you in the direction of "tokens are global/generic" vs "component variables are specific/semantic"

> Is there value in separating to reinforce the idea that certain properties are deferring to the theme, effectively out of the component's hands?

I think I would want to see a component variable that is set to the theme value, something like this:

```css
// in global/theme stylesheets:
:root { --color-heading: black }
.dark { --color-heading: white }
h1 { color: var(--color-heading) }

// in component stylesheet:
.card__heading { color: var(--color-heading) }
```

> But then again, keeping those properties with the component means outputting less CSS if we ever decided to modularize bundles... but would we ever do that?

Great question! Premature optimization is not ideal. I don't think we have ever modularized bundles including CSS and I don't think it's on the horizon, so maybe we can punt on that for now?

> Which of these questions should we answer now versus later?

I appreciate that we're thinking about things and trying to make the best decision, but I think this might be one of those cases where we just don't have the information to know for a fact which approach is ideal. Let's set aside the idea of what the perfect solution is, and just pick something that works well enough that we don't hate it for now. Refactoring where variables are stored isn't a huge amount of work if/when we need to revisit this.

tylersticka:
Thanks, @spaceninja! Those are all very helpful thoughts.

To your last point, I think I have some ideas for changing things a bit without getting _too_ in the weeds on making everything Super Perfect™️ (as much as I like doing that). But if I hit a snag with that, I'll comment on this and be like "can we punt on this for now?"

tylersticka:
@Paul-Hebert @spaceninja Just pushed these revisions:

- There are now three theme mixins:
   - One for defining properties for a theme. If the theme is the default, these properties are output for `:root` as well as the theme class.
   - One for defining theme-specific styles, for example from a child element. This was a suggestion that @Paul-Hebert made before that didn't make sense to me until the other changes were made.
   - One for defining the basic theme rules. This was there before.
- Themed properties are no longer defined in centralized mixins, but instead in two places:
   - Broader, global or default properties are defined in `src/base/_themes.scss`.
   - Component-specific properties are defined in the same `.scss` file as the component itself.
- The theme documentation has been moved into the `design/` folder. This was already where it appeared in the Storybook nav, and it no longer felt right to have a dedicated folder with no real source files in it.

Let me know if we're getting colder or warmer!

src/mixins/_theme.scss
```diff
@@ -0,0 +1,51 @@
...
+  @if $theme == $default-theme {
+    :root {
+      @content;
+    }
+  }
+
+  .t-#{$theme} {
+    @content;
+  }
+}
```
Paul-Hebert, line 23: Nit: Would we output less compiled CSS if we combined these property sets similar to how you did in the mixin below?

```scss
  @if $theme == $default-theme {
    :root,
    .t-#{$theme} {
      @content;
    }
  } @else {
    .t-#{$theme} {
      @content;
    }
  }
```

Paul-Hebert (commented):
I definitely think we're getting warmer! 

It feels a lot cleaner to me to be defining component properties in component files. It took me a second to wrap my head around the mixin setup but after understanding it I think it abstracts out repeated markup nicely and should help maintainability.

I left one nit inline about one of the mixins but it looks good to me besides that. I'm holding off on giving it a :+1: for now because I'm curious what @spaceninja thinks.

src/mixins/_theme.scss
tylersticka: @Paul-Hebert When I did that, I got a `selector-root-no-composition` error from Stylelint. I assumed there was a reason for that and reverted to what I have now.

Because of your question, I attempted to look up _why_ that rule exists. I found that it's actually been removed from the most recent versions of Stylelint.

I found [this discussion](https://github.com/stylelint/stylelint/issues/2226) where it's suggested that it _may_ have existed to promote compatibility with PostCSS plugins for automating custom property fallbacks, which (at least at the time) only worked for `:root` selectors without more complex selectors.

So I guess I can probably combine these selectors again and disable the linting error there?

tylersticka: @Paul-Hebert Selectors are now combined and this rule has been disabled in the project's lint config

Paul-Hebert (approved):
Awesome, thanks @tylersticka , this looks good to me! :+1:

spaceninja (approved):
Great work, @tylersticka, I _really_ like this approach. I'm especially fond of how much simpler all the component stylesheets became once you could define the light and dark mode props at the top, and then then just update the style to consume the theme colors.

PR #1064, opened 2020-12-01
Add breakpoint helper mixins

tylersticka:
## Overview

I've been working on a new layout pattern with breakpoint modifiers and I found myself growing a bit tired of typing the same breakpoint class logic over and over. At one point I accidentally mismatched a breakpoint class name with its value (I typed `\@l` but I was still using `breakpoint.$m`) and decided it was time to make a robot do that busy work.

So this PR introduces a new `mixins/_media-query.scss` module. You can see the source for details, but the two most important mixins are `breakpoint-classes` and `breakpoint-parent-classes`.

### Before these mixins

```scss
@use 'path/to/design-tokens/breakpoint.yml';

.example--red {
  color: red;
}

.example--red\@m {
  @media (width >= breakpoint.$m) {
    color: red;
  }
}

.example--red\@l {
  @media (width >= breakpoint.$l) {
    color: red;
  }
}

.example__child {
  color: blue;

  .example--red & {
    color: red;
  }

  .example--red\@m & {
    @media (width >= breakpoint.$m) {
      color: red;
    }
  }

  .example--red\@l & {
    @media (width >= breakpoint.$l) {
      color: red;
    }
  }
}
```

### After these mixins

```scss
@use 'path/to/mixins/media-query';

.example--red {
  @include media-query.breakpoint-classes(
    $from: m, 
    $to: l
  ) {
    color: red;
  }
}

.example__child {
  color: blue;

  @include media-query.breakpoint-parent-classes(
    $selector: '.example--red',
    $from: m, 
    $to: l,
    $include-default: false
  ) {
    color: red;
  }
}
```

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -7,6 +7,7 @@
 @use '../../mixins/fluid';
 @use '../../mixins/focus';
 @use '../../mixins/headings';
+@use '../../mixins/media-query';
 @use '../../mixins/ms';
 
 /// We can't use `grid-gap` exclusively due to some containers only being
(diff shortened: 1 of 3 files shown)


spaceninja:
This is definitely clever, but I find myself hesitating.

I had to read the mixins twice before I grokked what you were doing.

Looking back at the code you're replacing, as far as I can tell the main problem you're trying to solve is when you need to apply the same code to a range of breakpoints.

In other frameworks that use breakpoint classes, I've seen this addressed by adding classes that span breakpoints, like `@m-xl`

Personally, I think that approach is more easily understandable when reading the code.

Is there a feature I'm missing that makes these mixins work better than defining some breakpoint spanning classes?

(To be clear, I'm not saying we can't or shouldn't approve this code, I just wanted to pump the brakes and make sure I'm understanding what your goal is)

spaceninja (approved):
Coolio — thanks for explaining! I'm happy to approve, but feel free to bounce it off some of the designers as well

PR #1065, opened 2020-12-03
Component: Event Log

tylersticka:
## Overview

> Displays a list of event details in chronological order. Designed to show the presentation history for talks, workshops or other events we’ve conducted.

Improvements from the existing site's pattern ([visible lower on this page](https://cloudfour.com/presents/why-you-should-build-a-progressive-web-app-now/)):

1. Each item has a heading element, which makes them easier to navigate with screen readers.
1. Each date is a `<time>` element, which adds semantic meaning.
1. Uses [Subgrid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Subgrid) in supported ~browsers~ [browser](https://caniuse.com/css-subgrid) ([Microsoft is reported to be adding Chromium support](https://blog.chromium.org/2020/06/improving-chromiums-browser.html), though, so 🤞)
1. No need to manually add `padding` and `border` classes based on loop indexes in templates.
1. Simpler markup, requiring four fewer elements and just over half as many characters as the live version.
1. Works across themes.

This PR also renames the `sizes.$list-inline-gap` and `sizes.$media-gap` tokens to `sizes.$gap-inline-medium` and `sizes.$gap-inline-small`. The previous names were too specific to reflect the scope of their usage.

(Apologies if this is a rather anti-climactic PR after #1064: I started working on this PR while waiting for review on that one! I promise I'll be putting those mixins to use shortly. 😄)

## Screenshots

Screen Shot 2020-12-03 at 3 38 10 PM

Screen Shot 2020-12-03 at 3 38 29 PM

Screen Shot 2020-12-03 at 3 38 42 PM

## Testing

Check out [the Event Log component deploy preview](https://deploy-preview-1065--cloudfour-patterns.netlify.app/?path=/docs/components-event-log--example)

---

/CC @AriannaChau

### DIFF
--- a/src/components/event-log/event-log.scss
+++ b/src/components/event-log/event-log.scss
@@ -23,24 +23,48 @@ $_breakpoint-wide: breakpoint.$m;
   --theme-opacity-event-log-location: 1;
 }
 
+/**
+ * Main container
+ *
+ * 1. Defining content columns here, which will be referenced by subgrid in the
+ *    items themselves. This will keep the date column nicely sized in supported
+ *    browsers instead of relying on an arbitrary list.
+ */
+
 .c-event-log {
   display: grid;
   grid-column-gap: sizes.$gap-inline-medium;
-  grid-template-columns: auto 1fr;
+  grid-template-columns: auto 1fr; /* 1 */
 }
 
+/**
+ * Individual item container
+ *
+ * 1. Spans the full width of the container.
+ * 2. In browsers that don't support subgrid, we manually define columns. The
+ *    first column width is a "magic number" that looks okay most of the time
+ *    for a date in the format of "Mon D, YYYY".
+ */
+
 .c-event-log__item {
-  border: 0 solid var(--theme-color-border-event-log-item);
   display: grid;
-  grid-column: span 2;
-  grid-template-columns: 8rem 1fr;
+  grid-column: span 2; /* 1 */
+  grid-template-columns: 8rem 1fr; /* 2 */
+
+  /**
+   * If subgrid is supported, use that instead! ✨
+   */
 
   @supports (grid-template-columns: subgrid) {
     grid-template-columns: subgrid;
   }
 
+  /**
+   * Add space and dividing lines between adjacent items
+   */
+
   &:not(:first-child) {
-    border-top-width: 1px;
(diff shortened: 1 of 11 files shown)


src/components/event-log/event-log.scss
```diff
@@ -0,0 +1,153 @@
...
+ * Individual item container
+ *
+ * 1. Spans the full width of the container.
+ * 2. In browsers that don't support subgrid, we manually define columns. The
+ *    first column width is a "magic number" that looks okay most of the time
+ *    for a date in the format of "Mon D, YYYY".
+ */
+
+.c-event-log__item {
+  display: grid;
```
Paul-Hebert, line 50: Do we need any fallback for non-grid browsers? Or are we fine with default styles at that point?

src/components/event-log/event-log.stories.mdx
```diff
@@ -0,0 +1,32 @@
...
+import eventsDemo from './demo/events.twig';
+const eventsDemoStory = (args) => eventsDemo({ items: events, ...args });
+
+<Meta title="Components/Event Log" />
+
+# Event Log
+
+Displays a list of event details in chronological order. Designed to show the presentation history for talks, workshops or other events we’ve conducted.
+
+For a less specialized display of tabular data, consider using [Table](/?path=/docs/components-table--basic). For simple string/number pairs, consider [Dot Leader](/?path=/docs/components-dot-leader--list). To display a single or upcoming date, consider using [Calendar Date](/?path=/docs/components-calendar-date--basic) (optionally combined with [Media](/?path=/docs/objects-media--event-summary)).
```
Paul-Hebert, line 12: This is great! :+1:

I'd like to see more of this "when not to use this component" type documentation in component libraries

src/components/event-log/event-log.scss
Paul-Hebert: I was unable to actually test this in a browser without Grid since [storybook fails to load in IE11](https://github.com/storybookjs/storybook/issues/8884), however I tried disabling the grid rule and it looked fine, so I think we should be fine here:

Paul-Hebert (approved):
Looks good to me!

PR #1075, opened 2020-12-08
Object: Deck

tylersticka:
## Overview

Adds a layout object intended for displaying a grid of card components. By default, columns are created automatically, but they can be explicitly defined from a particular breakpoint. Horizontal cards will automatically span all available columns.

I am open to alternative names. See Docs and CSS comments for more info.

## Screenshots

Screen Shot 2020-12-07 at 3 58 32 PM

## Testing

1. Proofread [docs for new object](https://deploy-preview-1075--cloudfour-patterns.netlify.app/?path=/docs/objects-deck--basic)
1. Try out [canvas view](https://deploy-preview-1075--cloudfour-patterns.netlify.app/?path=/story/objects-deck--horizontal-card), play with controls

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -61,8 +61,8 @@ $_focus-overflow: (sizes.$edge-large * -1);
   @include fluid.grid-column-gap(
     breakpoint.$s,
     breakpoint.$xl,
-    ms.step(1, 1rem),
-    ms.step(3, 1rem)
+    ms.step(3, 1rem),
+    ms.step(6, 1rem)
   ); /* 3 */
 }
 
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -69,9 +69,11 @@ $_focus-overflow: (sizes.$edge-large * -1);
 /**
  * Responsive horizontal modifiers
  *
- * 1. We define the layout as three columns instead of `2fr 1fr` so that the
+ * 1. If this is in a multi-column grid layout, this insures it will fill the
+ *    available width.
+ * 2. We define the layout as three columns instead of `2fr 1fr` so that the
  *    gaps will line up with three-column elements below.
- * 2. The `fr` rows on the
+ * 3. The `fr` rows on the
  *    ends keep the content rows vertically centered.
  *
  * @todo Coordinate this with the eventual card grid object.
@@ -83,14 +85,15 @@ $_focus-overflow: (sizes.$edge-large * -1);
     $to: xl,
     $include-default: false
   ) {
+    grid-column: 1 / -1; /* 1 */
     grid-template-areas:
       'cover cover .'
       'cover cover header'
       'cover cover content'
       'cover cover footer'
       'cover cover .';
-    grid-template-columns: repeat(3, 1fr); /* 1 */
-    grid-template-rows: 1fr repeat(3, minmax(0, auto)) 1fr; /* 2 */
+    grid-template-columns: repeat(3, 1fr); /* 2 */
+    grid-template-rows: 1fr repeat(3, minmax(0, auto)) 1fr; /* 3 */
   }
 }
 
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -45,8 +45,6 @@ $_focus-overflow: (sizes.$edge-large * -1);
  *    to this container.
  * 3. We define our column gap here instead of in the `c-card--horizontal`
  *    modifiers so we don't have to define it within multiple media queries.
- *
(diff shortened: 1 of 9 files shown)


src/components/card/card.scss
```diff
@@ -61,17 +59,19 @@ $_focus-overflow: (sizes.$edge-large * -1);
...
- * 1. We define the layout as three columns instead of `2fr 1fr` so that the
+ * 1. If this is in a multi-column grid layout, this insures it will fill the
+ *    available width.
+ * 2. We define the layout as three columns instead of `2fr 1fr` so that the
  *    gaps will line up with three-column elements below.
- * 2. The `fr` rows on the
+ * 3. The `fr` rows on the
  *    ends keep the content rows vertically centered.
  *
  * @todo Coordinate this with the eventual card grid object.
```
Paul-Hebert, line 77: Is this TODO still necessary?

src/objects/deck/deck.scss
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * 1. If horizontal items are shown at the wrong column count, they will appear
+ *    to break the grid. This rule keeps items densely packed so it will always
+ *    appear visually correct.
+ */
+
+.o-deck {
+  display: grid;
+  grid-auto-flow: dense; /* 1 */
```
Paul-Hebert, line 14: Do we have any concerns about this leading to mismatched ordering of articles for screen reader users?

src/objects/deck/deck.scss
```diff
@@ -0,0 +1,52 @@
...
+@use '../../mixins/media-query';
+
+/**
+ * 1. If horizontal items are shown at the wrong column count, they will appear
+ *    to break the grid. This rule keeps items densely packed so it will always
+ *    appear visually correct.
+ */
+
+.o-deck {
+  display: grid;
```
Paul-Hebert, line 13: Do we plan to implement any fallback for non-grid supporting browsers?

Right now the experience is a little odd when grid isn't supported. I'm fine if we're consciously making the choice to focus on modern browsers, but I wanted to double check this was intentional and not an oversight.

Paul-Hebert (commented):
I _really_ like how this is looking @tylersticka ! Nice job! I left one comment nit inline and have a couple small concerns:

1. Are we worried about `auto-flow: dense` leading to a mismatch between visual order and tab order?
2. Do we want to provide a fallback for non-grid-supporting browsers?

(I could be convinced the answer to both of those questions is "no" but figured I'd bring them up before approving)

src/objects/deck/deck.scss
tylersticka: I re-read [CSS Grid Layout and Accessibility](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/CSS_Grid_Layout_and_Accessibility) and I think this qualifies as _visual_ rather than _logical_ re-ordering. Navigation through the document will continue to be chronological, and at most tabbing will be a single row off.

At one point I had this moved to a modifier class, but then I remembered this is not a general Grid component and I couldn't think of a use case where we _wouldn't_ want this. I'm open to further discussion, though!

src/components/card/card.scss
tylersticka: No, good catch 👍

src/objects/deck/deck.scss
tylersticka: Yes, I will add a to-do.

Note that our fallback strategy will be a bit different this time around: I don't think our goal should be to make things comparably in <= IE11, but to make the content accessible and to address low-hanging fruit.

src/objects/deck/deck.scss
Paul-Hebert: Cool, yeah it felt slightly jarring for me to jump down a row and then back up, but I don't think it's a major concern. I'm fine moving forward with this, but wanted to ensure we were thinking it through :+1:

src/objects/deck/deck.scss
Paul-Hebert: Totally! The main problems I had when disabling grid were:

1. Images were huge
2. no gaps between cards made it hard to tell what was connected to what

I'd guess we could pretty quickly make a fallback to address those without recreating the whole layout

src/components/card/card.scss
tylersticka: Removed 🗑️

src/objects/deck/deck.scss
tylersticka: To-do added ➕

Paul-Hebert (approved):
🚢

PR #1077, opened 2020-12-09
Component: Avatar

tylersticka:
## Overview

This component has taken a few different permutations. It started as a set of size modifiers, until @AriannaChau and I determined there was really only one size. Then it became a `Media` modifier, but it felt weird to introduce a modifier for styling a specific bit of content. Then finally after reviewing a bunch of other design systems and seeing the precedent for fixed-width avatar components, it seemed reasonable for it to be its own thing.

I also refactored the mixin group formally known as `embed`. I renamed it `ratio-box` so it was more about what it did than the content it accommodated, and restructured it so I could more easily re-use its styles for something like this.

Side note: Patterns like this and Card would work great with some progressively-enhanced lazy loading animations!

## Screenshots

Screenshot_2020-12-09 Storybook

## Testing

[Try the component out on the deploy preview](https://deploy-preview-1077--cloudfour-patterns.netlify.app/?path=/docs/components-avatar--empty)

---

Thank you @AriannaChau for doing discovery related to this component


src/components/avatar/avatar.scss
```diff
@@ -0,0 +1,44 @@
...
+ *    that here.
+ */
+
+.c-avatar {
+  background: var(--theme-color-background-avatar); /* 1 */
+  border-radius: sizes.$radius-full;
+  flex: none; /* 2 */
+  width: sizes.$avatar-medium; /* 3 */
+  // Ratio box hack lets us support fixed _and_ percentage widths with ease
+  @include ratio-box.core-styles();
```
spaceninja, line 35: This is absolutely a nit, but I usually encourage putting `@extend` and `@include` statements at the top of a CSS block. This makes it clear that they're inherited/injected, and that the following styles can be used to override.

Feel free to ignore this feedback if it doesn't align with the C4 house style.

spaceninja:
Overall, this looks great! Happy to approve after address/ignoring the nit I brought up.

src/components/avatar/avatar.scss
tylersticka: @spaceninja I don't have an opinion on this and can't remember why I'm putting it at the end!

Do you know if this is something that can be enforced in our linter?

(Side note: I've also noticed our linting/formatting isn't enforcing quote styles in Sass commands like `@use`.)

spaceninja: Yeah, we should be able to use stylelint to do this. In my stylelint config, I use stylelint-order or enforce alphabetical order. I think that extends to putting most at-rules at the top, but I dont' remember off the top of my head.

Enforcing quote styles is also possible.

The C4 stylelint config is 99% just importing SUIT's config. And since we weren't using SCSS when I set that up, it doesn't include stylelint-scss like my personal config does. There's definitely room to improve our config, I just haven't prioritized it because it seemed like everyone else was happy and I wasn't writing much CSS. But I'm happy to put a bit more work into it.

You know I love my lint configs. 😀

tylersticka: Thank you, @spaceninja, that's helpful info! I'll probably just manually move this for this PR but I think these tools are a good idea.

spaceninja (approved):

Issue #487, opened 2020-02-21
Add comments pattern


dromo77:
Create a pattern for comments, comments can be found on article pages.

<img width="802" alt="comments" src="https://user-images.githubusercontent.com/42841342/75056061-0d62fd80-548b-11ea-83db-c7a8462bb1b4.png">

tylersticka:
These patterns are feeling kind of cramped in longer threads.

I also think we may have flipped the hierarchy the opposite of what it should be, where author/date might be more important up top. I think we were designing these as if they were tweet-length, but since then we've had some pretty _long_ comments where it's unclear who is talking till you've finished reading.

tylersticka:
The last week or two has been kind of interrupty for me, before the holiday I thought I'd share some of my progress.

Here's a chunk of the comment thread for [one of our articles](https://cloudfour.com/thinks/an-html-attribute-potentially-worth-4-4m-to-chipotle/) on our live site on a narrow screen:

<img src="https://user-images.githubusercontent.com/69633/102943661-85180500-446d-11eb-8cb1-82f43e3bdd92.png" alt="Screen Shot 2020-12-22 at 15 47 59" width="375">

I like how clean it is and how the threading is straightforward. I dislike that the name and info is below the comment… it nicely consolidates meta information to one line, but when comments get longer, it omits the context of who is actually speaking. I also think it ends up feeling more and more cluttered the longer the thread gets, as everything is quite condensed.

I've been experimenting with a layout more like this (same content but with random placeholder avatars):

<img width="372" alt="Screen Shot 2020-12-22 at 3 46 32 PM" src="https://user-images.githubusercontent.com/69633/102943747-cad4cd80-446d-11eb-8ec8-6f73528ee963.png">

Things I like:

- I think the avatars are fun and add visual interest. I could see us adding a little Cloud Four badge to team member avatars, for example.
- Having the commenter name first improves a lot of the issues I mentioned earlier.
- I like 💬 for comment versus ↪️, I think it feels more human.
- I dig having more whitespace.

Thing I'm not quite satisfied with:
- The date and reply links feel pretty floaty now. The darker "Reply" links on the right versus the muted dates cause an odd reading line. I originally had dates up near the author but moved them to accommodate longer names, but maybe they'd work better as a unified double-line element with space between that and the comment body?
- The threads feel visually busy. The avatars, body content indentation and thread lines are suggesting  a lot of unresolved whitespace chunks, at least to my eye.

[...]

PR #1130, opened 2021-02-03
Object: Bunch

tylersticka:
## Overview

> A tight cluster of visual elements with a small amount of overlap. Useful for displaying avatars alongside bylines where multiple authors are equally credited.

## Screenshots

Screen Shot 2021-02-03 at 11 42 24 AM

## Testing

1. Confirm GitHub checks pass
2. Try out object in [deploy preview](https://deploy-preview-1130--cloudfour-patterns.netlify.app/?path=/docs/objects-bunch--of-avatars)

---

- See #487 #1093


src/objects/bunch/bunch.scss
```diff
@@ -0,0 +1,45 @@
...
+ * 3. It feels odd to have the first element underneath the last. In lieu of a
+ *    a built-in CSS method of reversing the natural `z-index` order, we use
+ *    a custom property that we'll set with `nth-child` a little further down.
+ *
+ * @see https://github.com/w3c/csswg-drafts/issues/1026#issuecomment-385285148
+ */
+
+.o-bunch > * {
+  flex: none; /* 1 */
+  margin-right: (sizes.$overlap-small * -1); /* 2 */
```
Paul-Hebert, line 31: Smart! I likely would have ended up doing something like this instead (I like your version better)

```
.o-bunch > * + * {
   margin-left: (sizes.$overlap-small * -1); /* 2 */ 
}
```

src/objects/bunch/bunch.scss
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * Set the `--bunch-index` property for the first ten elements. You can manually
+ * set `--bunch-index` for items past that point, but you should really question
+ * if this is the right CSS object to use if you need more than that!
+ */
+
+@for $i from 1 through 10 {
+  .o-bunch > *:nth-child(#{$i}) {
```
Paul-Hebert, line 42: Nit: is the `*` doing anything here?

Paul-Hebert (dismissed):
This looks great Tyler!

I left one super minor code nit inline.

I've also got a design question. Are we concerned about this component breaking out of its container? Should we add `flex-wrap: wrap` to `.o-bunch` so that in the off chance there are too many items to fit in a container it breaks to the next line instead of breaking the layout?

src/objects/bunch/bunch.scss
tylersticka: I don't think it is! Good catch

tylersticka:
> I've also got a design question. Are we concerned about this component breaking out of its container? Should we add `flex-wrap: wrap` to `.o-bunch` so that in the off chance there are too many items to fit in a container it breaks to the next line instead of breaking the layout?

Good idea 👍

Paul-Hebert (approved):
Looks good!

PR #1144, opened 2021-03-11
Expose Design Tokens in npm Package

### LINKED ISSUE #1095: Expose design tokens in npm package
Our new WordPress theme will support Gutenberg. Gutenberg's editor has tools that provide color and font size choices, [which can be specified by a theme](https://developer.wordpress.org/block-editor/developers/themes/theme-support/). Ideally, we'd specify these things by consuming our design tokens for colors and sizes. But we don't currently expose those compiled files in our npm package.

spaceninja:
## Overview

This PR adds our compiled design tokens to the npm package output for
consumption in the WordPress theme.

## Testing

To the best of my knowledge, there's no way to test this without publishing the package, which I can't do until this PR is merged. That said, the change is really simple, and I'm confident in it.

---

- Fixes #1095

### DIFF
--- a/.changeset/serious-candles-cough.md
+++ b/.changeset/serious-candles-cough.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Expose design tokens in npm package
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   "types": "./dist/cloudfour-patterns.d.ts",
   "files": [
     "/dist",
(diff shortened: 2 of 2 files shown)


Paul-Hebert (approved):
Code looks good!

> To the best of my knowledge, there's no way to test this without publishing the package, which I can't do until this PR is merged. That said, the change is really simple, and I'm confident in it.

Yeah I can't think of a way to test this. You could use `npm link` but I don't think that respects the `files` field. I'm fine merging this untested.

PR #1193, opened 2021-04-19
Switch tests to Test Mule

calebeby:
I switched the tests from Cypress to Test Mule. GitHub doesn't show the diff so it is probably helpful to open the old and new test files side by side.


tylersticka (commented):
Deferring review to dev team when they have availability.

spaceninja (approved):
Code LGTM, and I don't see any obvious regressions on the preview environment.

Given the scope of this change, I'd appreciate you getting an approval from @gerardo-rodriguez or @Paul-Hebert as well before merging.

.style-dictionary/build.js
```diff
@@ -92,7 +92,22 @@ StyleDictionary.registerFormat({
...
     });
-    return `module.exports = ${JSON.stringify(tokens, null, '  ')}`;
+    return `export default ${JSON.stringify(tokens, null, '  ')}`;
+  },
+});
+
+/**
+ * Custom Format: JS ESM
+ * This custom format is based on `javascript/module` but it uses `export default` instead of `module.exports`
+ */
```
Paul-Hebert, line 102: Just FYI there is also the [`es6` format](https://amzn.github.io/style-dictionary/#/formats?id=javascriptes6)

This format returns each individual property as an export instead of exporting an object, so it may not work  here.

Paul-Hebert (approved):
Looks good! Thanks Caleb!

.style-dictionary/build.js
calebeby: Yeah I tried that and it wasn't working because lots of the patterns rely on nested objects/destructuring and the es6 format flattened them

PR #1218, opened 2021-05-14
Simplify Storybook sorting

tylersticka:
## Overview

Storybook's sorting options are now very simple. This updates our configuration to take advantage of this, to remove "Design Tokens" (since that's no longer top-level), and to preemptively add a few potential future categories just to make our lives easier.


spaceninja (approved):
🎉

PR #1214, opened 2021-05-11
Upgrade to Storybook 6.2.9

tylersticka:
## Overview

Ran `npx sb upgrade`. This does not resolve the Babel warning (and introduces a new PostCSS warning) but _does_ resolve the other errors mentioned in #1213 and appears to work more reliably. Would benefit from further testing, though.

Note that new Sass files will not be detected without restarting the task.

## Testing

Thoroughly test environment to scare up any uncaught regressions:
- Try editing Twig files
- Try editing story files
- Try editing Sass files
- Try editing JavaScript files
- Try editing test definitions
- Try editing tokens
- Try editing SVG files
- Try various Storybook Canvas tabs

---

- Fixes #1213

Issue #1216, opened 2021-05-13
Comment form pattern(s) - Development


AriannaChau:
We have new designs for the comment forms, there are a few different uses.

### Form at the bottom of the page:

**Small screens:**
![Mobile Copy](https://user-images.githubusercontent.com/24928337/118170334-d0ac5a80-b3de-11eb-889c-d8ab48d6b23a.jpg)

**Large screens:**
![Desktop Copy](https://user-images.githubusercontent.com/24928337/118170337-d2761e00-b3de-11eb-99d3-7cb5a45b05af.jpg)

### When someone is logged in:

**Small screens:**
![Mobile Copy 2](https://user-images.githubusercontent.com/24928337/118171602-42d16f00-b3e0-11eb-9c23-60a255b9811c.jpg)

**Large screens:**
![Desktop Copy 3](https://user-images.githubusercontent.com/24928337/118171620-46fd8c80-b3e0-11eb-88f2-65f7d0625e7e.jpg)

### Response to a comment:

**Small screens:**
![Mobile](https://user-images.githubusercontent.com/24928337/118171756-6eecf000-b3e0-11eb-9bd8-263acb88b685.jpg)

**Large screens:**
![Desktop Copy 2](https://user-images.githubusercontent.com/24928337/118171775-73b1a400-b3e0-11eb-880a-c2a34613242b.jpg)

Here is the design card for this task for more context: https://github.com/cloudfour/cloudfour.com-patterns/issues/1087

Issue #1088, opened 2020-12-15
Email signup form pattern(s)


tylersticka:
## Overview

We syndicate our posts via an email newsletter for those that prefer that to feed readers or social media. There are two variations of that form.

The most prominent is displayed in [the Articles section](https://cloudfour.com/thinks/) header:

<img width="812" alt="Screen Shot 2020-12-15 at 9 58 53 AM" src="https://user-images.githubusercontent.com/69633/102253651-773a1100-3ebc-11eb-9724-9a1dbd39e4fe.png">

But the form is also displayed below individual articles. If JavaScript is disabled or notifications aren't supported, then the form is shown by default, otherwise it requires clicking a button to display:

<img width="787" alt="Screen Shot 2020-12-15 at 9 58 19 AM" src="https://user-images.githubusercontent.com/69633/102253738-9173ef00-3ebc-11eb-82d4-aebd85ba91ea.png">

## Considerations

- Since we already have `c-input`, `c-checkbox` and `c-button` components, the actual layout of the form will most likely be an object.
- It's possible that we could make a single layout object that accommodates this and the comment forms (see #1087), but that isn't a requirement.
- We are moving away from the "floating label" pattern because we no longer recommend it to clients given its unresolved usability and accessibility issues.

PR #1219, opened 2021-05-14
Replace theme add-on with toolbar and global args

tylersticka:
## Overview

After upgrading Storybook (#1214), our themes add-on became available in Docs view but was not functional. I attempted to use he add-on's `withThemes` decorator, but it does not appear to be compatible with inlined HTML stories. So this PR replaces it with a custom toolbar using the Toolbar add-on that ships with Storybook's essential add-ons.

This also helps us overcome an issue where we couldn't reliably configure a theme for a specific story. Now, we can specify a `theme` parameter corresponding to the desired theme class and the story will gain that theme. That means we can remove the `fill-canvas` template we were using as kind of a hack.

## Screenshots

Screen Shot 2021-05-14 at 2 42 32 PM

Screen Shot 2021-05-14 at 2 42 54 PM

Screen Shot 2021-05-14 at 2 43 07 PM

## Testing

On the deploy preview…

- Try the theme toolbar in a Canvas view
- Try the theme toolbar in a Docs view


Paul-Hebert:
This is looking really good Tyler! I noticed a few oddities:

### Broken Docs Pages

Some of our stories have broken docs pages with this change. For example, when I view one of our Prism docs pages I get a console error. Switching to the Canvas view shows that error in the canvas

Screen Shot 2021-05-20 at 4 03 38 PM

### Getting Stuck in a Theme

I'm not exactly sure why this happens but it seems like on the Input docs page specifically I can get stuck in a theme:

theme-stuck

### Bunch

For some reason the Bunch example disappears on the canvas in dark mode. Upping the z-index on `.o-bunch` fixes it. I'm not sure if this is a regression or an existing issue.

Screen Shot 2021-05-20 at 4 00 14 PM

### Table Theme Issues

This is unrelated to this PR, but the table has some issues in dark mode (low contrast caption color, incorrect striped color.)

What's the best place for me to report that? A new issue?

---

Other than that it's looking great! Let me know if you need any more info on any of this!

tylersticka:
These are all great catches, @Paul-Hebert , thank you for testing so thoroughly!

src/components/button/button.stories.mdx
```diff
@@ -35,10 +34,7 @@ Some buttons may not require as much emphasis as others. The `c-button--secondar
 All the above visual styles will appear slightly different within [our dark theme](/?path=/docs/themes--dark). These versions retain their visual emphasis relative to one another while remaining visually distinct from form inputs.
 
 <Canvas>
-  <Story
-    name="Styles (Dark)"
-    parameters={{ themes: { disable: true }, paddings: { disabled: true } }}
-  >
-    {(args) => stylesDarkDemo(args)}
+  <Story name="Styles (Dark)" parameters={{ theme: 't-dark' }}>
```
spaceninja, line 37: Should this be `dark` or `t-dark`? You used `dark` in `themes.stories.mdx`.

src/design/themes.stories.mdx
```diff
@@ -23,15 +13,17 @@ Themes may be applied to a containing element to change that content's appearanc
...
+  </Story>
 </Canvas>
 
 ## Dark
 
 Optimized for page intros and headings.
 
 <Canvas>
-  <Story name="Dark" args={{ theme: 'dark' }}>
+  <Story name="Dark" parameters={{ theme: 't-dark' }} args={{ theme: 'dark' }}>
```
spaceninja, line 26: Should this be `dark` or `t-dark`? You used `t-dark` in `button.stories.mdx`.

src/design/themes.stories.mdx
```diff
@@ -23,15 +13,17 @@ Themes may be applied to a containing element to change that content's appearanc
 The default theme, applied to the document `:root` by default.
 
 <Canvas>
-  <Story name="Light">{demoStory.bind({})}</Story>
+  <Story name="Light" args={{ theme: 'light' }}>
```
spaceninja, line 16: Should this be `light` or `t-light`?

src/design/themes.stories.mdx
tylersticka: It's kind of confusing, but I'm using `args` to tell the demo which theme is applied, and `parameters` to tell the decorator which theme to use on the container.

Storybook's weird.

src/design/themes.stories.mdx
tylersticka: In this case this is me telling the demo which theme is applied so it can modify its text and stuff accordingly, so even though it looks really close to the parameters it's not actually the same thing. I'm hoping this will only happen in the _theme_ demos.

src/components/button/button.stories.mdx
tylersticka: It should be `t-dark`, the `dark` was me passing a noun for the demonstrated theme to the theme demo specifically.

tylersticka:
@Paul-Hebert I just pushed some updates that _may_ address the issues you've observed...

> Broken Docs Pages

I couldn't reproduce this error, but I added extra checks to make sure elements exist before using `querySelector` or `classList` in hopes that it will prevent this from occurring.

> Getting Stuck in a Theme

I also couldn't reproduce this error, but I noticed a different console error relating to my use of `value: undefined` which could, theoretically, get the toolbar stuck. I changed to `null`, which seemed to eliminate the warning... but it may have nothing to do with what you were seeing.

> Bunch

I _could_ reproduce this one! I think I fixed it by applying the theme class to `document.body` instead of `document.documentElement`.

> Table Theme Issues

I've created an issue to resolve this:  #1229

spaceninja:
I'm seeing the same problem @Paul-Hebert saw on the Input page — changing the theme to dark on the Input docs page doesn't change anything.

tylersticka:
> I'm seeing the same problem @Paul-Hebert saw on the Input page — changing the theme to dark on the Input docs page doesn't change anything.

Okay, _now_ I can reproduce. Not sure what I'm doing differently now. I think this may have something to do with inline versus iframed stories.

Thanks, @spaceninja! Sorry for the runaround.

spaceninja:
Screen Shot 2021-05-20 at 5 04 29 PM

I'm not sure if this is a bug, because I'm not sure it's a real-world scenario, but if I apply the global dark theme on the Themes docs page, the nested dark card inside the light theme (that is now dark theme) remains dark. I would expect that if one changes, the other does also, but feel free to disregard if this isn't a situation a real user could end up in.

tylersticka:
> I'm not sure if this is a bug, because I'm not sure it's a real-world scenario, but if I apply the global dark theme on the Themes docs page, the nested dark card inside the light theme (that is now dark theme) remains dark. I would expect that if one changes, the other does also, but feel free to disregard if this isn't a situation a real user could end up in.

Good catch. I think at the very least we could force them to stay the same theme regardless, since this is kind of a special case and someone could (as you suggest) swap themes and then head over here.

spaceninja:
> I think at the very least we could force them to stay the same theme regardless

Could we disable the theme control on that page?

tylersticka:
Gonna have to pick this up tomorrow, but there may be a major blocker to getting this working with iframed docs: https://github.com/storybookjs/storybook/issues/13444

It doesn't even seem like the iframed docs are responding to a change in that toolbar, but I'd like to take a new look with fresh eyes tomorrow.

tylersticka:
> Could we disable the theme control on that page?

Probably!

tylersticka:
I think this bug may be more closely related to the issue we're seeing on Input: https://github.com/storybookjs/storybook/issues/14477

gerardo-rodriguez:
@tylersticka I can't tell the state of this PR at a glance. Are you in the process of addressing feedback or is the PR in a ready state for another review. Let me know how I can help! 🙂

tylersticka:
> I can't tell the state of this PR at a glance. Are you in the process of addressing feedback or is the PR in a ready state for another review. Let me know how I can help! 🙂

@gerardo-rodriguez I'm in the process of addressing feedback at the moment! I just removed the open request for Dev review, sorry about that!

tylersticka:
Okay, I think this is ready for re-review, with a few caveats:

- I can't easily disable the toolbar for specific story files. I've overtly set the theme on both theme stories to address [this comment](https://github.com/cloudfour/cloudfour.com-patterns/pull/1219#issuecomment-845560550).
- I couldn't find a solution so that themes would work for non-inlined stories in Docs mode. I've updated the custom decorator comments to include more information about why this is a struggle.

Both issues are solvable if we wanted to expand this from a custom toolbar and decorator to a full add-on, which would give us the ability to assign more complicated actions to the toolbar items and would allow us to force-refresh stories in contexts like the docs mode. I haven't done this because...

- The remaining issues are known and documented in Storybook with some dev effort already allocated to them. It's likely that these issues will be trivial to address in a future Storybook update.
- Every single existing add-on for this that's compatible with Storybook 6.x has _more_ bugs than this implementation (including what we were using before, which didn't work in Docs mode at all for HTML or Web Component "frameworks"), which suggests to me that this is a non-trivial problem to fully solve.
- The majority of our pattern stories are inlined, as that is the default. Of the seven patterns that are not inlined, three are layout objects, two are theme containers themselves (Sky Nav and Cloud Cover), one is only optimized for a light theme (Ground Nav), and the last pattern (Input) can be safely inlined once we have more form layouts to better visualize the alignment of inputs, labels and buttons (see #1088 and #1216).

TL;DR: This isn't perfect, but it's a lot better than our current solution, and is likely to improve more as our dependencies update and our project evolves. But let me know if you disagree!

.storybook/theme-decorator.js
```diff
@@ -0,0 +1,65 @@
...
+  } else if (context.viewMode === 'docs') {
+    useEffect(() => {
+      // Remove any existing theme classes from the root element
+      updateTheme(document.body);
+      // Query for the most appropriate story element
+      const storyElement = document.querySelector(`#story--${context.id}`);
+      // Only proceed if we actually found the relevant story.
+      if (storyElement) {
+        const previewElement = storyElement.closest('.docs-story');
+        updateTheme(previewElement, theme);
```
gerardo-rodriguez, line 59: I don't yet have an understanding of story elements and their contents. Is it a guaranty that there will always exist a `.docs-story` or do we need to check that `previewElement` is not `null` before attempting to update the theme? 🤔 ¯\\\_(ツ)_/¯

Paul-Hebert: This seems like a good safety check given that the underlying Storybook markup is outside our control

Paul-Hebert (approved):
> TL;DR: This isn't perfect, but it's a lot better than our current solution, and is likely to improve more as our dependencies update and our project evolves. But let me know if you disagree!

That makes sense to me :+1: I'm cool with merging this in, though agree with Gerardo that it would be nice to add a safety check to confirm the `.story-docs` element exists before interacting with it.

.storybook/theme-decorator.js
tylersticka: The `updateTheme` function does check that the element is truthy and has a `classList` property, so I don't think another check before calling it is necessary.

PR #1227, opened 2021-05-18
Favicons

### LINKED ISSUE #746: Update favicons

tylersticka:
## Overview

Adds and documents the following favicons:

- `favicon.ico` for legacy browsers
- `icon.svg` for modern browsers
- `icon-192.png` and `icon-512.png` for Android
- `apple-touch-icon.png` for iOS/macOS

It builds on the work that @aileenjeffries started (see #746) but with the following changes:

- It updates our Storybook build commands to allow `src/assets` as a static directory. This eliminates the need for duplicating icons in `./static` (and we should maybe consider deprecating `./static` eventually?)
- It replaces `manager-head.html` with the `managerHead` option in `main.js`, which gives us a bit more freedom formatting the `link` element so we can maintain the simplicity of `favicon.ico` as a filename. Without this change, implementers would have to rename `favicon-production.ico` to `favicon.ico` file to follow our recommendations.
- Our last pattern library provided a web app manifest file, but we threw it out as soon as we needed to do anything specific to that experience (such as add a service worker), so I went the route of documenting what those files should include instead of providing the whole file.
- Windows no longer requires separate icons for pinned tiles, so that image and configuration are no longer necessary.
- Safari no longer requires specific pinned icon images, and Apple has stopped using them on their own web properties.
- [We never needed `shortcut` in our link rel](https://mathiasbynens.be/notes/rel-shortcut-icon). 🤯
- I created a slightly simpler version of her tab demo that works better on small screens and isn't broken by our them selector.
- I also created some simple previews of the app icons.

## Screenshots

Screenshot_2021-05-17 Storybook

## Testing

[Deploy preview](https://deploy-preview-1227--cloudfour-patterns.netlify.app/?path=/docs/design-favicons--favicons#favicons)

---

- Fixes #746


Paul-Hebert (approved):
This looks great! The icons look good, the code looks good, and I like how it's documented :+1:

PR #1239, opened 2021-05-26
Add animation to sky nav

### LINKED ISSUE #1137: Animate Sky Nav opening/closing
Breaking this chunk of work out from #718

 * [ ] When toggling the nav, we should include an animation.
 * [ ] No animation should occur if the user prefers reduced motion.

calebeby:
## Overview

Closes #1137 

## Testing

1. Open [Sky Nav](https://deploy-preview-1239--cloudfour-patterns.netlify.app/?path=/story/components-sky-nav--dark)
2. Resize to small screen
3. Toggle the menu, make sure that it animates correctly, and the content is hidden/shown correctly

### DIFF
--- a/src/components/sky-nav/sky-nav.stories.mdx
+++ b/src/components/sky-nav/sky-nav.stories.mdx
@@ -38,8 +38,10 @@ The Sky Nav is intended for use with our [our dark theme](/?path=/docs/themes--d
   >
     {() => {
       useEffect(() => {
-        const {destroy} =initSkyNav(document.querySelector('.js-sky-nav-menu-toggle'));
-        return destroy
+        const { destroy } = initSkyNav(
+          document.querySelector('.js-sky-nav-menu-toggle')
+        );
+        return destroy;
       }, []);
       return template({ class: 't-dark', include_main_demo: true, menu });
     }}
@@ -60,8 +62,10 @@ But it works without a theme, too, just in case:
   >
     {() => {
       useEffect(() => {
-        const {destroy} =initSkyNav(document.querySelector('.js-sky-nav-menu-toggle'));
-        return destroy
+        const { destroy } = initSkyNav(
+          document.querySelector('.js-sky-nav-menu-toggle')
+        );
+        return destroy;
       }, []);
       return template({ include_main_demo: true, menu });
     }}
--- a/src/components/sky-nav/sky-nav.stories.mdx
+++ b/src/components/sky-nav/sky-nav.stories.mdx
@@ -74,7 +74,7 @@ But it works without a theme, too, just in case:
 
 ## Elements
 
-- `c-sky-nav`: A containing `<div>` element.
+- `c-sky-nav`: A containing `<div>` element. This element should also have the `js-sky-nav` class, which is used for the open/close animation. The animation uses it to figure out which elements below the containing element need to be pushed up or down.
 - `c-sky-nav__skip`: A link to the main content area, visibly hidden until focus. Works great with [our `c-button` pattern](/?path=/docs/components-button--elements).
 - `c-sky-nav__content`: A `<header>` element representing the actual menu contents.
 - `c-sky-nav__masthead`: Containing element for the logo and menu toggle.
--- a/src/components/sky-nav/sky-nav.stories.mdx
+++ b/src/components/sky-nav/sky-nav.stories.mdx
@@ -38,7 +38,8 @@ The Sky Nav is intended for use with our [our dark theme](/?path=/docs/themes--d
   >
     {() => {
       useEffect(() => {
-        initSkyNav(document.querySelector('.js-sky-nav-menu-toggle'));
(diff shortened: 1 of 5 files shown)


src/components/sky-nav/sky-nav.ts
```diff
@@ -13,55 +13,107 @@ import tokens from '../../compiled/tokens/js/tokens';
...
+    // We need to keep track of the siblings after the menu,
+    // because we will push them down for the animation
+    const elementsToShift: HTMLElement[] = [navWrapper];
+    let sibling: HTMLElement | null = navWrapper;
+    // eslint-disable-next-line no-unmodified-loop-condition
+    while ((sibling = sibling.nextElementSibling as HTMLElement | null)) {
+      elementsToShift.push(sibling);
+    }
+
+    const duration = 0.5;
```
calebeby, line 64: I copied this from https://github.com/cloudfour/cloudfour.com-patterns/blob/main/src/assets/toolkit/scripts/lib/component/sky.js#L7

src/components/sky-nav/sky-nav.stories.mdx
```diff
@@ -38,7 +38,10 @@ The Sky Nav is intended for use with our [our dark theme](/?path=/docs/themes--d
   >
     {() => {
       useEffect(() => {
-        initSkyNav(document.querySelector('.js-sky-nav-menu-toggle'));
+        const { destroy } = initSkyNav(
+          document.querySelector('.js-sky-nav-menu-toggle')
+        );
+        return destroy;
```
gerardo-rodriguez, line 44: 🎉

src/components/sky-nav/sky-nav.ts
```diff
@@ -13,55 +13,107 @@ import tokens from '../../compiled/tokens/js/tokens';
  */
 export const initSkyNav = (navButton: HTMLButtonElement) => {
   const menu = navButton.nextElementSibling as HTMLElement;
+  const navWrapper = navButton.closest('.c-sky-nav') as HTMLElement; // Review: should this use a .js- class or some different selector?
```
gerardo-rodriguez, line 16: > Review: should this use a .js- class or some different selector?

Good call out. Unless we've changed our conventions, this should use a `.js-` CSS class.

src/components/sky-nav/sky-nav.ts
```diff
@@ -13,55 +13,107 @@ import tokens from '../../compiled/tokens/js/tokens';
...
   const largeScreenMediaQuery = window.matchMedia(
     `(min-width: ${tokens.size.breakpoint.m.value})`
   );
+  const reducedMotionMediaQuery = window.matchMedia(
+    '(prefers-reduced-motion: reduce)'
+  );
```
gerardo-rodriguez, line 22: 🎉

src/components/sky-nav/sky-nav.ts
gerardo-rodriguez: Or...hmm...now I'm not sure. 🤔 

What do you think, @Paul-Hebert or @spaceninja?

gerardo-rodriguez: Am I correct in understanding before we didn't need to worry about the wrapper element which is why we never had to make this decision in the past? 🔍 👀

src/components/sky-nav/sky-nav.ts
```diff
@@ -13,55 +13,107 @@ import tokens from '../../compiled/tokens/js/tokens';
...
+    const elementsToShift: HTMLElement[] = [navWrapper];
+    let sibling: HTMLElement | null = navWrapper;
+    // eslint-disable-next-line no-unmodified-loop-condition
+    while ((sibling = sibling.nextElementSibling as HTMLElement | null)) {
+      elementsToShift.push(sibling);
+    }
+
+    const duration = 0.5;
+    const transition = `transform ${duration}s ease-in-out`;
+    clearTimeout(timeoutId);
```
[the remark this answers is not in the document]
calebeby: Yeah it is nice that it ignores it

src/components/sky-nav/sky-nav.ts
Paul-Hebert: Hmm, good question. If we need to target it with JS I'd lean towards adding a `js-` class.

gerardo-rodriguez: That makes sense, thanks, @Paul-Hebert. I am now leaning the same way. 

We should also update the [documentation](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/components-sky-nav--dark#javascript-instructions) as part of this change.

It's still a bit odd because we initialize on the button when calling `initSkyNav()` but the wrapper is also important and targeted via JS. If we document it well, it should be fine, though.

gerardo-rodriguez (approved):
Great work, @calebeby! 🎉

Paul-Hebert (approved):
Looks good, nice job Caleb!

PR #1244, opened 2021-05-27
Author pattern

### LINKED ISSUE #1215: Author pattern(s) - development
Here are the new designs for the byline:

byline-small
byline-large

The screenshots are examples of multiple authors. We've also decided for now to keep the extra information (technical reviews, illustrations) out of the byline.
 
Here is the design card for this task for more context: https://github.com/cloudfour/cloudfour.com-patterns/issues/1093

calebeby:
## Overview

Closes #1215

Things I haven't done / want feedback/advice on:

- ~I didn't break out the date on small screens, it always is right under the authors ([Tyler's comment](https://github.com/cloudfour/cloudfour.com-patterns/issues/1215#issuecomment-843520454))~ ✅ 
- ~I didn't change the text color. I couldn't find utility classes for colors (@AriannaChau are there color utility classes, if so which one should I use, if not which sass variable is the right color)~ ✅ 
- Tyler suggested [copying the property names from Timber](https://github.com/cloudfour/cloudfour.com-patterns/issues/1215#issuecomment-843520454) and I did, but I changed them to camelCase (I could change them back to snake_case if that is better)

## Screenshots

image

## Testing

Preview: https://deploy-preview-1244--cloudfour-patterns.netlify.app/

Use the control to change the # of authors displayed


calebeby:
cc @AriannaChau this is still a draft, but let me know what you think about the text color and breaking out the date onto the line below

AriannaChau:
@calebeby sorry I didn't get to this yesterday! I think where the date breaks is fine, especially if it makes implementation easier.

For the text color, we just want it slightly lighter than black, I think `$base-gray-dark` would be a good option.

AriannaChau (commented):
This _looks_ good to me but I would refer to a dev to look over the code.

tylersticka:
>     * Tyler suggested [copying the property names from Timber](https://github.com/cloudfour/cloudfour.com-patterns/issues/1215#issuecomment-843520454) and I did, but I changed them to camelCase (I could change them back to snake_case if that is better)

I think changing their case might defeat the purpose of using the Timber property names to streamline WordPress integration? 🤔

src/components/author/author.scss
```diff
@@ -0,0 +1,9 @@
+@use "../../compiled/tokens/scss/color";
+
+/**
+ * Container
+ */
+
+.c-author {
+  color: color.$base-gray-dark;
```
tylersticka, line 8: Seems to break with dark theme applied

Screen Shot 2021-06-01 at 9 27 12 AM

Paul-Hebert (commented):
Code looks good to me, but I agree with @tylersticka 's feedback. Happy to re-review after that's addressed or defer approval to @tylersticka

calebeby:
Fixed ✅ 

Ready for re-review

@tylersticka I was wondering about a couple more things for the avatars:

- Should we pass through alt text for screen readers? Right now they have alt="", which I _think_ is OK since that means that screen readers don't try to read them, and the visible text next to the image says everything anyways
- Should we make the avatars links too?

src/components/author/author.scss
```diff
@@ -0,0 +1,13 @@
...
+}
+
+.t-dark .c-author {
+  color: color.$base-white;
```
calebeby, line 12: Should I add a custom property color for this theme-dependent switch to the themes scss file?

tylersticka:
> Should we pass through alt text for screen readers? Right now they have alt="", which I _think_ is OK since that means that screen readers don't try to read them, and the visible text next to the image says everything anyways

I think that's fine, unless...

> Should we make the avatars links too?

…we do this, in which case there's functionality there. It's unclear to me in that case if the best thing to do is to add `alt` text at the risk of duplicating content, or to make the avatars `aria-hidden` or something. Or maybe just to keep them link-less.

Would be curious what folks like @gerardo-rodriguez and @Paul-Hebert think.

src/components/author/author.scss
tylersticka: What I would do is this:

1. Add a new `--theme-color-text-muted` or `--theme-color-text-subtle` (or whatever) property to `src/base/_themes.scss` for light and dark themes.
1. Update `.c-author` to reference this property.

But I'm open to other ideas.

src/base/_themes.scss
```diff
@@ -37,6 +38,7 @@
   --theme-color-text-base: #{color.$text-light};
   --theme-color-text-code: var(--theme-color-text-emphasis);
   --theme-color-text-emphasis: #{color.$text-light-emphasis};
+  --theme-color-text-muted: #{color.$base-white};
```
tylersticka, line 41:
```suggestion
  --theme-color-text-muted: inherit;
```

As-is, this will probably unintentionally make the text slightly brighter than `$text-light`. If the intention is for it to retain the parent text color, maybe we can just make `inherit` the default?

src/base/_themes.scss
```diff
@@ -21,6 +21,7 @@
   --theme-color-text-base: #{color.$text-dark};
   --theme-color-text-code: #{color.$base-fuchsia};
   --theme-color-text-emphasis: var(--theme-color-text-base);
+  --theme-color-text-muted: #{color.$base-gray-dark};
```
tylersticka, line 24: Can we add a `$text-dark-muted` token?

tylersticka (commented):
Some comments inline.

Also, after looking into examples online, I think we should leave the avatars un-linked for now.

src/base/_themes.scss
calebeby: There already is a token with that name, but it is much lighter:

image

Original design:

image

tylersticka: @calebeby Let's go ahead and reference the existing token, if we feel it's too light that should embolden us to update the token rather than add an _additional_ subtle/muted color.

gerardo-rodriguez:
> It's unclear to me in that case if the best thing to do is to add `alt` text at the risk of duplicating content, or to make the avatars `aria-hidden` or something. Or maybe just to keep them link-less.
> 
> Would be curious what folks like @gerardo-rodriguez and @Paul-Hebert think.

I argue the images fall under the "decoration" category, therefore, should have an empty `alt` value. No need to duplicate content, especially if they are not linked. 🙂 

One other idea that came to mind as I was testing with VoiceOver is the opportunity to add more context to the names and the date.

Visually I get extra context to let me know the linked names are the authors and the date is the publish date.

Without the visual context, though, it's unclear what the linked names or date means.

Could we add visually hidden text to add more context? Something like, "Authored by <name>" and "Published on <date>".

src/components/author/author.twig
```diff
@@ -0,0 +1,21 @@
...
+    {% endblock %}
+  {% endembed %}
+  <div class="o-media__content">
+    <div>
+      {% for author in authors %}
+        {% if loop.last and loop.length > 1 %}and {% endif %}
+        <a href="{{ author.link() }}">{{ author.name() }}</a>
+        {%- if not loop.last and loop.length > 2 %},{% endif %}
+      {% endfor %}
+    </div>
```
gerardo-rodriguez: I mentioned this in my [last comment](https://github.com/cloudfour/cloudfour.com-patterns/pull/1244#issuecomment-852561426), but figured I'd add something here too. 🙂 

Could we add visually hidden text to provide more context? 

Ideas include:
- Authored by `<name>`, `<name>`
- Authors: `<name>`, `<name>`

Ideally, the rendered output is something like:
```html
<div>
  <span class="u-hidden-visually">Authored by</span>
  <a href="https://cloudfour.com/is/megnotarte">Megan Notarte</a>,
  <a href="https://cloudfour.com/is/gerardo">Gerardo Rodriguez</a>,
  <a href="https://cloudfour.com/is/tyler">Tyler Sticka</a>
</div>
```

calebeby: Yep! Already done! [`c616e9e`](https://github.com/cloudfour/cloudfour.com-patterns/pull/1244/commits/c616e9e57b9b2431d6cd42f9cda251b0e2140b18) 🎉

src/components/author/author.twig
```diff
@@ -0,0 +1,23 @@
...
+      <span class="u-hidden-visually">Authored by</span>
+      {% for author in authors %}
+        {% if loop.last and loop.length > 1 %}and {% endif %}
+        <a href="{{ author.link() }}">{{ author.name() }}</a>
+        {%- if not loop.last and loop.length > 2 %},{% endif %}
+      {% endfor %}
+    </div>
+    <div>
+      <span class="u-hidden-visually">Published on</span>
+      <time datetime="{{ date | date("Y-m-d") }}">{{ date | date("F jS, Y") }}</time>
```
gerardo-rodriguez, line 20: The date gets read twice by VoiceOver so I got curious how others have handled this (both the `datetime` value and the string text is read). I found [this GH issue](https://github.com/bbc/simorgh/issues/1449) but it seems this is expected behavior. 

Leaving this note to document my findings but for now, it's not actionable as I was unable to find anything that helped confirm a different expected behavior.

calebeby: Good catch! I'm not sure the best way to handle that

tylersticka (approved):

src/components/author/author.twig
gerardo-rodriguez: > I'm not sure the best way to handle that

No worries, I did a bit more research this AM but still wasn't able to find anything. It's okay to leave as-is. 👍

gerardo-rodriguez (approved):
👍

PR #1240, opened 2021-05-26
Button Group layout object

gerardo-rodriguez:
## Overview

This PR introduces a Button Group layout object pattern.

Other changes include:
- Update Button component demos to use Button Group layout object.

## Screenshots

Screen Shot 2021-06-01 at 4 08 03 PM

Screen Shot 2021-05-28 at 5 08 47 PM

## Testing

1. Review the [Button Group layout object pattern](https://deploy-preview-1240--cloudfour-patterns.netlify.app/?path=/docs/objects-button-group--default-story) at smaller and larger viewports.

---

- Closes #1091

PR #1256, opened 2021-06-01
Add comment form pattern

### LINKED ISSUE #1216: Comment form pattern(s) - Development

calebeby:
## Overview

Closes #1216  

## Screenshots

image

## Testing

[Preview](https://deploy-preview-1256--cloudfour-patterns.netlify.app/?path=/story/components-comment-form--default-story)

Try on small screens and large screens
Try with control set to logged out, and set to logged in


src/components/comment-form/comment-form.twig
```diff
@@ -0,0 +1,29 @@
+<form
+  class="c-comment-form{% if class %} {{class}}{% endif %}"
+  {%- if action %}action="{{ action }}"{% endif -%}
+  {%- if method %}method="{{ method }}"{% endif -%}
+>
+  <h2>Leave a Comment</h2>
+  <p>Please be kind, courteous and constructive. You may use simple HTML or <a href="https://en.support.wordpress.com/markdown-quick-reference">Markdown</a> in your comments. All fields are required.</p>
```
tylersticka, line 7: Can this be an optional `{% block %}` so it may be overridden?

src/components/comment-form/comment-form.twig
```diff
@@ -0,0 +1,29 @@
...
+    <label class="c-comment-form__email">
+      <div class="c-comment-form__label">Email</div>
+      {% include '@cloudfour/components/input/input.twig' with { type: 'email', autocomplete: 'home email', required: true } only %}
+    </label>
+  {% endif %}
+  <label class="c-comment-form__subscribe">
+    {% include '@cloudfour/components/checkbox/checkbox.twig' with { name: 'subscribe_comments' } only %}
+    Notify me of follow-up comments by email.
+  </label>
+  {% include '@cloudfour/components/button/button.twig' with { label: 'Submit Comment' } only %}
```
tylersticka, line 28: This appears to only address the top-level comment use case and not the reply use case described in #1216. Is that intentional?

src/components/comment-form/comment-form.twig
```diff
@@ -0,0 +1,29 @@
...
+    <div class="c-comment-form__label">Message</div>
+    {% include '@cloudfour/components/input/input.twig' with { type: 'textarea', name: 'comment', required: true } only %}
+  </label>
+  {% if loggedInUser and logOutUrl %}
+    <p>Logged in as <a href="{{ loggedInUser.link() }}">{{ loggedInUser.name() }}</a>. <a href="{{ logOutUrl }}">Log out?</a></p>
+  {% else %}
+    <label class="c-comment-form__name">
+      <div class="c-comment-form__label">Name</div>
+      {% include '@cloudfour/components/input/input.twig' with { name: 'author', autocomplete: 'name', required: true } %}
+    </label>
```
tylersticka, line 18: This may be overkill, but is there a low-level "label + input" layout object we might want to create for this recurring combination?

src/components/comment-form/comment-form.twig
```diff
@@ -0,0 +1,29 @@
...
+  {%- if action %}action="{{ action }}"{% endif -%}
+  {%- if method %}method="{{ method }}"{% endif -%}
+>
+  <h2>Leave a Comment</h2>
+  <p>Please be kind, courteous and constructive. You may use simple HTML or <a href="https://en.support.wordpress.com/markdown-quick-reference">Markdown</a> in your comments. All fields are required.</p>
+  <label>
+    <div class="c-comment-form__label">Message</div>
+    {% include '@cloudfour/components/input/input.twig' with { type: 'textarea', name: 'comment', required: true } only %}
+  </label>
+  {% if loggedInUser and logOutUrl %}
```
tylersticka, line 12: Can we use snake_case to match Twig conventions?

tylersticka (commented):
Looks great! Some suggestions inline

src/components/comment-form/comment-form.twig
calebeby: Yeah, I was thinking about that while implementing this. Do you think it would make sense to modify the input component to have an optional (or required?) `label` prop that wraps the input in a label and stylizes the label?

src/components/comment-form/comment-form.twig
calebeby: I did not see that!

src/components/comment-form/comment-form.twig
tylersticka: I guess it depends on whether or not it's simpler to create a separate layout object for that or to extend the input component template. I would traditionally opt for the former, but I'm more open than usual to the latter after [some recent experiences](https://cloudfour.com/thinks/mighty-morphin-web-components/).

src/components/comment-form/comment-form.scss
```diff
@@ -0,0 +1,37 @@
...
+  }
+}
+
+.c-comment-form__subscribe {
+  display: flex;
+
+  & > input {
+    margin-right: ms.step(-2, 1rem);
+  }
+}
```
tylersticka, line 27: Can this use `o-media` instead?

Example: https://v-next--cloudfour-patterns.netlify.app/?path=/story/objects-media--checkbox-label

src/components/comment-form/comment-form.scss
```diff
@@ -0,0 +1,37 @@
...
+
+.c-comment-form__reply-buttons {
+  display: grid;
+  grid-gap: ms.step(-2, 1rem);
+  grid-template-columns: 1fr 1fr;
+
+  @media (min-width: size.$breakpoint-s) {
+    margin-right: auto;
+  }
+}
```
tylersticka, line 37: Could this leverage the Button Group stuff somehow? #1240

src/components/comment-form/comment-form.twig
```diff
@@ -0,0 +1,27 @@
+<form
+  class="c-comment-form{% if class %} {{class}}{% endif %}"
+  {%- if action %}action="{{ action }}"{% endif -%}
+  {%- if method %}method="{{ method }}"{% endif -%}
+>
+  <h2>Leave a Comment</h2>
+  <p>{% block prompt %}Please be kind, courteous and constructive. You may use simple HTML or <a href="https://en.support.wordpress.com/markdown-quick-reference">Markdown</a> in your comments. All fields are required.{% endblock %}</p>
```
tylersticka, line 7: We need a way to override all of this intro HTML, not just the one sentence.

src/components/input/input.scss
```diff
@@ -152,3 +153,7 @@
     -webkit-text-fill-color: color.$text-dark-muted; /* 1 */
   }
 }
+
+.c-input__label {
+  margin-bottom: ms.step(-5, 1rem);
+}
```
tylersticka, line 159: This isn't a valid element because it's a sibling.

We should use an object class for this relationship. It's okay to then _use_ that object class within the component template, but as-is this is breaking our BEM conventions.

tylersticka (commented):
It looks like there may be a regression in the layout at large sizes? I don't see the name and email columns side-by-side anymore...

Screen Shot 2021-06-02 at 1 21 07 PM

src/components/input/input.scss
calebeby: I'm not sure what the class to use for this should be. And could I add it to this input scss file or would it need to live in its own file in src/objects?

src/components/comment-form/comment-form.scss
calebeby: I think it might be easiest to do that in a follow up PR. I made a sticky note so I won't forget 🗒️

src/components/input/input.scss
calebeby: It feels like it might make sense for the .c-input to be applied to the label when there is a label, since it is the "root element" of the component. How about `.c-input` (root element, either input, textarea, select, or label), `.c-input__input` (input, select, or textarea), `.c-input__label` (the div containing the label text when there is a label)

tylersticka: If we want to encourage usage of `c-input` in varying layouts... for example, if we think we may want to have things like joined inputs or inputs where the label is to the side or something like that... then creating a new pattern like `o-form-group` ([example](https://primer.style/css/components/forms#form-groups)) is probably the right approach. That's my instinct.

Alternatively, we could refactor `c-input` so that it's structured more like this:

```html
<div class="c-input">
  <label class="c-input__label">...</label>
  <input class="c-input__control" ... />
</div>
```

But as I'm typing that out, I'm doubting myself a bit on that. 🤔

calebeby: I think I'm in favor of whichever solution is least-resistance to getting this PR merged. I think that creating a new pattern is a little out of scope for this PR.

In terms of restructuring `c-input`, I think it makes more sense to associate the label and form element by wrapping the label around the input, because dealing with id's is more complicated and it is easy to end up making mistakes with duplication. Here's what I suggested in my last comment, as a code block:

```html
<label class="c-input">
  <div class="c-input__label">...</div>
  <input class="c-input__control" ... />
</label>
```

tylersticka: > I think that creating a new pattern is a little out of scope for this PR.

Maybe, but in that case shouldn't we create the smaller pattern dependency as a separate PR and then update this one?

It seems like what we realize we need is this sort of structure:

```
c-comment-form
├── o-form-group
│   └── c-input
├── o-form-group
│   └── c-input
└── (etc.)
```

We've merged in `c-input`, and you've created a PR for `c-comment-form`. But we need `o-form-group`. I don't understand why we wouldn't go create `o-form-group` instead of refactoring `c-input`?

calebeby: Can you open an issue for o-form-group with examples of the use-cases?

tylersticka: I sure can! 👍

tylersticka: Done: #1258

Paul-Hebert:
Hey @calebeby I wasn't sure what the status of this PR and its ongoing conversations was. Is this ready for dev review or should I hold off?

calebeby:
@Paul-Hebert It is on hold until we create a form-group pattern: https://github.com/cloudfour/cloudfour.com-patterns/pull/1256#discussion_r645057365

Paul-Hebert:
Cool, thanks!

src/components/comment-form/comment-form.scss
calebeby: @tylersticka The button group object doesn't force the buttons to be the same width regardless of text length, but in https://github.com/cloudfour/cloudfour.com-patterns/issues/1216 it shows them being the same width.

I could:

1. Update button group pattern so the buttons don't grow/shrink based on text length
2. Let the buttons in comment form grow/shrink based on text length
3. Not use button group, just use the CSS above ☝️
4. Add a modifier to the button group pattern with the CSS above ☝️ 

What do you think?

calebeby: Also, should the button group pattern have the "full width on small screens" behavior shown here:

tylersticka: > 4. Add a modifier to the button group pattern with the CSS above ☝️

This one sounds good to me

> Also, should the button group pattern have the "full width on small screens" behavior shown here:

Probably, potentially as a modifier?

src/mixins/_media-query.scss
```diff
@@ -72,6 +95,34 @@ $last-breakpoint: list.nth($breakpoint-keys, -1);
...
+  $from: $first-breakpoint,
+  $to: $last-breakpoint,
+  $include-default: true
+) {
+  @if ($include-default) {
+    @content;
+  }
+
+  @include breakpoint-loop-max-width($from, $to) using ($key) {
+    &\@#{$key} {
```
calebeby, line 120: Should it use different classes here? It might be confusing for `@s` to mean "below s breakpoint" in some contexts and "above s breakpoint" in other contexts.

What do you think about using `@below-s` or `@max-s`. Also we could potentially rename the other ones too to `@above-s` or `@min-s` for clarity.

src/mixins/_media-query.scss
```diff
@@ -72,6 +95,34 @@ $last-breakpoint: list.nth($breakpoint-keys, -1);
...
+/// @param {String} $from [$first-breakpoint] - The first/narrowest breakpoint
+/// token key to start looping over.
+/// @param {String} $to [$last-breakpoint] - The last/widest breakpoint token
+/// key to finish looping over.
+/// @param {Bool} $include-default [true] - If true, the `@content` will be
+/// output in the current parent before any media queries.
+/// @content
+/// @output The `@content` nested within variations of the parent class that
+/// include media query token segments.
+@mixin breakpoint-classes-max-width(
```
calebeby, line 110: I don't love this name, open to suggestions

calebeby:
This is ready for review. I left a couple questions

src/mixins/_media-query.scss
tylersticka: I don't think we should rename the `min-width` ones. Since we design for mobile first, it's usually more common for the mobile to be the default and for changes to apply at larger sizes.

I'm re-reading some other CSS frameworks' reasoning behind only having `min-width` utilities ([example](https://tailwindcss.com/docs/responsive-design)) and I'm second-guessing this direction. Is it possible to carve off the `c-button-group` changes as a separate issue for now?

calebeby: > Is it possible to carve off the c-button-group changes as a separate issue for now?

Sure, I can open an issue. What should happen to this PR? Should I revert it back to not use the button-group pattern?

tylersticka: Or you could just have it use the button group pattern in its default state. Either way is fine.

calebeby:
@tylersticka Feel free to take another look!

tylersticka (approved):

Issue #1004, opened 2020-10-27
Improve source code previews in Storybook


tylersticka:
By default, Storybook displays the JavaScript source for stories when you click the "show code" button on story previews in docs mode:

<img width="1034" alt="Screen Shot 2020-10-26 at 3 55 54 PM" src="https://user-images.githubusercontent.com/69633/97237111-c6f53980-17a3-11eb-910b-90b267a852dc.png">

This is not particularly useful for this project, where most of the patterns are Twig templates. There's never a place where that code would be copied and pasted anywhere.

We recently updated Storybook to 6.x, which included a few improvements to the customization of source previews. Unfortunately these don't seem to be capable of full automation, but it would allow us to do something like this...

```jsx
import { Story, Canvas, Meta } from '@storybook/addon-docs/blocks';
import template from './example/template.twig';
import templateSrc from '!!raw-loader!./example/template.twig;
// ...

<Canvas>
  <Story name="Example" parameters={{
    docs: { source: { code: templateSrc } }
  }}>
    {(args) => template(args)}
  </Story>
</Canvas>
```

This would get us a lot closer, but it does call into question our current way of keeping things DRY through demos. Consider these two examples, a default and a prose container:

<img width="1026" alt="Screen Shot 2020-10-26 at 4 54 52 PM" src="https://user-images.githubusercontent.com/69633/97240689-28210b00-17ac-11eb-8423-4fb23a008e51.png">
<img width="1024" alt="Screen Shot 2020-10-26 at 4 55 00 PM" src="https://user-images.githubusercontent.com/69633/97240695-29eace80-17ac-11eb-83da-e746648810e2.png">

Notice that the examples are the same, and aren't really indicative of how things would work in the wild. That's because the demos were written from the point of view of the maintainer keeping things DRY and not as a point of reference for the viewer. Does this mean rewriting all of the demos so far to keep them more relevant?

Alternatively, we could manually include an additional example for each pattern:

```jsx
<Source
  language="twig"
  code={`{% embed '@cloudfour/objects/container/container.twig' %}
  {% block content %}
    Hello world
  {% endblock %}
{% endembed %}`}
></Source>
```

Which would display _outside_ the story:

<img width="1050" alt="Screen Shot 2020-10-26 at 5 01 25 PM" src="https://user-images.githubusercontent.com/69633/97240989-fb212800-17ac-11eb-9c74-b9757adc9f9a.png">

[...]

PR #1257, opened 2021-06-01
Support Twig highlights in Storybook

tylersticka:
## Overview

Storybook's kind of weird. It seems to use tree-shaking to dynamically include highlight language definitions for doc blocks, but not for code blocks specified via Markdown. Thankfully, they're using a well-documented library for code definitions, so I was able to extend that from our `preview.js`.

With this change, any Twig blocks in our docs should receive syntax highlighting.

## Screenshots

Before:
Screen Shot 2021-06-01 at 4 20 23 PM

After:
Screen Shot 2021-06-01 at 4 19 30 PM

## Testing

Try adding a Markdown code block with the language `twig` to any `.stories.mdx` file.

---

- Related to #1004 and #1240

/CC @gerardo-rodriguez

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,12 +1,17 @@
 import { Parser } from 'html-to-react';
 import { withPaddings } from 'storybook-addon-paddings';
 import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport';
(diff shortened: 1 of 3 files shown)


spaceninja (approved):
Oh, nice! I'm glad to see this is such a simple change. 🎉

Issue #1094, opened 2020-12-15
Toggle control pattern(s)


tylersticka:
Our PWA signup button (shown below articles) has two states for before and after notifications are active:

<img width="793" alt="Screen Shot 2020-12-15 at 10 15 08 AM" src="https://user-images.githubusercontent.com/69633/102257073-152fda80-3ec1-11eb-8a6a-e8c36f259b85.png">

<img width="791" alt="Screen Shot 2020-12-15 at 10 15 02 AM" src="https://user-images.githubusercontent.com/69633/102257087-1a8d2500-3ec1-11eb-9e50-c7ca2fb6c14a.png">

This design is fine (it re-uses some of our existing modifiers and other classes), but I can't help but think it might be a more elegant solution if we had a built-in pattern for this sort of thing that included built-in toggles for "on," "off" and maybe an in-between loading/processing state?

This may require more discussion to validate.

tylersticka:
Very relevant article: https://adrianroselli.com/2021/01/multi-function-button.html

(Hat tip to @spaceninja's [Friday Front-End](https://www.getrevue.co/profile/fridayfrontend/issues/friday-front-end-issue-278-the-devil-s-albatross-plus-responsive-html-video-multi-function-buttons-performance-and-more-308375))

gerardo-rodriguez:
> but I can't help but think it might be a more elegant solution if we had a built-in pattern for this sort of thing that included built-in toggles for "on," "off" and maybe an in-between loading/processing state?

Interesting!

@tylersticka I'd have to look at how the notifications button is implemented but I think it only has an "on" and "off" state. Was there an example you found on our site where the in-between loading/processing state exists? 

I'm still doing research to better understand the best way to create the most inclusive UX, but I believe depending on whether or not we need an in-between loading/processing state will impact the implementation.

tylersticka:
> Was there an example you found on our site where the in-between loading/processing state exists?

I'll admit I don't 100% remember what the sequence of events is for our notification subscription. If there's never a "waiting" state, then this isn't necessary now.

gerardo-rodriguez:
> I'll admit I don't 100% remember what the sequence of events is for our notification subscription. If there's never a "waiting" state, then this isn't necessary now.

After further research, I found there are only two states. There are two different notification buttons (one to turn on, the other to turn off notifications), their visibility gets toggled via adding/removing `u-hidden`.

gerardo-rodriguez:
[...]

PR #1264, opened 2021-06-04
Button component with icon

gerardo-rodriguez:
## Overview

This PR adds the ability to add icons to the Button component.

## Screenshots

Screen Shot 2021-06-04 at 5 12 40 PM

## Testing

1. Review the ["Button with icon" Button story](https://deploy-preview-1264--cloudfour-patterns.netlify.app/?path=/docs/components-button--icon)
2. Change the values of the `content_start_icon` and `content_end_icon` properties via the Storybook controls.

---

- GH issue #1094


gerardo-rodriguez:
An alternative component API design is to hardcode the Icon component into the Button component so that you only need to pass an icon string name instead.

Something like:
```twig
{% embed '@cloudfour/components/button/button.twig' with { iconStart: 'notifications' } %}
  {% block content %}Get notifications{% endblock %}
{% endembed %}
```
I just came up with this idea now. It seems a bit cleaner from an API design perspective, though, it is less flexible. What do you think, @tylersticka? 

Or maybe I can allow both to be an option? The `with { iconStart: 'notifications' }` allows us to handle basic use cases with all of our icons. If we needed custom icons (not available in our assets) then an `iconStart` block would allow us to pass in an `<svg>` element directly? 🤔

gerardo-rodriguez:
> * GH issue #1094

Is there a convention on this project for when one GH issue will have multiple PRs? I was curious if it was preferred to create a new "Button with icon" GH issue to track this PR separately or not. 🤔

tylersticka:
> An alternative component API design is to hardcode the Icon component into the Button component so that you only need to pass an icon string name instead.

I think that's a good idea, particularly since your solution requires the `inline` option on buttons. It would be easier to maintain if the button template took care of this.

> Or maybe I can allow both to be an option?

You could certainly do this. Something along the lines of this should work:

```twig
{% block start %}
  {% if icon_start %}
      {% include '@cloudfour/components/icon/icon.twig' with {
        name: icon_start
      } %}
  {% endif %}
{% endblock %}
```

gerardo-rodriguez:
> `icon_start`

I was trying to figure out the proper naming convention, are we using `icon_start` (underscore) instead of `iconStart` (camelCase)?

tylersticka:
@gerardo-rodriguez I just created this to help clarify that question: https://github.com/cloudfour/cloudfour.com-patterns/issues/1265

It felt wrong asking you to change it here because the button is one of the components that's using the wrong case, which is why I created a separate issue.

gerardo-rodriguez:
I've pushed up some updates! 🎉  

This PR is ready for review! 🔍 👀

src/components/button/button.twig
```diff
@@ -1,19 +1,36 @@
...
   {% endblock %}
-</{{tagName}}>
+  {% block icon_end %}
+    {% if icon_end %}
+      {% include '@cloudfour/components/icon/icon.twig' with {
+        name: icon_end,
+        inline: true
+      } %}
+    {% endif %}
+  {% endblock %}
```
gerardo-rodriguez, line 35: I opted to name both the block and the argument the same (e.g. `icon_end`).

This allows for the component consumer to use the `icon_end` both as an argument or as a block:
```twig
{% embed '@cloudfour/components/button/button.twig' with { icon_end: 'search' } %}
  {% block content %}Find it!{% endblock %}
{% endembed %}
```
Or:
```twig
{% embed '@cloudfour/components/button/button.twig' %}
  {% block content %}Button with a custom icon{% endblock %}
  {% block icon_end %}
    <svg viewBox="0 0 24 24" width="24" height="24" class="c-icon c-icon--inline">
      <circle r="8" cx="12" cy="12"></circle>
    </svg>
  {% endblock %}
{% endembed %}
```

I keep going back & forth on whether this is confusing or not. If both are used at the same time, the block wins.

Other ideas were to name the block something else (e.g. `icon_end_block`).

How do you all feel about this choice? 🤔

tylersticka: I'm open to differing points of view, but this is my take.

If I use the `icon_end` property (_not_ the block), the result _must_ be an icon. There is absolutely no way to make `icon_end` do anything other than display an icon at the end.

But that isn't really true of the block. Even though it's called `icon_end`, there is absolutely nothing stopping me from putting whatever I wanted to there... an image of a kitten, a bit of additional text, a `<span>`. Even though the block is _called_ `icon_end`, it's really just a placeholder for "stuff that follows the content."

So if the block's purpose is more generic than the property's, I'm a fan of naming the property something more specific than the block.

Happy to chat with voices or to consider other points of view on this! That's just my instinct.

gerardo-rodriguez: Excellent point, @tylersticka, thank you for your perspective!

I'll look at renaming the property and/or the block. 👍

gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1264/commits/2f712059b3fe0591b94d1cedff2aaf82bcc43ff3

gerardo-rodriguez:
Ready for another review! 🎉

src/components/button/button.twig
```diff
@@ -1,19 +1,36 @@
...
+  {% block content_end %}
+    {% if content_end_icon %}
+      {% include '@cloudfour/components/icon/icon.twig' with {
+        name: content_end_icon,
+        inline: true
+      } %}
+    {% endif %}
   {% endblock %}
-</{{tagName}}>
+</{{tag_name}}>
```
tylersticka, line 36: This is technically a breaking change, and a quick search of the this and the theme repo suggests the property is in use. Do you think we could save changes to the case of existing property names for a PR that handles all of those breakages in one swoop instead of bumping this from a feature to a breaking release?

I think it's a good idea for your new properties to use snake case to begin with.

.changeset/slow-planes-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
```
tylersticka, line 2: If this PR will introduce new features, I believe that makes this a `minor` release?

If this PR revises existing property name case, that makes it a `major` release.

https://semver.org/

src/components/button/button.twig
gerardo-rodriguez: Ah, good call. Thank you for catching this! 🔍 👀

.changeset/slow-planes-develop.md
gerardo-rodriguez: Changed, thank you! https://github.com/cloudfour/cloudfour.com-patterns/pull/1264/commits/82c294dbd697d9389f0f455a5021fd71b8e11a4e

src/components/button/button.twig
gerardo-rodriguez: Done! :https://github.com/cloudfour/cloudfour.com-patterns/pull/1264/commits/030174785617be60b3fc2fb77f82c96dbf6ec5a8

Paul-Hebert (approved):
This is looking good from a dev perspective @gerardo-rodriguez !

gerardo-rodriguez:
Thanks for the most recent feedback, @tylersticka. I have updated the PR. 😉

tylersticka (approved):

PR #1267, opened 2021-06-04
Add pseudo states add-on

tylersticka:
## Overview

Adds [storybook-addon-pseudo-states](https://storybook.js.org/addons/storybook-addon-pseudo-states), a rather magical add-on that spoofs pseudo states dynamically, allowing you to toggle various pseudo states in Storybook.

It isn't perfect. There's no API for toggling dev tools pseudo states, so it works by rewriting pseudo classes on the fly and re-applying those styles. This means it doesn't work for things like our `focus-visible` polyfill, because that's not using pseudo elements in the same way. (But that may be worth reconsidering in the future anyway.) But I believe I would find it very helpful for prototyping in the short-term, for documenting (or even testing?) designs in the long-term.

## Screenshots

pseudo

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -20,6 +20,7 @@ module.exports = {
     '@storybook/addon-a11y',
     'storybook-mobile',
     'storybook-addon-paddings',
+    'storybook-addon-pseudo-states',
     '@whitespace/storybook-addon-html',
     '@storybook/addon-postcss',
   ],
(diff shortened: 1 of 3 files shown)


tylersticka:
While this add-on is really cool, through the course of local development I've noticed a negative impact to performance. I'm not sure the trade-off is worth it, so I'm going to close this PR.

PR #1284, opened 2021-06-07
Broaden reach of alignment classes

### LINKED ISSUE #1270: Gutenberg wide and full alignment should apply to more than just images
WordPress themes can optionally support "wide" and "full" alignment for Gutenberg blocks.

"Wide" alignment means that the element will meet the edges of the viewport up to a certain point (so it won't get _too_ large but will still extend past the content edges):

Screen Shot 2021-06-04 at 3 38 25 PM

Whereas "full" alignment means that the element will _always_ meet the edges of the viewport:

Screen Shot 2021-06-04 at 3 38 31 PM

These use the classes `alignwide` and `alignfull`.

Currently, these classes are supported for the Image block (you can toggle them in the "Controls" tab): https://v-next--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-gutenberg--core-image&args=alignment:alignfull

But there are other blocks that support these alignment classes, including (but not necessarily limited to):

- Cover Image
- Gallery
- Pull Quote
- Video
- Table
- Columns
- Categories
- Embed

This means that instead of applying these transformations to `.wp-block-image` directly, we should probably make it a little more broad.

[This article](https://weblines.com.au/gutenberg-blocks-wide-alignment-full-width/) and [this one](https://jasonyingling.me/implementing-gutenberg-image-alignment-options-theme/#startertheme) both fence the styles inside of `.entry-content`, but that's just a class used by the default starter theme.

It might make more sense for us to use a wildcard class:

```scss
[class*="wp-block-"] {
  &.alignwide {
    // ...
  } 
  &.alignfull {
    // ...
  } 
}
```

Or alternatively we could recommend [an additional wrapper](https://codepen.io/webmandesign/post/gutenberg-full-width-alignment-in-wordpress-themes), though that would require coordination with our theme for compatibility.

tylersticka:
## Overview

Applies `alignfull` and `alignwide` to any `wp-block-*` class.

## Testing

Confirm that alignment control of [Gutenberg image story](https://deploy-preview-1284--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-gutenberg--core-image) still functions as it did before.

---

- Fixes #1270

### DIFF
--- a/.changeset/slimy-clocks-laugh.md
+++ b/.changeset/slimy-clocks-laugh.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add support for `alignfull` and `alignwide` to any Gutenberg block
--- a/src/vendor/wordpress/_wordpress.scss
+++ b/src/vendor/wordpress/_wordpress.scss
@@ -11,6 +11,34 @@
 $wp-button-gap: size.$spacing-list-inline-gap;
 
 /**
+ * Apply to any current or future WordPress block that supports these alignment
+ * options (unless they decide to update their class name prefixes!)
+ */
+[class^='wp-block-'],
+[class*=' wp-block-'] {
+  /**
+   * Releases a full-width child element from its limited width container
+   * @see https://cloudfour.com/thinks/breaking-out-with-viewport-units-and-calc/#flipping-the-script
+   */
+  &.alignfull,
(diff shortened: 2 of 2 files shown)


calebeby (approved):
🚢

Paul-Hebert (approved):
:+1:

PR #1285, opened 2021-06-07
Generate source code previews

### LINKED ISSUE #1004: Improve source code previews in Storybook

calebeby:
## Overview

Closes #1004 - @tylersticka is it fine to close that issue or should we leave it open until we manually go through and update whichever stories need manual source code snippets?

## Screenshots

Screen Shot 2021-06-07 at 2 07 18 PM

## Testing

Different cases to test the "show code" functionality:

- [Args passed directly from storybook into template](https://deploy-preview-1285--cloudfour-patterns.netlify.app/?path=/docs/components-icon--basic)
- [Args modified or replaced before passed into template](https://deploy-preview-1285--cloudfour-patterns.netlify.app/?path=/docs/objects-embed--picture)
- [Dynamic args that can be modified with controls](https://deploy-preview-1285--cloudfour-patterns.netlify.app/?path=/docs/objects-bunch--of-avatars)
- [Story that uses useEffect](https://deploy-preview-1285--cloudfour-patterns.netlify.app/?path=/docs/components-sky-nav--dark) (preview is not automatically generated, same behavior as on `v-next`)


.storybook/preview.js
```diff
@@ -67,6 +68,22 @@ export const parameters = {
...
+      try {
+        const storyFunction = storyContext.getOriginal();
+        const rendered = storyFunction(storyContext.args);
+        // The twing/source-inputs-loader.js file makes it so that whenever twig templates are rendered,
+        // the arguments and input path are stored in the window.__twig_inputs__ variable.
+        // __twig_inputs__ is a map between the output HTML and and objects with the arguments and input paths
+        // Here, since we have the rendered HTML, we can look up what the arguments and path were
+        // that correspond to that output
+        const input = window.__twig_inputs__?.get(rendered);
+        if (!input) return src;
```
Paul-Hebert, line 81: Are we confident that `input` will always have a `path` and `args`? If not, we may want to check for those here as well?

.storybook/preview.js
```diff
@@ -67,6 +68,22 @@ export const parameters = {
...
     prepareForInline: (storyFn) => htmlToReactParser.parse(storyFn()),
+    transformSource(src, storyContext) {
+      try {
+        const storyFunction = storyContext.getOriginal();
+        const rendered = storyFunction(storyContext.args);
+        // The twing/source-inputs-loader.js file makes it so that whenever twig templates are rendered,
+        // the arguments and input path are stored in the window.__twig_inputs__ variable.
+        // __twig_inputs__ is a map between the output HTML and and objects with the arguments and input paths
+        // Here, since we have the rendered HTML, we can look up what the arguments and path were
+        // that correspond to that output
```
Paul-Hebert, line 79: This is really interesting! Clever!

src/make-twig-include.js
```diff
@@ -0,0 +1,13 @@
...
+ * {% include '@cloudfour/components/button.twig' with { "type": "disabled" } only %}
+ * @param {string} path
+ * @param {any} args
+ */
+export const makeTwigInclude = (path, args) => {
+  const argsString =
+    Object.keys(args).length > 0
+      ? ` with ${JSON.stringify(args, null, 2)}`
+      : '';
+  return `{% include '${path}'${argsString} only %}`;
```
Paul-Hebert, line 12: Is there ever a use case where we wouldn't want to use `only`? 🤔 

If so, we might want to add a param to control this.

.storybook/preview.js
Paul-Hebert: Never mind, I just saw that we're defining those properties ourselves in `twing/source-inputs-loader.js`

CONTRIBUTING.md
```diff
@@ -73,3 +73,27 @@ This is generally not necessary, but in case you need to manually publish a vers
 You can run `npm publish --dry-run` to see everything that _would_ happen during publish, without actually publishing to the npm registry.
 
 Note the branch is `v-next` for now. When we we merge this branch to `master`, these instructions should be updated.
+
+## Overriding source code previews in Storybook
+
+For most stories, we are able to generate a twig source code snippet for Storybook to display automatically. In some edge cases, the source code snippet cannot be generated automatically, so the JS that was passed to `<Story>` is shown instead. In those cases, you can manually pass the source code to `<Story>`:
```
Paul-Hebert, line 79: Do we know what those edge cases are? If so it might be nice to spell them out here so future devs can know whether their use case is one of those edge cases or not.

Paul-Hebert (approved):
This is really clever @calebeby !

I assume it's a known short-coming that many of the code previews are loading a Storybook documentation-specific `/demo/` twig partial? (I'm not sure what we could do about that besides providing a custom code sample for those stories so I don't think that should be a blocker.)

LGTM!

.storybook/preview.js
calebeby: Yep!

.storybook/preview.js
calebeby: Thanks! This is the second approach, the first approach was to have the twig function return an object like this:

```js
{
  toString() {
    return '...html...'
  },
  path: '...',
  args: { ... }
}
```

which I think is a nicer solution, and it didn't have the limitation of not working for the stories which use useEffect / other hooks. But then I realized it breaks for the canvas page, because the canvas page has a typeof check and it doesn't allow non-strings (even though an object with a toString method can be implicitly coerced into a string)

src/make-twig-include.js
calebeby: I think it is best practice for includes to avoid implicitly passing through variables. For our docs, I put the `only` there so that if someone copy-pastes it it will probably "do the right thing", and they have to opt-in to the implicit passing through behavior. I can't think of any of our demos where someone using our patterns would always or usually want to have the implicit passing through behavior.

CONTRIBUTING.md
calebeby: As far as I know, the only case is if you use useEffect or another hook. But I don't know for sure if there could be other cases where the story function can't be called from the transformSource function

src/make-twig-include.js
Paul-Hebert: Yeah that makes sense :+1:

CONTRIBUTING.md
calebeby: I'll make a note about hooks causing it

src/make-twig-include.js
tylersticka: I think that makes sense. We can always revisit later, and this is probably a safer starting point than leaving it off.

calebeby:
> I assume it's a known short-coming that many of the code previews are loading a Storybook documentation-specific /demo/ twig partial? (I'm not sure what we could do about that besides providing a custom code sample for those stories so I don't think that should be a blocker.)

Yeah I talked through this with Tyler and it sounds like it is fine that it shows the include for the demo, and we can override the story source in those cases

tylersticka:
> is it fine to close that issue or should we leave it open until we manually go through and update whichever stories need manual source code snippets?

@calebeby I think it's fine to close that issue, this satisfies the intent for sure

tylersticka (commented):
Deferring final approval to @cloudfour/dev, this is a great step forward!

twing/webpack-options.js
```diff
@@ -5,12 +5,18 @@ const twingVirtualEnvironmentPath = require.resolve(
 
 const twingLoader = {
   test: /\.twig$/,
-  use: {
-    loader: 'twing-loader',
-    options: {
-      environmentModulePath: twingEnvironmentNodePath,
+  use: [
+    // Webpack loaders run in reverse, so this top loader will run last
```
gerardo-rodriguez, line 9: Thanks for the comment, @calebeby! Does that mean we want/need this loader to run last? In other words, if a new loader gets added in the future, should we ensure the new loader is below this loader? If so, let's also leave a note about this detail as well, thanks! 🙂

calebeby: It has to run after the twing loader, but not necessarily last. It sort of depends on what another loader is doing

calebeby: I put that comment there not to say that this specific loader has to run _last_, but to remind me or other people that webpack's ordering is opposite of what you'd expect (I forget literally every time)

Paul-Hebert: > I put that comment there not to say that this specific loader has to run last, but to remind me or other people that webpack's ordering is opposite of what you'd expect (I forget literally every time)

I also forget every time, so this note is very much appreciated!

Paul-Hebert (approved):

Issue #1269, opened 2021-06-04
Gutenberg groups with backgrounds should have rounded corners


tylersticka:
Our contained Card patterns (see #1254) establish rounded corners on cards.

These should be applied to our Gutenberg Group styles when `has-background` is present: https://v-next--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-gutenberg--core-group

<img width="326" alt="Screen Shot 2021-06-04 at 3 32 14 PM" src="https://user-images.githubusercontent.com/69633/120869012-37d6ae00-c54a-11eb-8ffe-cc1ab2d7cc1b.png">

tylersticka:
This is blocked by implementation of contained Card patterns. By waiting until those are done, we can insure consistency by sharing the same tokens and/or mixins between these things.

PR #1297, opened 2021-06-09
Gutenberg corners

AriannaChau:
## Overview

This PR adds a border-radius to the Gutenberg Group styles when `has-background` class is present.

## Screenshots

Screen Shot 2021-06-09 at 11 52 10 AM

## Testing

1. [Deploy link](https://deploy-preview-1297--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-gutenberg--core-group)

---

- Fixes https://github.com/cloudfour/cloudfour.com-patterns/issues/1269

### DIFF
--- a/src/tokens/size/border.json
+++ b/src/tokens/size/border.json
@@ -5,6 +5,10 @@
         "small": { "value": "0.125em" },
         "medium": { "value": "0.25em" },
         "large": { "value": "0.5em" },
+        "card-contained": {
+          "value": "{size.border.radius.large.value}",
+          "comment": "Current value for the contained cards."
+        },
         "full": {
           "value": "9999px",
           "comment": "A value of 50% would be more intuitive, but results in unexpectedly oblong rounding for non-square shapes. We use `px` to save the browser the trouble of recalculating the stupidly large value."
--- a/src/tokens/size/border.json
+++ b/src/tokens/size/border.json
@@ -5,10 +5,7 @@
         "small": { "value": "0.125em" },
         "medium": { "value": "0.25em" },
         "large": { "value": "0.5em" },
-        "card-contained": {
-          "value": "{size.border.radius.large.value}",
-          "comment": "Current value for the contained cards."
-        },
+        "card-contained": { "value": "{size.border.radius.large.value}" },
         "full": {
           "value": "9999px",
           "comment": "A value of 50% would be more intuitive, but results in unexpectedly oblong rounding for non-square shapes. We use `px` to save the browser the trouble of recalculating the stupidly large value."
--- a/src/vendor/wordpress/_wordpress.scss
+++ b/src/vendor/wordpress/_wordpress.scss
@@ -266,14 +266,14 @@ $wp-button-gap: size.$spacing-list-inline-gap;
 
 .wp-block-group.has-background {
   border-radius: size.$border-radius-card-contained; /* 1 */
-}
 
-.wp-block-group.has-background.alignfull {
-  border-radius: 0; /* 2 */
-}
+  &.alignfull {
+    border-radius: 0; /* 2 */
+  }
 
-.wp-block-group.has-background.alignwide {
-  @media (max-width: breakpoint.$l) {
-    border-radius: 0; /* 3 */
+  &.alignwide {
+    @media (max-width: breakpoint.$l) {
+      border-radius: 0; /* 3 */
+    }
   }
 }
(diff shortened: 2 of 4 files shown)


src/vendor/wordpress/_wordpress.scss
```diff
@@ -251,3 +251,12 @@ $wp-button-gap: size.$spacing-list-inline-gap;
   line-height: size.$height-control-multiline; /* 1 */
   padding: size.$padding-control-vertical 0;
 }
+
+/**
+ * Groups
+ * Rounds the corners for groups when the has-background class is applied.
```
tylersticka, line 257: Can we add a note about this being consistent with the contained card border radii?

src/vendor/wordpress/_wordpress.scss
```diff
@@ -251,3 +251,12 @@ $wp-button-gap: size.$spacing-list-inline-gap;
...
+ */
+
+.wp-block-group.has-background {
+  border-radius: size.$border-radius-large;
```
tylersticka, line 261: Do you think we should create a token specifically for card radii so the connection between the two is even more clear? (That might be overkill, just asking.)

tylersticka:
One other thought: We should probably not round the corners if the group is `alignfull` or `alignwide` below a certain size

src/vendor/wordpress/_wordpress.scss
AriannaChau: I wasn't going to because we already have a large border-radius with the same value and didn't want things to be too repetitive. Now that I'm looking at other tokens that doesn't seem to be a problem so I went ahead and added it.

src/vendor/wordpress/_wordpress.scss
```diff
@@ -251,3 +251,29 @@ $wp-button-gap: size.$spacing-list-inline-gap;
...
   padding: size.$padding-control-vertical 0;
 }
+
+/**
+ * Groups
+ * Rounds the corners for groups when the has-background class is applied.
+ * 1. This value is added to keep the Cloud Four card containers and
+ * the Gutenberg styles consistent.
+ * 2. When the alignfull class is applied the border radius will go
+ * back to 0 to keep the corners flesh with the edge of the page.
```
Paul-Hebert, line 261:
```suggestion
 * back to 0 to keep the corners flush with the edge of the page.
```

Nit: typo

src/tokens/size/border.json
```diff
@@ -5,6 +5,10 @@
         "small": { "value": "0.125em" },
         "medium": { "value": "0.25em" },
         "large": { "value": "0.5em" },
+        "card-contained": {
+          "value": "{size.border.radius.large.value}",
+          "comment": "Current value for the contained cards."
```
Paul-Hebert, line 10: Nit: I'm not sure if this comment is adding much value?

Paul-Hebert (commented):
This looks good to me! Deferring approval to Tyler since he opened this issue.

src/vendor/wordpress/_wordpress.scss
```diff
@@ -251,3 +251,29 @@ $wp-button-gap: size.$spacing-list-inline-gap;
...
+.wp-block-group.has-background {
+  border-radius: size.$border-radius-card-contained; /* 1 */
+}
+
+.wp-block-group.has-background.alignfull {
+  border-radius: 0; /* 2 */
+}
+
+.wp-block-group.has-background.alignwide {
+  @media (max-width: breakpoint.$l) {
```
tylersticka, line 276: Unfortunately this media query will take effect one pixel too late, because technically we're saying "take effect until anything _larger_ than `breakpoint.$l`."

Luckily, we have the [postcss-media-minmax](https://github.com/postcss/postcss-media-minmax) polyfill added so you should be able to revise this line like so:

```suggestion
  @media (width < breakpoint.$l) {
```

src/vendor/wordpress/_wordpress.scss
```diff
@@ -251,3 +251,29 @@ $wp-button-gap: size.$spacing-list-inline-gap;
...
+
+.wp-block-group.has-background {
+  border-radius: size.$border-radius-card-contained; /* 1 */
+}
+
+.wp-block-group.has-background.alignfull {
+  border-radius: 0; /* 2 */
+}
+
+.wp-block-group.has-background.alignwide {
```
tylersticka, line 275: Can we use nesting to keep this source a bit more DRY?

```scss
.wp-block-group.has-background {
  border-radius: ...;

  &.alignfull {
    // ...
  }

  &.alignwide {
    // ...
  }
}
```

tylersticka (approved):

PR #1298, opened 2021-06-09
Toggling buttons, CSS styles

gerardo-rodriguez:
## Overview

This PR introduces styling for when a button receives the `aria_pressed` argument with a value of `'true'` or `'false'`. 

This PR focuses on CSS and template changes only. My next follow-up PR will focus on the JS-specific effort.

## Screenshots

Screen Shot 2021-06-14 at 3 44 18 PM
Screen Shot 2021-06-14 at 3 44 34 PM

Icons were fine-tuned as well (notice they are a bit larger).
- Left: before
- Right: after

Screen Shot 2021-06-14 at 3 42 45 PM

## Testing

Preview: https://deploy-preview-1298--cloudfour-patterns.netlify.app/?path=/docs/components-button--toggling-button#toggling-button

1. Review the **Toggling button** stories across various browsers. 
2. Also check the "Dark" theme versions

---

- See  #1094


src/components/button/button.stories.mdx
```diff
@@ -105,9 +111,48 @@ The `content_start`/`content_end` blocks override the
...
+  <Story name="Custom Icon Toggling Button" args={{ aria_pressed: 'true' }}>
+    {(args) => iconButtonCustomDemo(args)}
+  </Story>
+</Canvas>
+
 ## Template Properties
 
-- `aria_expanded` (boolean): Sets the button's `aria-expanded` attribute.
+- `aria_expanded` (string, `'true'`/`'false'`): Sets the button's `aria-expanded` attribute.
+- `aria_pressed` (string, `'true'`/`'false'`): Sets the button's `aria-pressed` attribute.
```
gerardo-rodriguez, line 155: Thoughts on these being `string` values? I'm now questioning whether these should be `boolean` values. I'd need to add extra logic in the template, but it might make more sense.

src/components/button/button.twig
```diff
@@ -1,10 +1,27 @@
...
+{# Set a default icon if aria-pressed is set and no icon (or block content) was supplied #}
+{% if
+  aria_pressed and
+  content_start_icon is empty and
+  content_end_icon is empty and
+  _content_start|trim is empty and
+  _content_end|trim is empty
+%}
+  {% set content_start_icon = 'notifications' %}
+{% endif %}
```
gerardo-rodriguez, line 15: If the `aria_pressed` value stays as a `string` value, I might want to check specifically for `'true'` and/or `'false'`. Otherwise, a value of `asdf` (or any random string) will add the notifications icon, which doesn't seem correct.

src/mixins/_button.scss
```diff
@@ -17,6 +17,9 @@
 @include theme.props() {
   --theme-color-background-button-default: #{color.$brand-primary};
   --theme-color-background-button-secondary: transparent;
+  --theme-color-background-button-icon-slash-default: #{color.$brand-primary};
+  --theme-color-background-button-icon-slash-secondary: transparent;
+  --theme-color-background-button-icon-slash-tertiary: transparent;
```
gerardo-rodriguez, line 22: I'm unsure of the best way to handle this. The value of `transparent` doesn't actually do what I want it to. You'll notice that the extra slash is not visible in the secondary and tertiary buttons.

Screen Shot 2021-06-09 at 4 29 48 PM

My attempted solution before this was using `clip-path` with a custom `polygon` but there was a [Safari issue I was unable to resolve](https://cloudfour.slack.com/archives/C04AEEVK2/p1623178062068100).

Any ideas? 🤔

calebeby: @gerardo-rodriguez What do you think about a solution where there is a variant of each slashable icon with the slash in it (rather than adding the slash in CSS)? That way the slash is positioned nicely even for icons where the ideal position of the slash is not centered

gerardo-rodriguez: @calebeby Good idea! I was trying to create a flexible solution that didn't require multiple SVG icons but if this solution isn't up-to-par, then having two SVG icons may be the only way to handle it. 

I'll be curious how our design team feels about this. 🔍 👀

src/components/button/button.stories.mdx
tylersticka: Twig supports [ternary operators](https://twig.symfony.com/doc/3.x/templates.html#other-operators) if you did want to do this:

```twig
<button aria-pressed="{{aria_pressed ? 'true' : 'false'}}">
  ...
</button>
```

src/mixins/_button.scss
tylersticka: Hmm... 🤔 

Let me try something and get back to you.

src/components/button/button.stories.mdx
Paul-Hebert: > Thoughts on these being string values? I'm now questioning whether these should be boolean values. I'd need to add extra logic in the template to convert the value from a boolean to a string, but it might make more sense.

Good question... I'm fine with these being strings, since I think they're treated as strings by assistive technologies (e.g. I think `aria-pressed` !== `aria-pressed="true"`)

If we did switch them to booleans they'd need to be nullish booleans so that you can have the three different states: `null`, `true`, `false`. I don't know how Twig works well enough to know if there's an easy way to do that in Twig.

src/components/button/button.twig
Paul-Hebert: :+1: Yeah I think we'll want to check the value of `aria_pressed`

src/mixins/_button.scss
Paul-Hebert: I'm also curious how the design team feels about it!

Hand-rolled slashes might look nicer in certain contexts but would be more work.

>I'm unsure of the best way to handle this. The value of transparent doesn't actually do what I want it to. You'll notice that the extra slash is not visible in the secondary and tertiary buttons.

Hmm, should the extra slash's color be set to the background's current background? (I guess if the button has a transparent background that wouldn't work...)

I'm not opposed to revisiting the clip-path solution if this isn't working well. If we did that I wonder if there's a way to share SVG code between the clip path and the visible stroke? If that's possible we could DRY things up, and ensure the primary stroke and secondary stroke render the same way.

tylersticka: Okay, a few thoughts after poking around in this...

1. I'm curious if you ever tried using flex positioning instead of `inline-block` for the icon and if that would have resolved the Safari inconsistency?
2. I tried to experiment with using an SVG for this in a dynamic way, but I wasn't very happy with the results since they would require managing `defs` with `id` attributes that don't conflict: https://codepen.io/tylersticka/pen/783b2974d446edf1a8db02652e73afef
3. Supporting on/off icons is fine, but I'm really skeptical about `clip-path` not working when it did on a recent client project… 🤔

src/components/button/button.stories.mdx
gerardo-rodriguez: > If we did switch them to booleans they'd need to be nullish booleans so that you can have the three different states: `null`, `true`, `false`. I don't know how Twig works well enough to know if there's an easy way to do that in Twig.

Yeah, I figured I'd start with strings so that I didn't have to get too caught up with nullish boolean values within the context of Twig. I could explore it a bit more, but I'm still unsure if it's worth the extra time or not. Part of me feels it's better to match the native HTML experience (where the `aria-pressed` attribute expects string values of `'true'` or `'false'`).

src/mixins/_button.scss
gerardo-rodriguez: > Supporting on/off icons is fine, but I'm really skeptical about clip-path not working when it did on a recent client project… 🤔

I was confused by this as well! 😳 And that's why [I asked for help](https://cloudfour.slack.com/archives/C04AEEVK2/p1623178062068100), I wasn't sure if I was doing something incorrectly. 😅 

I'm going to take another look. It seems silly that it worked before but wasn't working for me this time. ¯\\\_(ツ)_/¯

> I'm curious if you ever tried using flex positioning instead of `inline-block` for the icon and if that would have resolved the Safari inconsistency?


I can look into this, I'm not sure I tried this. 🔍 

Thank you! 😄

Paul-Hebert: LOL, never mind. I started a Codepen and realized the clip paths you want are opposites for the two strokes.

Paul-Hebert: Anyways, here's my hacky codepen in case it's helpful: https://codepen.io/phebert/pen/ZEeMwWm

gerardo-rodriguez: > LOL, never mind. I started a Codepen and realized the clip paths you want are opposites for the two strokes.

@Paul-Hebert Can you elaborate what you mean by "are opposite for the two strokes"? 🤔

tylersticka: @gerardo-rodriguez I think what @Paul-Hebert means is that you can't use the stroke as a mask for the icon because that will mask the icon _within_ the stroke. What we actually need is to mask everything _outside_ of the stroke.

gerardo-rodriguez: Thanks, @tylersticka, I see.

I'm having a very hard time figuring this out. [This is my latest attempt](https://github.com/cloudfour/cloudfour.com-patterns/pull/1298/commits/cedd9e08785294cb9ece0b572a8dfa1bb124941f) but it's now harder for me to figure out the proper shape of the slash and Safari still renders differently from other browsers. 😕 

https://deploy-preview-1298--cloudfour-patterns.netlify.app/?path=/docs/components-button--toggling-button

Is anyone available to jump on a call? I am spinning my wheels. 😞

src/components/button/button.twig
gerardo-rodriguez: Udpated! https://github.com/cloudfour/cloudfour.com-patterns/pull/1298/commits/22724ac138627e739cc651043201d6d8b02f2e48

src/components/button/button.stories.mdx
gerardo-rodriguez: For now, I'm going to leave this as string values since there wasn't a strong argument otherwise. 😉

src/mixins/_button.scss
gerardo-rodriguez: Thanks for all your help, @Paul-Hebert, and @tylersticka!

I've pushed up my latest proposal, here's what it looks like:

Screen Shot 2021-06-11 at 4 10 45 PM

I ended up implementing a more _hacky_ solution to get Safari to render closer to the design intention (using `-webkit` prefix). The `supports` `view-box` route didn't produce the expected results for me, unfortunately.

tylersticka: This is good progress, Gerardo. Unfortunately, I'm still noticing some odd visual issues with this approach. There are subpixel artifacts around the slash, and the discrepancy between the slash size and the shadow size is pretty significant.

I'm going to come back with some clearer direction of what we should try next so you know where to go from here.

tylersticka: @gerardo-rodriguez I submitted a PR with a prototype approach that gets us to something I feel is a bit closer: #1302 

Happy to chat more on Monday!

gerardo-rodriguez: Thanks so much, @tylersticka!

I've updated this PR with the latest changes! 😄

gerardo-rodriguez:
This PR is ready for another review. 🔍 👀

src/mixins/_button.scss
```diff
@@ -89,6 +91,62 @@
...
+        -5% 87%,
+        -5% 101.4711%,
+        105% 37.9626%,
+        105% 105%,
+        -5% 105%
+      );
+
+      & > * {
+        clip-path: $clip-path;
+        transform: translate(0, 0);
```
gerardo-rodriguez, line 141: @tylersticka The [prototype had `height: 100%` and `width: 100%`](https://github.com/cloudfour/cloudfour.com-patterns/pull/1302/files#diff-0c4e5bc28b5fd976b8b3719d6811a920ca3c6e930db1b7839220eb386e809d92R50-R52) but I removed them here because the icons would blow up when I zoomed the browser in or out. 😮 🤔 ¯\\\_(ツ)_/¯

icons

tylersticka: Whoa!! Neat! I wonder why?

gerardo-rodriguez: It was very interesting! Even more so that the icons wouldn't reset back to their original size when I zoomed back to 100%. 🤯

tylersticka, line 141: Can we add a comment mentioning that this is to fix a bug where some browsers would render the element as slightly offset from others?

src/mixins/_button.scss
```diff
@@ -89,6 +91,62 @@
...
+
+        @supports (clip-path: view-box) {
```
tylersticka, line 143: Can we add a comment explaining that this helps normalize the box the clipping path renders to?

src/mixins/_button.scss
```diff
@@ -89,6 +91,62 @@
     transform: none;
   }
 
+  .c-button__extra {
+    display: flex;
+    font-size: ms.step(1);
+    margin: 0 ms.step(-4);
```
tylersticka, line 97: Can we create a design token for `ms.step(-4`)? Maybe `size.$spacing-gap-button-extra`?

tylersticka (commented):
Looking good! A couple of small suggestions inline. Thanks for being so willing to collaborate on this, @gerardo-rodriguez.

src/mixins/_button.scss
gerardo-rodriguez: Thanks, @tylersticka! Should I also create a design token for `ms.step(1)`?

tylersticka: @gerardo-rodriguez That's probably a good idea, but I didn't have a clear suggestion on what it should be called so I decided not to suggest it. 😆 

Maybe we need to create a set of icon size tokens, with the default being modular scale step 0 and `medium` or `large` or something for this? I'm open to suggestions.

gerardo-rodriguez: LOL Got it, thanks! I'll think a bit more about this! 😉

src/mixins/_button.scss
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1298/commits/44bf076b5a171a9c727f021becf49d5f12f32a18

src/mixins/_button.scss
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1298/commits/44bf076b5a171a9c727f021becf49d5f12f32a18

src/mixins/_button.scss
gerardo-rodriguez: > Can we create a design token for `ms.step(-4`)? Maybe `size.$spacing-gap-button-extra`?

Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1298/commits/5fcb4fa6baffe2d3e174051703522456391e7042

gerardo-rodriguez: > Maybe we need to create a set of icon size tokens, with the default being modular scale step 0 and `medium` or `large` or something for this? I'm open to suggestions.

Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1298/commits/884cb0a7c198322b017029b8b96ee21d6bd279cd

gerardo-rodriguez:
This PR has been updated and is ready for another review. 👍

tylersticka (approved):
LGTM!

Issue #1289, opened 2021-06-07
Refine source examples in individual stories


tylersticka:
Our code previews to date were developed before we had anything close to useful source previews.

As of #1285, we not only have really nice automatic source previews, but we also have a new utility function we can use to generate previews more manually in cases where our demo source _must_ be disconnected somewhat from the actual theme code.

This issue represents going through our existing stories and making the necessary changes to improve the source previews where possible. Changes may include:

- Simplifying demos that may be replaced with the actual template and `args`
- Specifying the `Story` element's `source` or `templateSource` parameters using the aforementioned utility function
- When a story is particularly complex, unique or time-consuming, adding a TODO comment or creating a separate issue to refine at a later date

gerardo-rodriguez:
Am I correct in understanding the source code previews can't handle `block` content? For example, the [Card component](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/components-card--content-blocks) has multiple content blocks, is this an example of a component we leave alone for now?

tylersticka:
I don't think there's any component we couldn't address in some fashion using one of the bullets in this issue's description. If all else fails, you can manually specify a string (`{% block 'content' %}...{% endblock %}`) representing the block.

gerardo-rodriguez:
@tylersticka I created #1447 to break out the work left for the Objects. 

I will go ahead and call this card "done". 😉

PR #1300, opened 2021-06-12
Make the Author story code example better/update component

gerardo-rodriguez:
## Overview

This PR updates the Author component to not explicitly call functions for the properties.

## Screenshots

Left: Before
Right: After

Screen Shot 2021-06-11 at 4 55 18 PM

## Testing

1. Review https://deploy-preview-1300--cloudfour-patterns.netlify.app/?path=/docs/components-author--default-story
2. Confirm no regressions

---

- See #1289

### DIFF
--- a/.changeset/silver-dodos-turn.md
+++ b/.changeset/silver-dodos-turn.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Update Author component story code example
--- a/.changeset/silver-dodos-turn.md
+++ b/.changeset/silver-dodos-turn.md
@@ -1,5 +1,0 @@
----
-'@cloudfour/patterns': patch
----
-
-Update Author component story code example
--- a/src/components/author/author.stories.mdx
+++ b/src/components/author/author.stories.mdx
@@ -58,19 +58,9 @@ Displays an author, or set of authors, for article bylines.
 
   ```ts
   {
-    name: string;
-    avatar: string;
-    link: string;
-  }
-  ```
-
-  or
-
-  ```ts
-  {
-    name: () => string;
-    avatar: () => string;
-    link: () => string;
+    name: string | () => string;
+    avatar: string | () => string;
+    link: string | () => string;
   }
   ```
 
--- a/src/components/author/author.stories.mdx
+++ b/src/components/author/author.stories.mdx
@@ -3,7 +3,6 @@ import template from './author.twig';
 import gerardoImage from './demo/gerardo.png';
 import tylerImage from './demo/tyler.png';
 import meganImage from './demo/megan.png';
-import { makeTwigInclude } from '../../make-twig-include';
 const gerardo = {
   name: 'Gerardo Rodriguez',
   avatar: gerardoImage,
--- a/src/components/author/author.stories.mdx
+++ b/src/components/author/author.stories.mdx
@@ -3,6 +3,7 @@ import template from './author.twig';
 import gerardoImage from './demo/gerardo.png';
 import tylerImage from './demo/tyler.png';
 import meganImage from './demo/megan.png';
+import { makeTwigInclude } from '../../make-twig-include';
 const gerardo = {
   name: () => 'Gerardo Rodriguez',
   avatar: () => gerardoImage,
@@ -41,7 +42,30 @@ const megan = {
 Displays an author, or set of authors, for article bylines.
 
 <Canvas>
-  <Story name="Default" args={{}}>
+  <Story
+    name="Default"
+    parameters={{
+      docs: {
+        source: {
+          code: makeTwigInclude('@cloudfour/components/author/author.twig', {
+            authors: [
+              {
+                name: 'Megan Notarte',
+                avatar: meganImage,
+                link: 'https://cloudfour.com/is/megnotarte',
+              },
+              {
+                name: 'Tyler Sticka',
+                avatar: tylerImage,
+                link: 'https://cloudfour.com/is/tyler',
+              },
+            ],
+            date: new Date('March 31, 2021'),
+          }),
+        },
+      },
+    }}
+  >
     {(args) =>
       template({
         authors: [gerardo, tyler, megan].slice(0, args.count),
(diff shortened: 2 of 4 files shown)


src/components/author/author.stories.mdx
```diff
@@ -41,7 +42,30 @@ const megan = {
...
-  <Story name="Default" args={{}}>
+  <Story
+    name="Default"
+    parameters={{
+      docs: {
+        source: {
+          code: makeTwigInclude('@cloudfour/components/author/author.twig', {
+            authors: [
+              {
+                name: 'Megan Notarte',
```
calebeby, line 53: @gerardo-rodriguez I don't think this code example would work, because each of the properties are supposed to be a function, so when twig tries to call author.name(), it can't call a string

.changeset/silver-dodos-turn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Update Author component story code example
```
calebeby, line 5: I don't think a changeset is necessary because this does not affect the published package, so no version release is necessary

src/components/author/author.stories.mdx
calebeby: I wonder if there is a way to specify an inline function in twig syntax

gerardo-rodriguez: Oh...I didn't catch that was calling a function, interesting! I hadn't noticed this pattern in other components. What's the benefit of setting it up as a function versus just passing a value directly?

calebeby: Tyler suggested being able to pass Timber user objects directly, and [most of the properties on those objects are methods](https://timber.github.io/docs/reference/timber-user/#overview)

gerardo-rodriguez: Ah, thank you, @calebeby! TIL 🎓 

> I wonder if there is a way to specify an inline function in twig syntax

🤔

calebeby: We could also change the author component to allow `string | () => string`

gerardo-rodriguez: > We could also change the author component to allow `string | () => string`

Good idea! 💡 

> I wonder if there is a way to specify an inline function in twig syntax

I can't seem to figure this out. 

I'm now not sure how to proceed. If we do decide to change the template logic, that effort seems out of scope for this PR where the intent was only to update the code example.

Hmm... 🤔 

I'll leave this PR open a bit longer but if no one else has any other suggestions for how to approach this component, I'll close the PR. I'll then leave a comment in https://github.com/cloudfour/cloudfour.com-patterns/issues/1289 pointing out this component needs a second look.

gerardo-rodriguez: @cloudfour/dev Any other ideas for how to handle this? If not, I'll close this PR and leave this story as-is.

calebeby: I have no ideas other than to make the story accept strings. But I don't know how to check `typeof` to see if a value is a string or function in twig, so I don't know if that is possible

tylersticka: Sorry I'm late to the party here.

In Twig, if you output `{{something}}` and `something` is a property, it will output the property. But if `something` is a function, it will call the function.

This behavior is described under the "Implementation" heading in [this section of the Twig docs](https://twig.symfony.com/doc/3.x/templates.html#variables):

> For convenience's sake foo.bar does the following things on the PHP layer:
> - check if foo is an array and bar a valid element;
> - if not, and if foo is an object, check that bar is a valid property;
> - if not, and if foo is an object, check that bar is a valid method (even if bar is the constructor - use __construct() instead);
> - if not, and if foo is an object, check that getBar is a valid method;
> - if not, and if foo is an object, check that isBar is a valid method;
> - if not, and if foo is an object, check that hasBar is a valid method;
> - if not, return a null value.

This is why you'll see in Timber examples `author.name` referenced and not `author.name()` even though `name` is a method: https://github.com/timber/starter-theme/blob/14605ce1782b7645e6cc5be6df69e0cab30fc480/templates/single.twig#L11

tylersticka (dismissed):

src/components/author/author.stories.mdx
gerardo-rodriguez: Thanks, @tylersticka, I appreciate your response!

If I'm understanding correctly, we don't need the [`author.twig` component to call the functions](https://github.com/cloudfour/cloudfour.com-patterns/blob/188a4ebdc187df071de499fc65431ede1f8bf9bf/src/components/author/author.twig#L14), then?

It currently uses:
- `author.link()`
- `author.name()`
- `author.avatar()`

If so, I can update the template in this PR (plus the code example if needed). Updating the template is _technically_ tangentially related but I'd rather just make the update and take care of it now. 🙂

tylersticka: > If I'm understanding correctly, we don't need the [`author.twig` component to call the functions](https://github.com/cloudfour/cloudfour.com-patterns/blob/188a4ebdc187df071de499fc65431ede1f8bf9bf/src/components/author/author.twig#L14), then?

Correct. In fact, dropping the parentheses would probably make these templates more flexible.

gerardo-rodriguez:
This PR has been updated and is ready for another review.

tylersticka (dismissed):

src/components/author/author.stories.mdx
```diff
@@ -56,6 +56,16 @@ Displays an author, or set of authors, for article bylines.
 - `class`: Append a class to the root element.
 - `authors`: Array of [author objects](https://timber.github.io/docs/reference/timber-user/#properties) of the type:
 
+  ```ts
+  {
+    name: string;
```
calebeby, line 61: @gerardo-rodriguez Would it be more clear to have the "or" inline?

```
    name: string | () => string;
```

calebeby (approved):

src/components/author/author.stories.mdx
gerardo-rodriguez: Thanks, @calebeby!

https://github.com/cloudfour/cloudfour.com-patterns/pull/1300/commits/9778e03e064c88a036730f63cc660780c6b5abbd

Can you approve the PR one more time so I can merge? Thank you! 🙂

calebeby (approved):

calebeby:
Oh you could add a changeset now for the changes to the component

gerardo-rodriguez:
> Oh you could add a changeset now for the changes to the component

@calebeby Oh, right, good call! 😝

calebeby (approved):

PR #1302, opened 2021-06-12
Add dynamic toggle icon button prototype

tylersticka:
## Overview

This PR introduces a prototype of how dynamic toggle icons should look within toggle buttons. It is intended to be a point of reference for #1094 and #1298.

If this PR's designs are reflected in #1298 prior to this PR's approval, it can be safely closed. Because this applies prototype styles to `.c-button` classes directly, it is likely to break eventually anyway, so no need to be precious with it.

Some changes from what is in #1298 as of this writing:

- The icons are larger visually, which I believe makes them a bit more balanced.
- The button content was given a wrapper to make its alignment to the adjacent iconography more predictable.
- The spacing has been tweaked between icons and content and the sides of the button.
- The `clip-mask` polygon has been updated. I attempted to mimic the layout of the pseudo-element slash in Illustrator, then expanded to a `100px` square grid, then used the point values to determine the percentages of the `polygon`.
- `@supports` is used instead of `-webkit-clip-mask`.
- A `transform` hack is used to force Safari to play a little nicer. I avoided a `translateZ` or `will-change` hack to avoid spinning up the GPU unnecessarily.

This appears to work mostly consistently across the latest versions of Firefox, Chrome, Safari (Mac/iOS) and Edge.

## Screenshots

Screen Shot 2021-06-11 at 5 04 31 PM

File

---

/CC @gerardo-rodriguez @Paul-Hebert

### DIFF
--- a/src/prototypes/buttons/buttons.stories.js
+++ b/src/prototypes/buttons/buttons.stories.js
@@ -1,0 +1,11 @@
+import toggleIconsPrototype from './toggle-icons.twig';
+import './toggle-icons.scss';
+
+export default {
+  title: 'Prototypes/Buttons',
+  parameters: {
+    docs: { page: null },
+  },
+};
+
+export const ToggleIcons = () => toggleIconsPrototype({});
--- a/src/prototypes/buttons/toggle-icons.scss
+++ b/src/prototypes/buttons/toggle-icons.scss
@@ -1,0 +1,59 @@
+@use '../../compiled/tokens/scss/size';
+@use '../../mixins/ms';
+
+#tcs-button-toggle-icons {
+  .c-button {
+    align-items: center;
+    display: inline-flex;
+    vertical-align: middle;
+  }
+
+  .c-button__extra {
+    display: flex;
+    font-size: ms.step(1);
+    margin: 0 ms.step(-4);
+    position: relative;
+
+    &:first-child {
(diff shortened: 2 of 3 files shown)


gerardo-rodriguez:
Thanks, @tylersticka! I'm going to take a closer look at the CSS in this draft PR to help with #1298. 

One thing I'd like to confirm, you don't need a review of this PR since it's still in "draft" mode, correct?

tylersticka:
> One thing I'd like to confirm, you don't need a review of this PR since it's still in "draft" mode, correct?

@gerardo-rodriguez It was left in draft mode since tests still hadn't passed when I left on Friday, thanks for pinging me about that!

We can either:
- Approve this PR and remove the prototype later after we no longer need it.
- Use this PR as a point of reference for your work on #1298 and close it when we no longer need it.

I have no preference.

gerardo-rodriguez:
Thanks, @tylersticka!

I'm good with this: 
> Use this PR as a point of reference for your work on #1298 and close it when we no longer need it.

Thank you! 😄

tylersticka:
@gerardo-rodriguez I'm going to close this since it seems to have translated nicely to #1298!

Deploy preview will remain here: https://deploy-preview-1302--cloudfour-patterns.netlify.app/

PR #1318, opened 2021-06-17
[BLOCKED] Toggling Button, JavaScript logic

### LINKED ISSUE #1094: Toggle control pattern(s)

gerardo-rodriguez:
## Overview

This PR adds JavaScript logic to the Toggling Button pattern so that the `aria-pressed` attribute updates when clicking the butotn.

## Screenshots

toggling

## Testing

Preview: https://deploy-preview-1318--cloudfour-patterns.netlify.app/?path=/story/components-button--toggling-button

1. Review the toggling button stories and confirm they toggle.
2. Review the updated documentation.

---

- Closes #1094

### DIFF
--- a/src/components/button/button.test.ts
+++ b/src/components/button/button.test.ts
@@ -4,6 +4,7 @@ import { withBrowser } from 'pleasantest';
 import { loadTwigTemplate } from '../../../test-utils';
 
 const buttonMarkup = loadTwigTemplate(path.join(__dirname, './button.twig'));
+// Helper to test the JS-driven toggling button
 const initTogglingButtonJS = (
   utils: PleasantestUtils,
   togglingButton: ElementHandle
@@ -39,7 +40,7 @@ test(
 );
 
 test(
-  'should be a button when aria_pressed is set',
+  'should render a button when aria_pressed is set',
   withBrowser(async ({ utils, screen }) => {
     await utils.injectHTML(
       await buttonMarkup({ tag_name: 'a', aria_pressed: 'false' })
--- a/src/components/button/button.test.ts
+++ b/src/components/button/button.test.ts
@@ -25,7 +25,7 @@ test(
     await utils.injectHTML(await buttonMarkup({}));
 
     const button = await screen.getByRole('button');
-    await expect(button).toBeVisible();
+    await expect(button).toHaveTextContent(/^hello world$/i);
   })
 );
 
@@ -35,7 +35,7 @@ test(
     await utils.injectHTML(await buttonMarkup({ href: '#' }));
 
     const linkButton = await screen.getByRole('link');
-    await expect(linkButton).toBeVisible();
+    await expect(linkButton).toHaveTextContent(/^hello world$/i);
   })
 );
 
@@ -45,11 +45,15 @@ test(
     await utils.injectHTML(
       // The `aria_pressed` value should override the `tag_name` value and always
       // render a button element. This test verifies that expecation.
-      await buttonMarkup({ tag_name: 'a', aria_pressed: 'false' })
+      await buttonMarkup({
+        tag_name: 'a',
+        aria_pressed: 'false',
+        label: 'I am a button',
+      })
     );
 
     const button = await screen.getByRole('button');
-    await expect(button).toBeVisible();
+    await expect(button).toHaveTextContent(/^i am a button$/i);
   })
 );
 
--- a/src/components/button/button.test.ts
+++ b/src/components/button/button.test.ts
@@ -1,0 +1,64 @@
+import path from 'path';
+import type { ElementHandle, PleasantestUtils } from 'pleasantest';
+import { withBrowser } from 'pleasantest';
+import { loadTwigTemplate } from '../../../test-utils';
+
+const buttonMarkup = loadTwigTemplate(path.join(__dirname, './button.twig'));
+const initTogglingButtonJS = (
+  utils: PleasantestUtils,
+  togglingButton: ElementHandle
+) =>
+  utils.runJS(
+    `
+    import { initTogglingButton } from './toggling-button'
+    export default (togglingButton) => initTogglingButton(togglingButton)
+    `,
+    [togglingButton]
+  );
+
+test(
+  'should render a button element by default',
+  withBrowser(async ({ utils, screen }) => {
+    // If I don't pass _something_ to buttonMarkup(), I get an error.
+    // To get around that, I pass an empty object.
+    await utils.injectHTML(await buttonMarkup({}));
+
+    const button = await screen.getByRole('button');
+    await expect(button).toBeVisible();
+  })
+);
+
+test(
+  'should render an anchor element when href is provided',
+  withBrowser(async ({ utils, screen }) => {
+    await utils.injectHTML(await buttonMarkup({ href: '#' }));
+
+    const linkButton = await screen.getByRole('link');
+    await expect(linkButton).toBeVisible();
+  })
+);
+
+test(
+  'should be a button when aria_pressed is set',
+  withBrowser(async ({ utils, screen }) => {
+    await utils.injectHTML(
+      await buttonMarkup({ tag_name: 'a', aria_pressed: 'false' })
+    );
+
+    const button = await screen.getByRole('button');
+    await expect(button).toBeVisible();
+  })
+);
+
+test(
+  'should toggle aria-pressed state',
+  withBrowser(async ({ utils, screen }) => {
+    await utils.injectHTML(await buttonMarkup({ aria_pressed: 'false' }));
+    // await utils.loadCSS('../../../dist/standalone.css');
+
+    const togglingButton = await screen.getByRole('button');
+    await initTogglingButtonJS(utils, togglingButton);
+
+    await expect(togglingButton).toBeVisible();
+  })
+);
--- a/src/components/button/button.test.ts
+++ b/src/components/button/button.test.ts
@@ -52,13 +52,15 @@ test(
 
 test(
   'should toggle aria-pressed state',
-  withBrowser(async ({ utils, screen }) => {
+  withBrowser(async ({ utils, screen, user }) => {
     await utils.injectHTML(await buttonMarkup({ aria_pressed: 'false' }));
-    // await utils.loadCSS('../../../dist/standalone.css');
-
     const togglingButton = await screen.getByRole('button');
     await initTogglingButtonJS(utils, togglingButton);
 
-    await expect(togglingButton).toBeVisible();
+    await expect(togglingButton).toHaveAttribute('aria-pressed', 'false');
+    await user.click(togglingButton);
+    await expect(togglingButton).toHaveAttribute('aria-pressed', 'true');
+    await user.click(togglingButton);
+    await expect(togglingButton).toHaveAttribute('aria-pressed', 'false');
   })
 );
--- a/src/components/button/button.test.ts
+++ b/src/components/button/button.test.ts
@@ -59,8 +59,10 @@ test(
     await initTogglingButtonJS(utils, togglingButton);
 
     await expect(togglingButton).toHaveAttribute('aria-pressed', 'false');
+
     await user.click(togglingButton);
     await expect(togglingButton).toHaveAttribute('aria-pressed', 'true');
+
     await user.click(togglingButton);
     await expect(togglingButton).toHaveAttribute('aria-pressed', 'false');
   })
--- a/src/components/button/button.test.ts
+++ b/src/components/button/button.test.ts
@@ -20,9 +20,7 @@ const initTogglingButtonJS = (
 test(
   'should render a button element by default',
   withBrowser(async ({ utils, screen }) => {
-    // If I don't pass _something_ to buttonMarkup(), I get an error.
-    // To get around that, I pass an empty object.
-    await utils.injectHTML(await buttonMarkup({}));
+    await utils.injectHTML(await buttonMarkup());
 
     const button = await screen.getByRole('button');
     await expect(button).toHaveTextContent(/^hello world$/i);
--- a/src/components/button/button.test.ts
+++ b/src/components/button/button.test.ts
@@ -43,6 +43,8 @@ test(
   'should render a button when aria_pressed is set',
   withBrowser(async ({ utils, screen }) => {
     await utils.injectHTML(
+      // The `aria_pressed` value should override the `tag_name` value and always
+      // render a button element. This test verifies that expecation.
       await buttonMarkup({ tag_name: 'a', aria_pressed: 'false' })
     );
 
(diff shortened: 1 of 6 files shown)


gerardo-rodriguez:
>  Review the toggling button stories in the **Canvas** view to see the toggling in action.

I was fighting Storybook and gave up. The toggling JS functionality only works in the **Canvas** view and I don't understand why. 

¯\\\_(ツ)_/¯

tylersticka:
> I was fighting Storybook and gave up. The toggling JS functionality only works in the **Canvas** view and I don't understand why.

@gerardo-rodriguez Did you try adding `inline={false}` to the `Story` element? This will display the story in its own iframe instead of inline.

gerardo-rodriguez:
> > I was fighting Storybook and gave up. The toggling JS functionality only works in the **Canvas** view and I don't understand why.
> 
> @gerardo-rodriguez Did you try adding `inline={false}` to the `Story` element? This will display the story in its own iframe instead of inline.

I did not, I'll give it a try. Thanks, @tylersticka! TIL 🎓 😄

gerardo-rodriguez:
> @gerardo-rodriguez Did you try adding `inline={false}` to the `Story` element? This will display the story in its own iframe instead of inline.

@tylersticka The layout gets all funny, but the JS works now! 🙃 

Screen Shot 2021-06-17 at 4 35 16 PM

tylersticka:
> The layout gets all funny, but the JS works now! 🙃

If you add `isColumn` to `Canvas` (`<Canvas isColumn>`) it will put them in a stack instead of trying to arrange them side by side (which it has a harder time doing when they're in iframes). They'll also probably be closer to the size they were before because their frames will have wider viewports.

tylersticka (commented):
Deferring approval to @cloudfour/dev, but from a design standpoint this works great!

src/components/button/button.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    `,
+    [togglingButton]
+  );
+
+test(
+  'should render a button element by default',
+  withBrowser(async ({ utils, screen }) => {
+    // If I don't pass _something_ to buttonMarkup(), I get an error.
+    // To get around that, I pass an empty object.
+    await utils.injectHTML(await buttonMarkup({}));
```
calebeby, line 25: @gerardo-rodriguez Can we change the function returned by `loadTwigTemplate` to not require a parameter?

src/components/button/button.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    const linkButton = await screen.getByRole('link');
+    await expect(linkButton).toBeVisible();
+  })
+);
+
+test(
+  'should render a button when aria_pressed is set',
+  withBrowser(async ({ utils, screen }) => {
+    await utils.injectHTML(
+      await buttonMarkup({ tag_name: 'a', aria_pressed: 'false' })
```
calebeby, line 46: What does the `tag_name` do in this case?

src/components/button/button.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    );
+
+    const button = await screen.getByRole('button');
+    await expect(button).toBeVisible();
```
calebeby, line 50: @gerardo-rodriguez I'm wondering whether the visibility check is really needed here (and in the other tests). `getByRole` acts as its own assertion, since it will throw if no matching element is found, so I don't think the visibility check is necessary.

Maybe another assertion that would be helpful is to check that the button has the right text passed through using `expect(button).toHaveTextContent(/.../)`?

src/components/button/toggling-button.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Create Toggling Button
+ * Adds an event listener to toggle the `aria-pressed` value
+ *
+ * @param togglingButton - The button to apply the toggling functionality to
+ */
+export const initTogglingButton = (togglingButton: HTMLButtonElement) => {
```
calebeby, line 7: Will it always be initialized with a `<button>` or are there cases where the server-rendered HTML could render on an `<a>` or another element?

src/components/button/toggling-button.ts
```diff
@@ -0,0 +1,32 @@
...
+  // Handler for when a toggling button is clicked
+  const onTogglingButtonClick = () => {
+    // Get the current aria-pressed state and toggle it
+    const isAriaPressed = !(
+      togglingButton.getAttribute('aria-pressed') === 'true'
+    );
```
calebeby, line 13:
```suggestion
    const isAriaPressed =
      togglingButton.getAttribute('aria-pressed') !== 'true';
```

src/components/button/toggling-button.ts
```diff
@@ -0,0 +1,32 @@
...
+  const destroy = () => {
+    togglingButton.removeEventListener('click', onTogglingButtonClick);
+  };
+
+  // Intialize
+  const init = () => {
+    togglingButton.addEventListener('click', onTogglingButtonClick);
+  };
+
+  init();
```
calebeby, line 28: I don't think we need a function here since it is single-use. We could inline the addEventListener call. The comment is enough to help understand what is going on

calebeby (changes requested):
I'm super excited that you got to write Pleasantest tests! :tada:

I left a couple of comments but overall this is great!

calebeby:
@gerardo-rodriguez I have a functionality question: Would it make sense for this JS to handle changing out the text of the button depending on the toggle state? Are there use cases for that? (I am thinking of subscribe/unsubscribe button potentially). Would that best be handled in the JS (maybe via a data attribute) or in CSS?

gerardo-rodriguez:
> @gerardo-rodriguez I have a functionality question: Would it make sense for this JS to handle changing out the text of the button depending on the toggle state? Are there use cases for that? (I am thinking of subscribe/unsubscribe button potentially). Would that best be handled in the JS (maybe via a data attribute) or in CSS?

Great question and a great catch, @calebeby!

Yes, we actually would want the text to update so this is something I need to consider and think about more. Thank you!

src/components/button/button.test.ts
gerardo-rodriguez: Maybe! I gave it a quick look but other tests started failing (if I recall correctly) when I did that so I left it alone.

I can take a second look today, though. 😉

src/components/button/button.test.ts
gerardo-rodriguez: @calebeby My intent was to test that `aria_pressed` has a higher priority over setting the `tag_name` value in the template logic. The element should be a `button` anytime `aria_pressed` is present and valid.

calebeby: Oh, that makes sense. Can you put a comment there?

src/components/button/button.test.ts
gerardo-rodriguez: Thanks, @calebeby, appreciate the feedback!

src/components/button/toggling-button.ts
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1318/commits/b0c54dc7e78a8acb220710ffb7c9079876f3689b

src/components/button/toggling-button.ts
gerardo-rodriguez: @calebeby I introduced [template logic with the intent to only render a `<button>` element](https://github.com/cloudfour/cloudfour.com-patterns/pull/1318/files#diff-7db3e7fdff7b358ebbc9d41aa97a09ae3e5d9ce13ed04e9d21e5f5c0a454fc6aR6) when the `aria_pressed` value is set. 

So in theory, there shouldn't be cases where an `<a>` or another element are rendered. 

Can you elaborate a bit more on your question? Did you spot an opportunity to improve the logic? 🔍 👀

src/components/button/button.test.ts
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1318/commits/ff473746650ae1f6d1dd8c1b720376448b8a8122

src/components/button/button.test.ts
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1318/commits/b66277c3db90e23bbef3b1b3085194c1f24f8d7d

src/components/button/toggling-button.ts
calebeby: Got it! 👍 I wasn't sure if there were any cases where we would use an `<a>` for the toggle button but it sounds like there aren't.

src/components/button/toggling-button.ts
gerardo-rodriguez: Thanks for calling this section out, I wasn't happy with it and found it confusing. I like your suggestion but also it still feels a bit confusing to me.

I decided to refactor it, do you feel this is more straightforward? https://github.com/cloudfour/cloudfour.com-patterns/pull/1318/commits/7a73ff30441c4bc16e8cdc06f822e4e3dfa76085

src/components/button/button.test.ts
gerardo-rodriguez: @calebeby Hmm...I must be doing something wrong. 

When I modify the function returned by `loadTwigTemplate()` as follows:

Screen Shot 2021-06-21 at 5 16 06 PM

I get this error: 

Screen Shot 2021-06-21 at 5 15 53 PM

🤔

src/components/button/toggling-button.ts
calebeby: Yes that makes more sense 👍

src/components/button/button.test.ts
calebeby: Hmm, when I tried that locally, all the tests still passed:

Screen Shot 2021-06-22 at 9 09 35 AM

gerardo-rodriguez: @calebeby Very odd! 😮  

Do you have time to jump on a call today after lunch? I don't understand why it works for you but it fails for me. 😅

calebeby: Yeah sounds good

gerardo-rodriguez:
> @gerardo-rodriguez I have a functionality question: Would it make sense for this JS to handle changing out the text of the button depending on the toggle state? Are there use cases for that? (I am thinking of subscribe/unsubscribe button potentially). Would that best be handled in the JS (maybe via a data attribute) or in CSS?

I am going to do more research. Yesterday I ended my day thinking this made sense but now after thinking about it overnight, I'm not so sure. 🤔 

[This seems relevant](https://inclusive-components.design/toggle-button/#changinglabels): 
> As a rule of thumb, you should never change pressed state and label together. If the label changes, the button has already changed state in a sense, just not via explicit WAI-ARIA state management.

But what does that mean for us and the notifications button? I'll report back after doing further research but at the moment I see two possible outcomes. 

1. We end up with two different toggling patterns

OR 

2. We redesign the UI/UX for the notifications button

Those may or may not be the best way to move forward. I'll update this PR with my findings.

src/components/button/button.test.ts
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1318/commits/52abc2de54cd81880d6a6895a4e4e486bc31f857

For those following, I reinstalled dependencies and it magically resolved itself. ¯\\\_(ツ)_/¯

gerardo-rodriguez:
@cloudfour/dev @tylersticka Thoughts on the information below?

## Original implementation

If I follow the [original implementation](https://github.com/cloudfour/cloudfour.com-patterns/issues/1094#issue-767994616), there is a conflict between the visual information and audible information.

Button UI state | Visually | Audibly
-|-|-
🔔 Get notifications | **Current state:** "off"<br><br>**I see:** Notifications are off.<br><br>**Expect:** If I press it, I will turn notifications on. | **Current state:** "off"<br><br>**I hear:** _"Get notifications, off, switch"_ <br><br>**Expect:** If I press it, I will turn notifications on.
🔕 Turn off notifications | **Current state:** "on"<br><br>**I see:** Notifications are on.<br><br>**Expect:** If I press it, I will turn notifications off. | **Current state:** "on"<br><br>**I hear:** _"Turn off notifications, on, switch"_  🤔 <br><br>**Expect:** "Turn off notifications" is set to "on", so if I press it, I will turn notifications on. 🤔 

- Changing the label makes the state of the button confusing from an audible perspective.

## Proposed solution

Button UI state | Visually | Audibly
-|-|-
Notifications 🔔  | **Current state:** "on"<br><br>**I see:** Notifications are on <br><br>**Expect:** If I press it, I will turn notifications off  | **Current state:** "on"<br><br>**I hear:** _"Notifications, on, switch"_ <br><br>**Expect:** If I press it, I will turn notifications off.
Notifications 🔕  | **Current state:** "off"<br><br>**I see:** Notifications are off.<br><br>**Expect:** If I press it, I will turn notifications on. | **Current state:** "off"<br><br>**I hear:** _"Notifications, off, switch"_  <br><br>**Expect:** If I press it, I will turn notifications on. 

- Putting the icon second might help reinforce "Notifications" as a _label_ to the toggle state.
- The "on" state (`aria-checked="true"`) renders the icon with no slash (🔔).
- The "off" state (`aria-checked="false"`) renders the icon with a slash (🔕).

tylersticka:
@gerardo-rodriguez My concern is that these two solutions seem to invert the usability of the control.

In the original implementation, the voice experience is counter-intuitive, but the buttons are obvious: I know exactly what the button will do before clicking it.

In the proposed implementation, I would never guess as a sighted user that a button labeled "Notifications" was going to _subscribe_ me to something.

Is it possible that we've mistakenly cast this control as a toggle button? Is it actually one button to distinctly subscribe, and a separate button to distinctly unsubscribe?

Or is this a sign that a toggle button is not the right UI element? Should this be a checkbox or some sort of toggle switch, like a small settings panel?

This might be worth discussing with voices tomorrow (if possible).

gerardo-rodriguez:
> My concern is that these two solutions seem to invert the usability of the control.

Yes, I've struggled with this. I feel I'm having trouble providing a coherent UX both audibly and visually.

> Is it possible that we've mistakenly cast this control as a toggle button? Is it actually one button to distinctly subscribe, and a separate button to distinctly unsubscribe?

Maybe! 

> Or is this a sign that a toggle button is not the right UI element? Should this be a checkbox or some sort of toggle switch, like a small settings panel?

I've also started wondering this so it's worth discussing.

> This might be worth discussing with voices tomorrow (if possible).

Yep, let's discuss tomorrow, thanks! 🙂

gerardo-rodriguez:
For a paper trail, we are pausing this effort for the time being, refer to #1339.

gerardo-rodriguez:
Closing this PR as the goals have changed. See #1508.

PR #1351, opened 2021-06-29
Slightly refactor Rhythm, support Jetpack Markdown Block

### LINKED ISSUE #1314: Add CSS for relevant Jetpack Gutenberg blocks
In testing our new WordPress theme, I've noticed that [Jetpack](https://jetpack.com/) includes some complimentary Gutenberg blocks that seem quite useful. In most cases, they seem very close to working but may need some CSS tweaks to fit in our responsive containers.

## Tasks per block

- Use a local environment with Jetpack installed and Gutenberg active to source markup for the block
- Use this markup to create a story similar to our Gutenberg stories (but maybe housed in a separate file/area so it's clear that these are Jetpack blocks)?
- Write necessary CSS (if necessary) so that the block will work within our pages

## Setup tasks

- ~Determine method for including whatever CSS or JS is necessary for visualizing these stories. Ideally this would be a dependency we add to the project similar to what we're doing for the core Gutenberg blocks so we don't have to maintain it.~

## Blocks

This is not [the full list](https://jetpack.com/support/jetpack-blocks/), as not all blocks seem useful to us.

No need to create stories for blocks that work as-is without modification. Check those off and move on!

If there are major gotchas, hurdles or other issues with any of these blocks, we should make note of that and move on to the next simplest block. No reason to get hung up on any one of these.

- ~[Image Compare Block](https://jetpack.com/support/jetpack-blocks/image-compare-block/) (Does not appear to resize with container currently)~
- [x] [Markdown Block](https://jetpack.com/support/markdown/) (Needs space between adjacent block elements)

(I thought there'd be more than two we'd use prior to writing this issue. 😄)

tylersticka:
## Overview

See changelog and docs for details.

## Screenshots

Screen Shot 2021-06-29 at 1 38 55 PM

Screen Shot 2021-06-29 at 1 38 42 PM

## Testing

[On the deploy preview](https://deploy-preview-1351--cloudfour-patterns.netlify.app/):

- Review Rhythm Object docs for regressions
- Review Vendor WordPress Jetpack Markdown story

---

- Fixes #1314

### DIFF
--- a/src/mixins/_adjacent.scss
+++ b/src/mixins/_adjacent.scss
@@ -1,0 +1,29 @@
+@use "../compiled/tokens/scss/size";
+
+/// Applies styles for a vertical rhythm layout object. Used by the `o-rhythm`
+/// pattern and for relevant Gutenberg blocks. Makes heavy use of Heydon
+/// Pickering's "lobotomized owl" selectors: By adding `margin-top` to adjacent
+/// elements, we avoid impacting the layout of the container itself (which would
+/// complicate alignment of adjacent columns).
+/// @link https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/
+/// @param {Number} $space - The default space between.
+@mixin rhythm($space: size.$rhythm-default) {
+  & > * + * {
+    margin-top: $space;
+    // Allow space to be overridden using custom properties
+    margin-top: var(--rhythm);
+  }
+
+  // Where custom properties are supported, also allow headings to receive their
+  // own rhythm.
+  @supports (--custom: properties) {
+    & > * + h1,
+    & > * + h2,
+    & > * + h3,
+    & > * + h4,
+    & > * + h5,
+    & > * + h6 {
+      margin-top: var(--rhythm-headings);
+    }
+  }
+}
--- a/src/vendor/wordpress/jetpack-blocks.stories.mdx
+++ b/src/vendor/wordpress/jetpack-blocks.stories.mdx
@@ -8,7 +8,7 @@ WordPress developer Automattic provides a [Jetpack](https://jetpack.com/) plugin
 
 ## Markdown Block
 
-Reenforces the containing element's rhythm. If contained within a [Rhythm object](/docs/objects-rhythm--default-story), its children will inherit the same spacing.
+The markdown block applies vertical rhythm between its children. If contained within a [Rhythm object](/docs/objects-rhythm--default-story), the markdown block will inherit the same spacing. For example, if the Rhythm object uses the `o-rhythm--generous` modifier, the markdown block will inherit the generous spacing.
 
 See also: [Markdown Block documentation](https://jetpack.com/support/markdown/)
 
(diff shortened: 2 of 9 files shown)


src/mixins/_adjacent.scss
```diff
@@ -0,0 +1,29 @@
+@use "../compiled/tokens/scss/size";
```
Paul-Hebert, line 1: Should this be included in the changeset? Or is that not necessary since this is just an internal change? 🤔 

It's probably fine as-is, but figured I'd ask since I wasn't sure.

src/vendor/wordpress/jetpack-blocks.stories.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+<Meta title="Vendor/WordPress/Jetpack Blocks" />
+
+# Jetpack Blocks
+
+WordPress developer Automattic provides a [Jetpack](https://jetpack.com/) plugin that includes additional blocks. We support a selection of these.
+
+## Markdown Block
+
+Reenforces the containing element's rhythm. If contained within a [Rhythm object](/docs/objects-rhythm--default-story), its children will inherit the same spacing.
```
Paul-Hebert, line 11: >If contained within a [Rhythm object](/docs/objects-rhythm--default-story), its children will inherit the same spacing.

I'm a little confused by this. It appears that the CSS below will apply the rhythm mixin regardless of whether the markdown block is in a rhythm object or not? Am I missing something?

Paul-Hebert: Oh, do you mean that the level of spacing will be inherited via custom props? (e.g. if the parent has generous spacing then the markdown block will also have generous spacing?) That's pretty clever! I wonder if this comment could be rewritten somehow to make that clearer?

Paul-Hebert:
```suggestion
The markdown block applies vertical rhythm between its children. If contained within a [Rhythm object](/docs/objects-rhythm--default-story), the markdown block will inherit the same spacing. For example, if the Rhythm object uses the `rhythm--generous` modifier, the markdown block will inherit the generous spacing.
```

Or something...

Paul-Hebert (dismissed):
This looks great! The use of custom properties to make the markdown block inherit the rhythm spacing is clever!

I left a few doc-related comments inline. My only other critique is I found the mixin name a little confusing at first. Something about "adjacent" had me confused about where the rhythm would be applied. A few other ideas. Feel free to ignore if they don't jibe with you:

- `nested.rhythm`
- `child.rhythm`
- `inner.rhythm`
- `spacing.rhythm`
- `rhythm.apply`

src/mixins/_adjacent.scss
tylersticka: I don't think so since it's just an internal change.

tylersticka:
@Paul-Hebert I renamed the mixin from `adjacent.rhythm` to `spacing.vertical-rhythm`. This is ready for re-review!

Paul-Hebert (approved):
I like the new name! Looks great! 🚢

PR #1356, opened 2021-06-30
Refine the Dot Leader list code example

gerardo-rodriguez:
## Overview

This PR refines the code example for the Dot Leader list example.

## Screenshots

Screen Shot 2021-06-30 at 1 38 00 PM

## Testing

Preview: https://deploy-preview-1356--cloudfour-patterns.netlify.app/?path=/docs/components-dot-leader--list

1. Review the Dot Leader list code example for accuracy

---

- See #1289

### DIFF
--- a/src/components/dot-leader/dot-leader.stories.mdx
+++ b/src/components/dot-leader/dot-leader.stories.mdx
@@ -44,5 +44,29 @@ If `c-dot-leader` is an `<a>` element, it will gain some additional styles.
 Although `c-dot-leader` applies to a single row, the pattern works best as part of [a list object](/docs/objects-list--default-story). The dotted lines clearly associate each column even when content lengths vary.
 
 <Canvas>
-  <Story name="List">{listDemo({ topics })}</Story>
+  <Story
+    name="List"
+    parameters={{
+      docs: {
+        source: {
+          code: `{% embed '@cloudfour/objects/list/list.twig' only %}
+  {% block content %}
+    {% for topic in topics %}
+      <li>
+        {% include '@cloudfour/components/dot-leader/dot-leader.twig' with {
+          label: topic.title,
+          count: topic.count,
+          href: topic.url
+        } only %}
+      </li>
+    {% endfor %}
+  {% endblock %}
+{% endembed %}
+`,
+        },
+      },
+    }}
+  >
+    {listDemo({ topics })}
+  </Story>
 </Canvas>
--- a/src/components/dot-leader/dot-leader.stories.mdx
+++ b/src/components/dot-leader/dot-leader.stories.mdx
@@ -1,5 +1,6 @@
 import { Story, Canvas, Meta } from '@storybook/addon-docs/blocks';
 import listDemo from './demo/list.twig';
+// eslint-disable-next-line @cloudfour/import/no-webpack-loader-syntax
 import listDemoSource from '!!raw-loader!./demo/list.twig';
 import topics from './demo/topics.json';
 import template from './dot-leader.twig';
--- a/src/components/dot-leader/dot-leader.stories.mdx
+++ b/src/components/dot-leader/dot-leader.stories.mdx
@@ -1,7 +1,14 @@
 import { Story, Canvas, Meta } from '@storybook/addon-docs/blocks';
-import listDemo from './demo/list.twig';
+// The '!!raw-loader!' syntax is a non-standard, Webpack-specific, syntax.
+// See: https://github.com/webpack-contrib/raw-loader#examples
+// For now, it seems likely Storybook is pretty tied to Webpack, therefore, we are
+// okay with the following Webpack-specific raw loader syntax. It's better to leave
+// the ESLint rule enabled globally, and only thoughtfully disable as needed (e.g.
+// within a Storybook docs page and not within an actual component).
+// This can be revisited in the future if Storybook no longer relies on Webpack.
 // eslint-disable-next-line @cloudfour/import/no-webpack-loader-syntax
 import listDemoSource from '!!raw-loader!./demo/list.twig';
+import listDemo from './demo/list.twig';
 import topics from './demo/topics.json';
 import template from './dot-leader.twig';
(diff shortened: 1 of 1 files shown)


src/components/dot-leader/dot-leader.stories.mdx
```diff
@@ -44,5 +44,29 @@ If `c-dot-leader` is an `<a>` element, it will gain some additional styles.
...
+        {% include '@cloudfour/components/dot-leader/dot-leader.twig' with {
+          label: topic.title,
+          count: topic.count,
+          href: topic.url
+        } only %}
+      </li>
+    {% endfor %}
+  {% endblock %}
+{% endembed %}
+`,
```
tylersticka, line 65: @gerardo-rodriguez Unless I'm mistaken, this looks like it's exactly the contents of the `demo/list.twig` file.

In cases where you want to show the raw template source, it's possible to do so without copying and pasting code.

Near the beginning of the file, you should see the import for the demo itself:

```js
import listDemo from './demo/list.twig';
```

Duplicate this line, but give it a unique name and prepend `!!raw-loader!` to the path:

```js
import listDemo from './demo/list.twig';
import listDemoSource from '!!raw-loader!./demo/list.twig';
```

Then, you can use `listDemoSource` (or whatever you called it) as the value of `docs.source.code`. It should show the contents of the `list.twig` file without compiling them.

More info: https://github.com/webpack-contrib/raw-loader#examples

gerardo-rodriguez: What!? 🤯 

Thanks, @tylersticka, I'll take a look! 🔍 👀

gerardo-rodriguez: Done, thanks, @tylersticka! https://github.com/cloudfour/cloudfour.com-patterns/pull/1356/commits/e38b9e6f1b213f9c8bc9715e482a2231a80a22f2

gerardo-rodriguez: Oh...lint failure... 🔍

gerardo-rodriguez: I was getting a lint error. [I disabled it](https://github.com/cloudfour/cloudfour.com-patterns/pull/1356/commits/cd366c7cf4935fe0b7143c8672d0324f0cd40028). Not sure if disabling this specific use case is frowned upon. Do you have any insight, @tylersticka?

tylersticka (commented):
Deferring approval to @cloudfour/dev. This seems a lot more DRY than before, but it's unclear to me what the alternative solution would be to overcome the `eslint` rule, or if we should be updating our configuration to account for this use case.

src/components/dot-leader/dot-leader.stories.mdx
Paul-Hebert: Interesting! Here's the [info about the lint rule](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-webpack-loader-syntax.md), and the relevant bit:

> This syntax is non-standard, so it couples the code to Webpack. The recommended way to specify Webpack loader configuration is in a Webpack configuration file.

I get the intent here. `import` is a native feature that works anywhere. By using custom webpack syntax we're breaking that interoperability and tying ourselves to webpack. That said,  it seems unlikely that Storybook would move away from webpack any time soon, and since we're already tied to Storybook it seems like we're already tied to webpack? 

I think the worst case scenario is that a new Storybook version moves away from webpack and then we'd need to either:
- Add our own webpack config.
- Update anywhere we use this syntax to do something else.
- Not update Storybook.

That doesn't seem all that scary? I'd be more concerned about this lint rule being used in our actual components, but it seems fine for a Storybook docs page?

I'm trying to think through alternatives:

1. We could update our webpack config so that you could add a query string to specify a specific loader: (e.g. `import myRawModule from 'myModule?raw'`. I think this could be done using [resource query rules](https://webpack.js.org/configuration/module/#ruleresourcequery)? 
2. Write some custom node code to grab a file's contents and spit them out? That seems like reinventing the wheel for no reason and I'm not sure how it would work with Storybook.

It seems like option 1 above might be slightly cleaner since we're not tying ourselves to webpack as directly? But we'd still be relying on webpack to handle the resource query rules. That solution doesn't seem radically better but it sounds more complicated to get working.

My gut feeling is we should leave this rule enabled globally (since we shouldn't be doing this in our component code) but not worry about disabling it in Storybook doc files.

That said I'm curious if @calebeby or @spaceninja have any other thoughts.

Paul-Hebert (dismissed):

src/components/dot-leader/dot-leader.stories.mdx
tylersticka: Option 1 feels like we're just recreating a native Webpack feature ourselves for no reason, and option 2 does feel like reinventing the wheel for no reason.

> My gut feeling is we should leave this rule enabled globally (since we shouldn't be doing this in our component code) but not worry about disabling it in Storybook doc files.

I agree with this 👍

spaceninja: I agree with @Paul-Hebert — leave the rule enabled, and disable it for places where we're intentionally working around it like Storybook docs.

That said, I'd like to see a comment above the eslint-disable line explaining why we're disabling it. Just a sentence or two summarizing what we've said here.

gerardo-rodriguez: Thank you all! I'll add a comment shortly. 👍

gerardo-rodriguez: Comment added! https://github.com/cloudfour/cloudfour.com-patterns/pull/1356/commits/f6178df483e6acc6baf64c21b5bee1fa452ef4b4

gerardo-rodriguez:
@Paul-Hebert @spaceninja I've updated this PR, can I get a final review/approval? Thanks! 😄

Paul-Hebert (approved):
Looks good!

Issue #1238, opened 2021-05-24
Add comment thread pattern - dev.


AriannaChau:
We've added a prototype for the new comment thread designs, issue here: https://github.com/cloudfour/cloudfour.com-patterns/issues/487

They just now need to be turned into an actual pattern.

tylersticka:
Remaining effort for this is represented by these issues...

- #1389 
- https://github.com/cloudfour/cloudfour.com-wp/issues/515

PR #1373, opened 2021-07-08
Replace Message with Alert, use in Comment

tylersticka:
## Overview

The Comment component introduced in #1370 lacked a message for communicating when a comment had been submitted but was still in moderation. The live version of this feature used a component we called [Message](https://cloudfour-patterns.netlify.app/patterns/components/message.html), which had been recreated in #604. But this pattern's size made it appear more unwieldy in that context, and it didn't seem to respond to our themes. I also found its name somewhat confusing: A comment is a type of message, after all, so it felt a bit off to say "display a message inside of a comment."

So this PR removes Message, introduces a new Alert pattern, and integrates that into Comment for the moderation message feature.

## Screenshots

Screenshot 2021-07-08 at 15-38-10 Storybook

Screen Shot 2021-07-08 at 3 38 23 PM

Screen Shot 2021-07-08 at 3 44 11 PM

## Testing

On the deploy preview…

- [Alert](https://deploy-preview-1373--cloudfour-patterns.netlify.app/?path=/docs/components-alert--basic#alert)
- [Comment ▸ Unapproved](https://deploy-preview-1373--cloudfour-patterns.netlify.app/?path=/docs/components-comment--unapproved#unapproved)

---

- See #1238


Paul-Hebert (approved):
This looks good! 

My one question is whether the alert component should support HTML being passed in instead of text? (So we could include a link, icon, etc.)

If we want to support that use case we may want to replace the `message` property with a content slot. If we don't want to support that use case then this looks good to go!

tylersticka:
> My one question is whether the alert component should support HTML being passed in instead of text? (So we could include a link, icon, etc.)
> 
> If we want to support that use case we may want to replace the `message` property with a content slot. If we don't want to support that use case then this looks good to go!

@Paul-Hebert [Either is supported!](https://github.com/cloudfour/cloudfour.com-patterns/blob/ced0f4f359a00e305d279dd67df19a7a76a09b74/src/components/alert/alert.stories.mdx#template-blocks)

https://github.com/cloudfour/cloudfour.com-patterns/blob/ced0f4f359a00e305d279dd67df19a7a76a09b74/src/components/alert/alert.twig#L5-L11

Paul-Hebert:
Oh, got it. I'd missed that. Awesome! Looks good!

PR #1399, opened 2021-07-15
Add column modifier to list object

### LINKED ISSUE #1381: Column pattern
We added a new prototype of the [article listing page](https://v-next--cloudfour-patterns.netlify.app/?path=/story/prototypes-article-listing--example). In it, the dot leader pattern is split into multiple columns. We need either a modifier for the dot leader pattern or a new pattern that divides content into multiple columns.

**Large Screens**
Screen Shot 2021-07-09 at 11 47 23 AM

**Small Screens**
Screen Shot 2021-07-09 at 11 47 36 AM

dromo77:
## Overview

This PR adds a modifier so we can divide lists into multiple columns.

## Screenshots

Screen Shot 2021-07-15 at 11 56 42 AM

## Testing

1. [Deploy Preview](https://deploy-preview-1399--cloudfour-patterns.netlify.app/?path=/docs/objects-list--default-story#columns)
---

- Fixes #1381


src/objects/list/list.scss
```diff
@@ -40,3 +43,17 @@
...
+@for $i from 2 through 3 {
+  .o-list--#{$i}-column {
+    @include media-query.breakpoint-classes($from: s, $to: xl) {
+      columns: #{$i};
+      @include fluid.grid-gap(
+        breakpoint.$s,
+        breakpoint.$xl,
+        size.$spacing-gap-fluid-min,
+        size.$spacing-gap-fluid-max
+      );
```
tylersticka, line 56: Is it intentional that we're setting `grid-gap` rather than `column-gap` here?

tylersticka (commented):
Neat! One question inline

src/objects/list/list.scss
dromo77: No I missed this, thanks!

src/objects/list/list.stories.mdx
```diff
@@ -12,8 +13,26 @@ const tagNames = ['ul', 'ol'];
 
 <Meta
   title="Objects/List"
+  parameters={{ docs: { inlineStories: false } }}
   argTypes={{
     class: { type: { name: 'string' } },
+    columns: {
+      control: {
+        type: 'range',
+        min: 2,
```
tylersticka, line 22: Should `min` be `1`? Otherwise there isn't a way from the story controls to enable and then disable columns.

src/objects/list/demo/columns.twig
```diff
@@ -0,0 +1,15 @@
+{% if columns and columns > 1 and columnsBreakpoint and columnsBreakpoint != 'none' %}
```
tylersticka, line 1: Is it intentional that `columnsBreakpoint` is required? Is there no use case where someone would want columns on by default?

tylersticka: How would someone use multiple breakpoints?

```
o-list--2-columns o-list--3-columns@xl
```

dromo77: @tylersticka Would something like this work better?

```
{% if columns and columns > 1 %}
  {% set _list_class = 'o-list--' ~ columns ~ '-column' %}
{% elseif columns and columns > 1 and columnsBreakpoint %}
  {% set _list_class = 'o-list--' ~ columns ~ '-column' ~ columnsBreakpoint %}
{% else %}
  {% set _list_class = '' %}
{% endif %}
```

tylersticka: That would alleviate some of the problems, but if you wanted to do something like this:

```
o-list--2-columns@s  o-list--3-columns@m
```

I think it would prevent that?

(This is one of the reasons I've just been using a `class` property for defining the class attribute in some of my other patterns. It gets complicated pretty quickly!)

gerardo-rodriguez: > (This is one of the reasons I've just been using a class property for defining the class attribute in some of my other patterns. It gets complicated pretty quickly!)

Good idea! 👍 🙂

dromo77: Something like this?
```
args={{
 class: 'o-list--2-column@s o-list--3-column@m'
}}
```

PR #1406, opened 2021-07-16
Add "Overview" object for use on the Article Listing page

### LINKED ISSUE #1376: Index pattern
We added a new prototype of the [article listing page](https://v-next--cloudfour-patterns.netlify.app/?path=/story/prototypes-article-listing--example). There is a section to navigate articles by topic or search. We need a pattern that lays out the content in this section. There are 3 main areas:

- heading
- search
- main content

**Large Screens**
Screen Shot 2021-07-09 at 11 47 23 AM

**Small Screens**
Screen Shot 2021-07-09 at 11 47 36 AM

Paul-Hebert:
## Overview

This PR adds an "Overview" object which is used on the Article Listing pages to show a full list of topics and a search bar. The Article Listings prototype was updated to use this component

## Screenshots

Screen Shot 2021-07-16 at 12 32 02 PM
Screen Shot 2021-07-16 at 12 32 08 PM

## Testing

1. Compare the Article Listings prototype on the [current site](https://v-next--cloudfour-patterns.netlify.app/iframe.html?id=prototypes-article-listing--example&args=&viewMode=story) and the [preview deploy](https://deploy-preview-1406--cloudfour-patterns.netlify.app/iframe.html?id=prototypes-article-listing--example). They should look pretty much identical
2. Confirm on the [preview deploy](https://deploy-preview-1406--cloudfour-patterns.netlify.app/iframe.html?id=prototypes-article-listing--example)you can [use the VoiceOver rotor in Safari](https://support.apple.com/guide/voiceover/with-the-voiceover-rotor-mchlp2719/mac) and see there is a "More Topics region" listed in the Landmarks section.  (enable VoiceOver. press Option, Command, U. If necessary, use the arrow keys to navigate to the Landmarks section)
3. Confirm there are no regressions with [heading permalinks.](https://deploy-preview-1406--cloudfour-patterns.netlify.app/?path=/story/components-heading--permalinks)

---

- Fixes #1376

/CC @dromo77

### DIFF
--- a/.changeset/five-turkeys-double.md
+++ b/.changeset/five-turkeys-double.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': major
+---
+
+Renamed the Heading property from `permalink_id` to `id` and allowed you to use `id` when `permalink` is false.
--- a/src/objects/overview/overview.scss
+++ b/src/objects/overview/overview.scss
@@ -32,7 +32,7 @@
       grid-template-areas:
         'header header actions'
         'content content content';
-      grid-template-columns: repeat(3, minmax(max-content, 1fr));
+      grid-template-columns: repeat(3, 1fr);
       row-gap: ms.step(3);
     }
 
--- a/src/objects/overview/overview.scss
+++ b/src/objects/overview/overview.scss
@@ -17,38 +17,40 @@
   margin-bottom: ms.step(3);
 }
 
-@media (min-width: breakpoint.$l) {
-  .o-overview {
-    @include fluid.column-gap(
-      breakpoint.$s,
-      breakpoint.$xl,
-      size.$spacing-gap-fluid-min,
-      size.$spacing-gap-fluid-max
-    );
+@supports (display: grid) {
+  @media (min-width: breakpoint.$l) {
+    .o-overview {
+      @include fluid.column-gap(
+        breakpoint.$s,
+        breakpoint.$xl,
+        size.$spacing-gap-fluid-min,
+        size.$spacing-gap-fluid-max
+      );
 
-    align-items: center;
-    display: grid;
-    grid-template-areas:
-      'header header actions'
-      'content content content';
-    grid-template-columns: repeat(3, minmax(max-content, 1fr));
-    row-gap: ms.step(3);
-  }
+      align-items: center;
+      display: grid;
+      grid-template-areas:
+        'header header actions'
+        'content content content';
+      grid-template-columns: repeat(3, minmax(max-content, 1fr));
+      row-gap: ms.step(3);
+    }
 
-  .o-overview__header,
-  .o-overview__actions {
-    margin: 0;
-  }
+    .o-overview__header,
+    .o-overview__actions {
+      margin: 0;
+    }
 
-  .o-overview__header {
-    grid-area: header;
-  }
+    .o-overview__header {
+      grid-area: header;
+    }
 
-  .o-overview__actions {
-    grid-area: actions;
-  }
+    .o-overview__actions {
+      grid-area: actions;
+    }
 
-  .o-overview__content {
-    grid-area: content;
+    .o-overview__content {
+      grid-area: content;
+    }
   }
 }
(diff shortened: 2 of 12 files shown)


.changeset/five-turkeys-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': major
+---
+
+Renamed the Heading property from `permalink_id` to `id` and allowed you to use `id` when `permalink` is false.
```
Paul-Hebert, line 5: This is _technically_ a breaking change, though as far as I can tell we're not using`permalink_id` anywhere yet.

src/prototypes/article-listing/example/example.scss
```diff
@@ -74,31 +74,8 @@
     }
   }
 
-  .topics-header {
```
Paul-Hebert, line 77: This prototype CSS was replaced by the new Object CSS.

src/objects/overview/overview.scss
```diff
@@ -0,0 +1,56 @@
...
+
+.o-overview__header {
+  margin-bottom: ms.step(1);
+}
+
+.o-overview__actions {
+  margin-bottom: ms.step(3);
+}
+
+@supports (display: grid) {
```
tylersticka, line 20: Do you think this is necessary? We aren't optimizing for IE, and the content seems perfectly accessible without grid if the spacing's a little off in this section.

I'm open to it, I was just surprised to see these fallbacks and then the margin resets later on for a relatively simple layout object.

src/objects/overview/overview.scss
```diff
@@ -0,0 +1,56 @@
...
+  @media (min-width: breakpoint.$l) {
+    .o-overview {
+      @include fluid.column-gap(
+        breakpoint.$s,
+        breakpoint.$xl,
+        size.$spacing-gap-fluid-min,
+        size.$spacing-gap-fluid-max
+      );
```
tylersticka, line 28: I'm torn on this. `column-gap` does seem to be [well supported in Grid](https://caniuse.com/mdn-css_properties_column-gap_grid_context), though not _quite_ as well supported as `grid-column-gap`. (We added this mixin for use with the `columns` property.)

I think it's probably fine, just typing out my initial pause.

src/objects/overview/overview.scss
```diff
@@ -0,0 +1,56 @@
...
+
+      align-items: center;
+      display: grid;
+      grid-template-areas:
+        'header header actions'
+        'content content content';
+      grid-template-columns: repeat(3, minmax(max-content, 1fr));
```
tylersticka, line 35: Why `max-content`?

src/objects/overview/overview.scss
Paul-Hebert: I don't have strong feelings here. 

The only change for IE11 was the `@supports` statement itself. The margins and margin resets were to handle the layout changes between small screens and large screens. I attempted to simplify this an use `display: grid` with gaps on all screen sizes, but this didn't quite match the prototype design. (On small screens there are two different margin sizes, so it doesn't map to a consistent gap.)

All that said, I'm totally fine removing the `supports` statement if it's just extra cruft.

tylersticka: I didn't realize the margin resets were also a responsive concern. In that case this isn't much cruft at all.

src/objects/overview/overview.scss
Paul-Hebert: I had initially used `grid-column-gap` when building out this component but VS Code warned me it was deprecated and that I should use `column-gap`. The support for `column-gap` and `grid-column-gap` seems the same in Safari which is the main modern browser I'd worry about since updates are tied to OS updates.

That said, I don't have strong feelings about it and would be fine swapping over.

src/objects/overview/overview.scss
Paul-Hebert: Good question. I'd copied that over from the prototype without thinking about it. This can be simplified.

Paul-Hebert: I simplified this to `grid-template-columns: repeat(3, 1fr);`. Let me know if you think a different value would make more sense.

tylersticka (approved):

PR #1415, opened 2021-07-19
Add Logo Group Object

### LINKED ISSUE #1392: Client logo cloud pattern
For the [case study listings](https://v-next--cloudfour-patterns.netlify.app/iframe.html?id=prototypes-case-study-listings--example&args=&viewMode=story) page, there is a small "cloud" of client logos. Right now it's being styled with flexbox so they can wrap and adjust.

Screen Shot 2021-07-12 at 9 21 53 AM

Small screens:
Screen Shot 2021-07-12 at 10 49 03 AM

dromo77:
## Overview

This PR adds a new object called Logo Group for displaying a group of logos together.

## Screenshots

Screen Shot 2021-07-19 at 12 37 11 PM

## Testing

1. [Deploy Preview](https://deploy-preview-1415--cloudfour-patterns.netlify.app/?path=/story/objects-logo-group--default-story)

---

- Fixes #1392


src/tokens/size/sizing.js
```diff
@@ -29,6 +29,12 @@ module.exports = {
           comment: 'The minimum space to display a default card in a grid.',
         },
       },
+      logo_group: {
+        item: {
```
tylersticka, line 33: Should this be `item_width` so that it's clear it represents a width without reading the comment?

src/objects/logo-group/logo-group.scss
```diff
@@ -0,0 +1,16 @@
+@use "../../compiled/tokens/scss/size";
+@use '../../mixins/ms';
+
+.o-logo-group {
+  align-items: center;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  margin: 0 auto;
+  max-width: size.$l;
```
tylersticka, line 10: This should probably be handled by the `o-container` object wrapping the logo group instead of the logo group itself?

src/objects/logo-group/logo-group.scss
```diff
@@ -0,0 +1,16 @@
...
+}
+
+.o-logo-group > * {
+  margin: ms.step(3);
```
tylersticka, line 14: Is there a way to apply margin gaps that won't result in extra margin/padding around the perimeter of the element?

src/objects/logo-group/logo-group.twig
```diff
@@ -0,0 +1,7 @@
+<div class="o-logo-group">
+  {% block content %}
+    {% for item in items %}
+      <img src="{{ item.src }}" alt="{{ item.alt|default('') }}">
```
tylersticka, line 4: I don't think `|default('')` is necessary?

src/objects/logo-group/logo-group.twig
```diff
@@ -0,0 +1,7 @@
+<div class="o-logo-group">
+  {% block content %}
+    {% for item in items %}
```
tylersticka, line 3: Should this be in a demo? I'm not sure we know enough about the intended data structure to bake in support for an array into the template itself.

src/objects/logo-group/demo/logos.json
```diff
@@ -0,0 +1,30 @@
...
+  {
+    "alt": "Walmart",
+    "src": "media/client-logos/logo-walmart.svg"
+  },
+  {
+    "alt": "Deschutes Brewery",
+    "src": "media/client-logos/logo-deschutes.svg"
+  },
+  {
+    "alt": "Obama",
```
Paul-Hebert, line 11: Nit: Should this be `Obama '08` to provide additional context?

src/objects/logo-group/logo-group.scss
dromo77: `o-container` has a max-width of 80em. In the prototype, the containing element for the logos has a max-width of 60em. Should I create an `o-container` modifier with `max-width: 60em`?

tylersticka: The first question I'd ask is whether or not that difference is meaningful. Does the pattern not work well with one of the existing container sizes? And is that something that could be solved by adjusting the size of the logos?

If the size difference is meaningful, then we should create a modifier with a name derived from what that difference is.

src/objects/logo-group/logo-group.stories.mdx
```diff
@@ -0,0 +1,13 @@
+import { Story, Canvas, Meta } from '@storybook/addon-docs/blocks';
+import defaultDemo from './demo/demo.twig';
```
gerardo-rodriguez, line 2: There's an opportunity to refine the story code example by using the Webpack `!!raw-loader` syntax here as well. 

Also see [the next suggestion](). 😉 
```suggestion
// The '!!raw-loader!' syntax is a non-standard, Webpack-specific, syntax.
// See: https://github.com/webpack-contrib/raw-loader#examples
// For now, it seems likely Storybook is pretty tied to Webpack, therefore, we are
// okay with the following Webpack-specific raw loader syntax. It's better to leave
// the ESLint rule enabled globally, and only thoughtfully disable as needed (e.g.
// within a Storybook docs page and not within an actual component).
// This can be revisited in the future if Storybook no longer relies on Webpack.
// eslint-disable-next-line @cloudfour/import/no-webpack-loader-syntax
import defaultDemoSource from '!!raw-loader!./demo/demo.twig';
import defaultDemo from './demo/demo.twig';
```

src/objects/logo-group/logo-group.stories.mdx
```diff
@@ -0,0 +1,13 @@
...
+import logos from './demo/logos.json';
+
+<Meta title="Objects/Logo Group" />
+
+# Logo Group
+
+Logo Group can be used to display a group of client logos. Logos break onto multiple lines depending on the available space and are vertically centered.
+
+<Canvas>
+  <Story name="Default">{defaultDemo({ items: logos })}</Story>
```
gerardo-rodriguez, line 12: See the [previous suggestion](). 🙂 

```suggestion
  <Story
    name="Default"
    parameters={{ docs: { source: { code: defaultDemoSource } } }}
  >
    {defaultDemo({ items: logos })}
  </Story>
```

dromo77: Thanks @gerardo-rodriguez!

src/objects/logo-group/logo-group.scss
```diff
@@ -0,0 +1,15 @@
+@use "../../compiled/tokens/scss/size";
+@use '../../mixins/ms';
+
+.o-logo-group {
+  align-items: center;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  margin: (-1 * ms.step(3));
```
tylersticka, line 9: This addresses my earlier feedback, but I notice now that this component needs a lot of padding to avoid horizontal scrollbars. Maybe it makes sense to add the padding back to the parent element (instead of requiring utilities), we should just make sure to include adequate comments explaining why... for example, "we add generous padding to balance out the whitespace differentiating brands" or something to that effect.

tylersticka (commented):
Looks good, second-guessing an earlier piece of feedback I gave (see comment inline)

src/objects/logo-group/logo-group.scss
dromo77: I removed the negative margin, curious what you think of it now?

tylersticka: @dromo77 Did you push that change? I still see the negative margin.

dromo77: No 😆  should be there now.

src/objects/logo-group/logo-group.scss
```diff
@@ -0,0 +1,20 @@
+@use "../../compiled/tokens/scss/size";
+@use '../../mixins/ms';
+
+/**
+ * Because there is generous spacing between the logos, we added a lot of white
+ * space around Logo Group to help balance the pattern visually. Including the
+ * spacing as part of the pattern means we don't need to rely on utilities.
+ */
```
tylersticka, line 8: This comment feels a little odd since there isn't any padding around the logo group. Maybe it should be rewritten and put closer to the items themselves (which have margin)?

src/objects/logo-group/logo-group.scss
```diff
@@ -0,0 +1,20 @@
...
+
+.o-logo-group {
+  align-items: center;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+}
+
+.o-logo-group > * {
+  margin: ms.step(3);
```
tylersticka, line 18: Should this be a design token? Maybe `size.spacing.gap.logo_group`?

tylersticka (approved):

PR #1422, opened 2021-07-20
Only run CI on PRs and pushes to v-next/main

### LINKED ISSUE #1421: Only run CI on PRs and pushes to `v-next`/`main`
Right now we run our CI whenever anyone pushes a commit to any branch on GitHub.

I will push feature branches to GitHub regularly as a way to back up and share my work. Every time I do that our CI runs. There are a couple problems with that:

- If I have lint/test/type errors I get an email that clogs up my email inbox and doesn't provide me any value. (I generally wait to resolve things like linting before creating a PR)
- Each time a computer somewhere has to spend about a minute and a half installing dependencies, building the site, compiling assets, linting, testing, type checking etc. This seems wasteful from an energy use perspective. Sure, it's a tiny drop in the bucket, but it also doesn't provide much value.

Instead we should limit our CI to running in the following situations where it does provide value:

- When PRs are created or updated
- When someone pushed to a production branch (`main` or `v-next`)

calebeby:
Closes #1421, see #1421 for reasons

[here's the docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,11 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+    - main
+    - v-next
 
 jobs:
   lint:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,8 +4,8 @@ on:
   pull_request:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


gerardo-rodriguez (approved):
:shipit:

PR #1423, opened 2021-07-20
Integrate reply Comment Form into Comment pattern

### LINKED ISSUE #1389: Comment form reply behavior
Our Comment component contains a placeholder for replies, but it does not yet integrate our Comment Form.

The experience should behave similarly to how it does on the live site:

reply-behavior

Note the focus of the first field and the reply form popping up at the top rather than scrolling below the other replies.

Paul-Hebert:
## Overview

The primary focus of this PR is integrating the Comment Form for replies into the Comment pattern.

While I was working on this I made some changes to the Comment Form component and updated the code examples for all the Comment stories.

I could use both design + dev review on this.

Follow-up issues will handle:

- finalizing template properties
- handling form submission/validation

## Screenshots

reply-pattern

## Testing

1. Review the [Comment docs page](https://deploy-preview-1423--cloudfour-patterns.netlify.app/?path=/docs/components-comment--single) for regressions
2. Review the [Comment Form docs page](https://deploy-preview-1423--cloudfour-patterns.netlify.app/?path=/docs/components-comment-form--default-story) for regressions
3. Review the [Reply experience](https://deploy-preview-1423--cloudfour-patterns.netlify.app/iframe.html?id=components-comment--with-reply-button&args=). (Also try toggling the `isLoggedIn` and `allowReplies` controls
4. Review [a more complex Reply experience](https://deploy-preview-1423--cloudfour-patterns.netlify.app/iframe.html?id=components-comment--with-reply-thread&args=allowReplies:true&viewMode=story) using VoiceOver and make sure the screen reader experience is clear and intuitive

---

- Fixes #1389


src/components/comment-form/comment-form.stories.mdx
```diff
@@ -92,6 +96,10 @@ There is also a reply version of this component (`is_reply: true`), used to repl
 ## Template Properties
 
 - `class`: Append a class to the root element.
+- `heading_id`: A unique ID for your heading element. This will be used as the accessible name for the form.
+- `heading_tag`: The tag for your heading (defaults to `h2`).
+- `heading_text`: The text for your heading.
+- `heading_class`: An optional class for your heading.
```
Paul-Hebert, line 102: I was tempted to replace this all with a heading block,  but that makes it more likely for the `aria-labelledby` relationship between the form and heading to get messed up.

src/components/comment-form/comment-form.twig
```diff
@@ -1,11 +1,18 @@
 <form
   class="c-comment-form{% if class %} {{class}}{% endif %}"
+  {% if heading_id %}aria-labelledby="{{ heading_id }}"{% endif %}
```
Paul-Hebert, line 3: This was added because I found out forms don't get a role of `form` without an accessible name:

https://github.com/testing-library/dom-testing-library/issues/937

src/components/comment-form/comment-form.twig
```diff
@@ -34,10 +41,10 @@
...
         {% include '@cloudfour/components/button/button.twig' with { label: 'Submit Reply' } only %}
-        {% include '@cloudfour/components/button/button.twig' with { label: 'Cancel', class: 'c-button--secondary' } only %}
+        {% include '@cloudfour/components/button/button.twig' with { label: 'Cancel', class: 'c-button--secondary js-cancel-reply' } only %}
       {% endblock %}
     {% endembed %}
   {% else %}
     {% include '@cloudfour/components/button/button.twig' with { label: 'Submit Comment' } only %}
   {% endif %}
-</div>
+</form>
```
Paul-Hebert, line 50: This was using the wrong closing tag

src/components/comment/comment.ts
```diff
@@ -0,0 +1,41 @@
...
+      ) as HTMLElement | null;
+
+      // Without this Timeout, VoiceOver does not properly focus the first input
+      // (though it works outside of VoiceOver).
+      // I went back and forth on whether or not to force this behavior in
+      // VoiceOver, since it will skip them past context around the comment
+      // (e.g. you may use simple HTML or markdown).
+      // I ended up adding this timeout to force the focus so that we provide
+      // an equivalent experience for all users.
+      setTimeout(() => firstInput?.focus(), 0);
```
Paul-Hebert, line 32: I'm conflicted on this. Thoughts?

src/components/comment/comment.scss
```diff
@@ -23,6 +23,7 @@
 
 /// Display a vertical line if the comment contains a reply thread to connect
 /// those comments visually with their parent.
+.c-comment.is-replying::after,
 .c-comment--thread::after {
```
Paul-Hebert, line 27: I added a separate class that does the same thing as `--thread` so our JS doesn't have to handle correctly toggling `--thread` and account for whether a comment already has a thread or not.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "esnext",
```
Paul-Hebert, line 3: This was changed to fix Typescript complaining about this syntax: `[...domNodeList]`

More details in slack: https://cloudfour.slack.com/archives/C03ECJ38E/p1626716597196500

src/components/comment/comment.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      name: /cancel/i,
+    });
+    await user.click(cancelButton);
+
+    // Back to our initial state
+    await expect(form).not.toBeVisible();
+    await expect(replyButton).toBeVisible();
+    await expect(comment).not.toHaveClass('is-replying');
+  })
+);
```
Paul-Hebert, line 72: @calebeby I wrote my first Pleasantest test for this PR! Let me know if you've got any feedback.

src/components/comment/comment.ts
gerardo-rodriguez: @Paul-Hebert Can you elaborate? You are conflicted on using a `setTimeout()`? Or something else?

gerardo-rodriguez:
> 4\. Review [a more complex Reply experience](https://deploy-preview-1423--cloudfour-patterns.netlify.app/iframe.html?id=components-comment--with-reply-thread&args=allowReplies:true&viewMode=story) using VoiceOver and make sure the screen reader experience is clear and intuitive

@Paul-Hebert Something is causing an error. 🤔 

Screen Shot 2021-07-20 at 4 05 04 PM

src/components/comment/comment.test.ts
```diff
@@ -0,0 +1,72 @@
...
+            name: 'Test',
+          },
+          comment_content: 'Test',
+          approved: true,
+        },
+        allow_replies: true,
+      })
+    );
+
+    await utils.loadCSS('../../../dist/standalone.css');
```
calebeby, line 33: Would it make sense to extract this into a function in the test-utils file? The same thing is happening in the sky nav tests and the elastic textarea tests

src/components/comment/comment.ts
```diff
@@ -0,0 +1,41 @@
+export const initComments = () => {
+  const commentsWithRepliesEnabled = document.querySelectorAll(
+    '.js-comment-with-replies-enabled'
+  );
+
+  for (const comment of [...commentsWithRepliesEnabled]) {
```
calebeby, line 6: No need to spread here! For loops work on iterators, not just arrays (just like how spread works on iterators, not just arrays)

Paul-Hebert:
>@Paul-Hebert Something is causing an error. 🤔

Oh no! 😱 

I'll take a look, thanks!

src/components/comment/comment.ts
```diff
@@ -0,0 +1,41 @@
...
+    // - Having the Comment Form have a block that you could pass a button into
+    const cancelButton = comment.querySelector('.js-cancel-reply');
+
+    replyButton?.addEventListener('click', () => {
+      comment.classList.add('is-replying');
+      replyButton.setAttribute('hidden', '');
+      replyForm?.removeAttribute('hidden');
+      const firstInput = replyForm?.querySelector(
+        'textarea, input'
+      ) as HTMLElement | null;
```
calebeby, line 23: Since we are in a .ts file we can use a type parameter. querySelector accepts a type parameter to tell it what types of element(s) it may find: https://www.typescriptlang.org/play?#code/MYewdgzgLgBATgUwA4BsCeAxEcC2MC8MA9AFQwCWA5mNgjDnSUTAAwwCGEHYaHXAEgBUAsgBkAoigQMwUAFBzQkWADNycaAEkwSAK6xCiVJmw4A-ADoAjroRw0AZQRTgUbAB4hYwQgAeUAEFEdklpBFkYAB8YL1FtPShQmSgAPgAKAHIoPyh2YIAaCh19DIBKIA

It'll automatically add the `| null` when you are using the type parameter

src/components/comment/comment.ts
Paul-Hebert: Sure! 

The initial issue (#1389) specifically mentioned the first input should gain focus when the form is shown. I think this is a great UX for most people since it allows them to skip a click and start typing right away.

To get this working in VoiceOver I had to add the timeout. As I was doing this, I started to question whether this was a good experience for screen reader users or not. If they're not able to visually see the page, the text in between the reply button and the first textarea will not be read:

Screen Shot 2021-07-20 at 4 12 37 PM

It seemed weird to not have this feature work the same for screen reader users and folks not using screen readers, but it also seemed weird to add in a `timeout` hack to enable an experience for screen reader users which is arguably worse.

Does that context help?

src/components/comment/comment.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    await expect(form).toBeVisible();
+    // Reply button is hidden
+    await expect(replyButton).not.toBeVisible();
+    await expect(comment).toHaveClass('is-replying');
+    // The first input or textarea should be focused.
+    // (In practice this will always be a textarea, but if the form is
+    // changed to lead with an input in the future, and that input is selected,
+    // this test should still pass.)
+    const inputs = await screen.getAllByRole(/textbox|input/);
+    await expect(inputs[0]).toHaveFocus();
```
calebeby, line 59: Great idea thinking ahead!

src/components/comment/comment.ts
```diff
@@ -0,0 +1,41 @@
+export const initCommentsWithRepliesEnabled = () => {
```
Paul-Hebert, line 1: This function is a mouthful but I wanted to be clear what it did. I'm open to other naming ideas.

src/components/comment/comment.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    const comment = await screen.getByRole('article');
+    const form = await screen.getByRole('form', { hidden: true });
+    const replyButton = await screen.getByRole('button', {
+      name: /reply/i,
+    });
+
+    // Initial state: reply form is hidden
+    await expect(form).not.toBeVisible();
+    await expect(comment).not.toHaveClass('is-replying');
```
calebeby, line 45: In general, I try to avoid using `.toHaveClass`. When I think I need it, I try to ask myself, "how would someone using a screen reader tell if it was in this state", and then I try to use matchers based on the answer to that, rather than `.toHaveClass`.

In this case, I was toggling `.is-replying` in the UI to try to figure out what it does, but I couldn't figure it out. Is it a part of the "public API" of the component (i.e. consumers are supposed to hook into this class)?

Paul-Hebert:
@gerardo-rodriguez should be fixed now. I got cocky with the global find-and-replace.

calebeby (changes requested):
@Paul-Hebert This looks great! I left some comments

src/components/comment/comment.ts
Paul-Hebert: Oh gotcha! Is it just `forEach` that you need to spread for?

src/components/comment/comment.test.ts
Paul-Hebert: Yeah, that's a good idea! I'll do that/

src/components/comment/comment.test.ts
Paul-Hebert: Good point. `.is-replying` adds the "comment thread" styles (the gray border along the left-hand side which is shown if a comment has replies, or the reply form is open)

From a user perspective I would expect to see a "border" (which is actually a pseudo-element).

Screen Shot 2021-07-20 at 4 39 51 PM

I'm not sure of another way to test that 🤔.

I guess I could also have `is-replying` control the visibility of the comment reply form instead of toggling `hidden`. This would simplify my JS and make this easier to test! I'll make that change.

src/components/comment/comment.ts
Paul-Hebert: Oh neat, thanks! I'll try that out.

src/components/comment/comment.ts
calebeby: Well you can only use .forEach on objects that have a forEach method on their prototype, (not all iterators, but most). Arrays, NodeLists, Sets, and Maps all have forEach methods, but that is because they each implement that, not because they are all iterators.

Maybe you are thinking of `.map` or other Array methods, which most other iterators don't have implemented.

src/components/comment/comment.test.ts
calebeby: Good idea! Is there any accessibility difference between `[hidden]` and `display: none`?

src/components/comment/comment.ts
gerardo-rodriguez: Thanks for the extra context, @Paul-Hebert! 

> As I was doing this, I started to question whether this was a good experience for screen reader users or not. If they're not able to visually see the page, the text in between the reply button and the first textarea will not be read

Good point. I'm wondering if this is a case where we might consider setting the focus on the heading (would require adding a `z-index`)? Or maybe wrapping the heading + instructions in an ARIA Live region and still have the focus jump to the first input? 🤔

src/components/comment/comment.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    // Click the cancel button to get back to our initial state
+    const cancelButton = await screen.getByRole('button', {
+      name: /cancel/i,
+    });
+    await user.click(cancelButton);
+
+    // Back to our initial state
+    await expect(form).not.toBeVisible();
+    await expect(replyButton).toBeVisible();
+    await expect(comment).not.toHaveClass('is-replying');
```
gerardo-rodriguez, line 70: When I was testing this manually, I was expecting to have the focus jump back to the "Reply" button after pressing the "Cancel" button. I felt as though I lost where I was at otherwise. What do you think, @Paul-Hebert? 🤔

Paul-Hebert: Sure, that makes sense :+1: I'll make that change

src/components/comment/comment.test.ts
Paul-Hebert: The main difference is [`hidden` will work if stylesheets aren't loaded](https://cloudfour.com/thinks/see-no-evil-hidden-content-and-accessibility/#hiding-content-for-everyone). I prefer using `hidden` in general, but I think in this case it would be fine to switch to `display: none;`

src/components/comment/comment.ts
Paul-Hebert: Got it. I think I was getting tripped up because [IE11 doesn't support `nodeList.forEach`](https://caniuse.com/?search=nodelist%20foreach) but that shouldn't be a concern here :+1:

src/components/comment/comment.test.ts
```diff
@@ -0,0 +1,72 @@
+import path from 'path';
+import type { PleasantestUtils } from 'pleasantest';
```
gerardo-rodriguez, line 2: TIL `import type` 👍

src/components/comment/comment.ts
Paul-Hebert: Yeah, good question 🤔 

Those both feel a little weird to me but I can't put my finger on why 🤔 . That said, the current experience also feels a little weird in a screen reader

I can try them out and see how they feel. :+1:

gerardo-rodriguez: > Those both feel a little weird to me but I can't put my finger on why 🤔 . That said, the current experience also feels a little weird in a screen reader

@Paul-Hebert I hear ya'. This is a very interesting UX to figure out. If you wanna jump on a call to share what each experience feels like, I'm available. 😉

Paul-Hebert: Yeah that sounds good. I'm going to respond to the other pieces of feedback first and then circle back to this.

Paul-Hebert: Depending on timing, maybe we could walk through this as a group in the design team meeting?

src/components/comment/comment.test.ts
calebeby: It is relatively new! It makes things easier for tools that are not type-aware (like babel) to know which imports need to be kept in the output, vs which imports are only used for types so they can't be in the output

src/components/comment/comment.ts
```diff
@@ -0,0 +1,41 @@
...
+      // an equivalent experience for all users.
+      setTimeout(() => firstInput?.focus(), 0);
+    });
+
+    cancelButton?.addEventListener('click', () => {
+      comment.classList.remove('is-replying');
+      replyButton?.removeAttribute('hidden');
+      replyForm?.setAttribute('hidden', '');
+    });
+  }
```
calebeby, line 38: @Paul-Hebert Should we return a cleanup function to remove the event listeners, like we've done with the other JS files? https://github.com/cloudfour/cloudfour.com-patterns/blob/v-next/src/components/sky-nav/sky-nav.ts#L111-L119

tylersticka (commented):
The reply auto-focus does not seem to work on my phone (iPhone X whatever the most recent stable version of iOS is, Safari, viewed on its own rather than in canvas view). The field appears to have focus but isn't actually focused.

IMG_4327

If focus is forced, the heading scrolls out of view. Could this be simplified so that the "message" label says "reply" or something like that to avoid the need for an additional visual heading?

IMG_4328

When typing in the message field, the height does not expand. Our text areas have an option for enabling this.

IMG_4329

src/components/comment/comment.ts
Paul-Hebert: This is a good question. I just went down a rabbit hole of SO answers and I'm still not sure:

- https://stackoverflow.com/questions/12528049/if-a-dom-element-is-removed-are-its-listeners-also-removed-from-memory
- https://stackoverflow.com/questions/18986520/does-removing-an-element-also-remove-its-event-listeners
- https://stackoverflow.com/questions/53256662/react-why-should-i-remove-event-listeners

It seems like it would be safest to return a cleanup function. I'll add one.

Paul-Hebert:
Thanks @tylersticka !

I'll look into the focus bug and enable the expanding textareas. 

Does this match the visual design you're describing?

Screen Shot 2021-07-21 at 10 00 02 AM

Do you think the "Please be kind..." text needs additional context or some sort of visual separator?

src/components/comment/comment.ts
gerardo-rodriguez: I see no harm in providing an API to remove event listeners. If the component consumer doesn't need it, no harm is done. If it is needed, though, it will be appreciated.

calebeby: @Paul-Hebert Once you add it we'll also want to use the cleanup function within storybook, as a cleanup function for the useEffect hook. I'm not sure how storybook's useEffect compares to React's, but React's useEffect hook reruns after any prop change, so if you change props, right now it probably keeps reattaching more and more event listeners

tylersticka:
> Do you think the "Please be kind..." text needs additional context or some sort of visual separator?

It might! Would it look okay with our Alert component?

src/components/comment/comment.ts
Paul-Hebert: Oh, wait! What if I used `aria-describedby` on the help text. Then it might be read aloud... 🤔

Paul-Hebert: Yep. `aria-describedby` is the way to go :+1: 

I'll share more info in the design meeting.

Paul-Hebert:
@tylersticka here's what it looks like using an Alert. I think it works alright:

Screen Shot 2021-07-21 at 1 41 16 PM

Another possible option would be including a border divider, though this may make the form feel _too_ disconnected from the comment above

Screen Shot 2021-07-21 at 1 49 04 PM

Maybe the best solution would be to just increase the margin above the help text a bit? Right now it matches the spacing of the containing rhythm object, but maybe it should match the grid gap of the comment form instead?

Screen Shot 2021-07-21 at 1 51 44 PM

Let me know what you think. Thanks

src/components/comment/comment.ts
gerardo-rodriguez: Yay, great solution, @Paul-Hebert! 🎉

src/components/comment/comment.ts
gerardo-rodriguez: I think this is fine. ¯\\\_(ツ)_/¯

tylersticka:
@Paul-Hebert I prefer the version with the message enclosed to the border, but there's probably still room for improvement.

Have you considered making the reply form closer in appearance to a comment, maybe using the icon as a placeholder for the avatar?

Screen Shot 2021-07-21 at 2 21 11 PM

Also, it would be completely fair to get this working well from a voiceover and technical perspective and to create a separate issue for better differentiating the comment form from its thread. I don't want to bog down the heart of this PR with stuff we could iterate on separately.

Paul-Hebert:
That makes sense @tylersticka ! I like the idea of mirroring the comment layout. I'm going to make a follow-up issue to keep this ball rolling.

Paul-Hebert:
Alright, I believe I've addressed all of the technical feedback! Further design improvements will take place in a follow-up PR. Re-review appreciated!

Making cards for follow-up tasks now. Thanks!

src/components/comment/comment.stories.mdx
```diff
@@ -28,22 +61,100 @@ At minimum, a single comment consists of:
...
 
 <Canvas>
-  <Story name="Single">{template({ comment: makeComment() })}</Story>
+  <Story
+    name="Single"
+    parameters={{
+      docs: {
+        source: {
+          code: makeTwigInclude(
+            '@cloudfour/components/comment-form/comment-form.twig',
```
gerardo-rodriguez, line 70: Does this need to be updated? I'm not sure. 🤔 
```suggestion
            '@cloudfour/components/comment/comment.twig',
```

src/components/comment/comment.stories.mdx
```diff
@@ -28,22 +61,100 @@ At minimum, a single comment consists of:
...
 <Canvas>
-  <Story name="Role: Author">
-    {template({ comment: makeComment(), demo_post_author: true })}
+  <Story
+    name="Role: Author"
+    parameters={{
+      docs: {
+        source: {
+          code: makeTwigInclude(
+            '@cloudfour/components/comment-form/comment-form.twig',
```
gerardo-rodriguez, line 104: Same here? 🤔 
```suggestion
            '@cloudfour/components/comment/comment.twig',
```

src/components/comment/comment.stories.mdx
```diff
@@ -28,22 +61,100 @@ At minimum, a single comment consists of:
...
 <Canvas>
-  <Story name="Role: Cloud Four">
-    {template({ comment: makeComment(), demo_cloud_four_member: true })}
+  <Story
+    name="Role: Cloud Four"
+    parameters={{
+      docs: {
+        source: {
+          code: makeTwigInclude(
+            '@cloudfour/components/comment-form/comment-form.twig',
```
gerardo-rodriguez, line 136: And here? 🤔 
```suggestion
            '@cloudfour/components/comment/comment.twig',
```

src/components/comment/comment.stories.mdx
```diff
@@ -52,8 +163,36 @@ It is helpful for context within a discussion to know when a commentor is the or
...
 <Canvas>
-  <Story name="Unapproved">
-    {template({ comment: makeComment({ approved: false }) })}
+  <Story
+    name="Unapproved"
+    parameters={{
+      docs: {
+        source: {
+          code: makeTwigInclude(
+            '@cloudfour/components/comment-form/comment-form.twig',
```
gerardo-rodriguez, line 172: Here too? 🤔 
```suggestion
            '@cloudfour/components/comment/comment.twig',
```

src/components/comment/comment.stories.mdx
```diff
@@ -62,27 +201,82 @@ If `comment.approved` is not `true`, an [Alert](/docs/components-alert--basic) w
...
-      source: {
-        url: 'https://twitter.com/smashingmag/status/1371521325236416516',
-        name: 'twitter.com',
+  <Story
+    name="With source"
+    parameters={{
+      docs: {
+        source: {
+          code: makeTwigInclude(
+            '@cloudfour/components/comment-form/comment-form.twig',
```
gerardo-rodriguez, line 210:
```suggestion
            '@cloudfour/components/comment/comment.twig',
```

src/components/comment/comment.stories.mdx
```diff
@@ -62,27 +201,82 @@ If `comment.approved` is not `true`, an [Alert](/docs/components-alert--basic) w
...
-      demo_control: true,
-    })}
+  <Story
+    name="With reply button"
+    args={{ allowReplies: true, isLoggedIn: true }}
+    parameters={{
+      docs: {
+        source: {
+          code: makeTwigInclude(
+            '@cloudfour/components/comment-form/comment-form.twig',
```
gerardo-rodriguez, line 257:
```suggestion
            '@cloudfour/components/comment/comment.twig',
```

src/components/comment/comment.stories.mdx
```diff
@@ -96,7 +290,70 @@ While it is theoretically possible to infinitely nest `children`, it's recommend
...
 <Canvas>
-  <Story name="With reply thread">
-    {template({ comment: makeComment({ replies: 2 }) })}
+  <Story
+    name="With reply thread"
+    parameters={{
+      docs: {
+        source: {
+          code: makeTwigInclude(
+            '@cloudfour/components/comment-form/comment-form.twig',
```
gerardo-rodriguez, line 299:
```suggestion
            '@cloudfour/components/comment/comment.twig',
```

src/components/comment/comment.stories.mdx
```diff
@@ -96,7 +290,70 @@ While it is theoretically possible to infinitely nest `children`, it's recommend
...
+    name: string | () => string;
+    link: string | () => string;
+  }
+  ```
+- `log_out_url`: URL used for log out link.
+- `source`: An optional object containing a `url` and `name` for the comment source.
+
+## JavaScript Instructions
+
+You can run `initCommentReplyForm` to initialize a comment's reply form. Comments with reply forms will have the
```
gerardo-rodriguez, line 342: Will have the…? 🤔

src/components/comment-form/comment-form.stories.mdx
```diff
@@ -51,12 +55,13 @@ The form used to add a comment to an article.
         logged_in_user: isLoggedIn ? tyler : undefined,
         log_out_url: '/logout',
         is_reply: isReply,
+        heading_id: 'leave-a-comment',
```
gerardo-rodriguez, line 58: Is this needed here also? 🤔 
```suggestion
        heading_id: 'leave-a-comment',
        help_text_id: 'leave-a-comment-help-text',
```

src/components/comment/comment.ts
```diff
@@ -0,0 +1,49 @@
...
+  // we're directly manipulating a child component, but this is a lot simpler
+  // than other potential solutions. Other options included:
+  // - Having the Comment Form emit a `cancel` event
+  // - Having the Comment Form have a block that you could pass a button into
+  const cancelButton =
+    comment.querySelector<HTMLButtonElement>('.js-cancel-reply');
+
+  const show = () => {
+    comment.classList.add('is-replying');
+    replyButton?.setAttribute('hidden', '');
```
calebeby, line 19: @Paul-Hebert Instead of using optional chaining for accessing all these elements, could we check that they all exist near the top? Or does there need to be partial functionality if only some exist?

```js
const replyButton = ...
const replyForm = ...
const cancelButton = ...

if (!replyButton || !replyForm || !cancelButton) return { destroy: () => {} }
```

Paul-Hebert: Sure :+1:

src/components/comment-form/comment-form.stories.mdx
Paul-Hebert: Yes, thank you :+1:

src/components/comment/comment.stories.mdx
Paul-Hebert: Oh yeah, good catch :+1:

Paul-Hebert:
The latest round of feedback has been addressed. Re-review appreciated, thanks!

tylersticka (commented):
Works much better on small screens! Thanks, @Paul-Hebert! Deferring further design feedback to the separate issues created and deferring approval to @cloudfour/dev since they had more immediate feedback to resolve.

calebeby:
@Paul-Hebert It looks like [this comment](https://github.com/cloudfour/cloudfour.com-patterns/pull/1423#discussion_r674414493) is not resolved?

src/components/comment/comment.stories.mdx
Paul-Hebert:
```suggestion
You can run `initCommentReplyForm` to initialize a comment's reply form. Comments with reply forms will have the `js-comment-with-reply-form` class which you can use to query and initialize them.
```

Paul-Hebert:
>@Paul-Hebert It looks like this comment is not resolved?

Ack, thanks @calebeby , fixed!

calebeby (approved):
Code and tests look great! I didn't closely test it, but I assume that Tyler and Gerardo did

Paul-Hebert:
Oh no! Since we updated Eslint the lint rules are different in the CI task from what I've got locally. One moment while I resolve this...

calebeby (approved):

PR #1426, opened 2021-07-21
Refine Table component code examples

gerardo-rodriguez:
## Overview

This PR refines the Table component code examples.

---

**Note:** I had a hard time figuring out what to do for this component. See the following options. Do you all agree/disagree with my choice?

### Option 1: Use raw source code
This is what I implemented but the source code has its own custom template logic which makes the code example slightly confusing.

### Option 2: Use `makeTwigInclude` 

This option did not apply because the Table component expects `block` content.

### Option 3: Manually write custom (simplified) `code` example in the `docs.source.code` value of the `parameters` attribute

We've been avoiding duplicating the source code so I did not consider this option a viable option.

## Testing

1. instructions for reviewers
1. to test your changes

---

- Fixes # (issue)

/CC @person

### DIFF
--- a/src/components/table/table.stories.mdx
+++ b/src/components/table/table.stories.mdx
@@ -1,4 +1,13 @@
 import { Story, Canvas, Meta } from '@storybook/addon-docs/blocks';
+// The '!!raw-loader!' syntax is a non-standard, Webpack-specific, syntax.
+// See: https://github.com/webpack-contrib/raw-loader#examples
+// For now, it seems likely Storybook is pretty tied to Webpack, therefore, we are
+// okay with the following Webpack-specific raw loader syntax. It's better to leave
+// the ESLint rule enabled globally, and only thoughtfully disable as needed (e.g.
+// within a Storybook docs page and not within an actual component).
+// This can be revisited in the future if Storybook no longer relies on Webpack.
+// eslint-disable-next-line @cloudfour/import/no-webpack-loader-syntax
+import basicDemoSource from '!!raw-loader!./demo/basic.twig';
 import basicDemo from './demo/basic.twig';
 import data from './demo/data.json';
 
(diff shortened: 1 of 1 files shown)


Paul-Hebert (dismissed):
I'm fine with this approach. This is definitely a step up from the current experience, and gives you enough info to know how to use the component. Alternately I'd also be fine manually adding a simplified example:

```twig
{% embed '@cloudfour/components/table/table.twig' with { table_class: _table_class } %}
  {% block caption %}
    <!-- Optionally add a caption string -->
  {% endblock %}

  {% block header %}
    <!-- Optionally add a single `<tr>` element -->
  {% endblock %}

  {% block body %}
    <!-- Add `<tr>` elements -->
  {% endblock %}

  {% block footer %}
    <!-- Optionally add a single `<tr>` element -->
  {% endblock %}
{% endembed %}
```

You could store this as a variable you pass into the various stories. If you wanted, you could make it a function so you could pass in a different `table_class` for each story.

gerardo-rodriguez:
> I'm fine with this approach. This is definitely a step up from the current experience, and gives you enough info to know how to use the component. Alternately I'd also be fine manually adding a simplified example:
> 
> ```
> {% embed '@cloudfour/components/table/table.twig' with { table_class: _table_class } %}
>   {% block caption %}
>     
>   {% endblock %}
> 
>   {% block header %}
>     
>   {% endblock %}
> 
>   {% block body %}
>     
>   {% endblock %}
> 
>   {% block footer %}
>     
>   {% endblock %}
> {% endembed %}
> ```
> 
> You could store this as a variable you pass into the various stories. If you wanted, you could make it a function so you could pass in a different `table_class` for each story.

Excellent idea, @Paul-Hebert! 🎉

src/components/table/table.stories.mdx
```diff
@@ -1,6 +1,26 @@
 import { Story, Canvas, Meta } from '@storybook/addon-docs/blocks';
 import basicDemo from './demo/basic.twig';
 import data from './demo/data.json';
+const basicTableCodeExample = (tableClass) => {
+  let withArgs = '';
+  if (tableClass) {
+    withArgs = ` with { table_class: '${tableClass}' }`;
+  }
```
calebeby, line 8: Nit: would it be clearer/more concise with a ternary? If you don't think so that's totally fine, I'm not totally sure which I like better.

calebeby (approved):
Left one comment, LGTM

calebeby (approved):

PR #1434, opened 2021-07-22
Refine the visual display of the comment reply form help text

### LINKED ISSUE #1427: Refine design of comment reply form
We reused the Comment Form pattern to reply to comments. The heading felt bulky and distracting so we hid it visually, but now it's not visually clear that the form help text is connected to the form. Here are a couple PR comments describing potential solutions:

- https://github.com/cloudfour/cloudfour.com-patterns/pull/1423#issuecomment-884509325
- https://github.com/cloudfour/cloudfour.com-patterns/pull/1423#issuecomment-884492052

We should explore and implement a design solution for this.

This ticket is marked as design and dev, but if someone wants to carve off the design portion and tackle that alone, that's totally fine.

Paul-Hebert:
## Overview

This PR wraps the help text in and alert, and adjust the spacing between the reply form and the comment above.

The goal is to more clearly distinguish the reply form from the comment, and make it clear that the help text is connected to the reply form, not the comment above.|

## Screenshots

Screen Shot 2021-07-22 at 12 48 33 PM

## Testing

1. Review the [preview deploy](https://deploy-preview-1434--cloudfour-patterns.netlify.app/?path=/story/components-comment--with-reply-button)

---

Fixes #1427

/CC @tylersticka

### DIFF
--- a/src/components/comment-form/comment-form.twig
+++ b/src/components/comment-form/comment-form.twig
@@ -1,13 +1,3 @@
-{% set help_text %}
-  <p {% if help_text_id %}id="{{ help_text_id }}"{% endif %}>
-      Please be kind, courteous and constructive.
-      You may use simple HTML or
-      <a href="https://en.support.wordpress.com/markdown-quick-reference">Markdown</a>
-      in your comments.
-      All fields are required.
-  </p>
-{% endset %}
-
 <form
   class="c-comment-form{% if class %} {{class}}{% endif %}"
   {% if heading_id %}aria-labelledby="{{ heading_id }}"{% endif %}
@@ -23,15 +13,17 @@
       {{ heading_text|default("Leave a Comment") }}
     </{{_heading_tag}}>
 
-    {% if wrap_help_text_in_alert %}
-      {% embed '@cloudfour/components/alert/alert.twig' %}
-        {% block content %}
-          {{help_text}}
-        {% endblock %}
-      {% endembed %}
-    {% else %}
-      {{help_text}}
-    {% endif %}
+    {% embed '@cloudfour/components/alert/alert.twig' %}
+      {% block content %}
+        <p {% if help_text_id %}id="{{ help_text_id }}"{% endif %}>
+          Please be kind, courteous and constructive.
+          You may use simple HTML or
+          <a href="https://en.support.wordpress.com/markdown-quick-reference">Markdown</a>
+          in your comments.
+          All fields are required.
+        </p>
+      {% endblock %}
+    {% endembed %}
   {% endblock %}
   {% embed '@cloudfour/objects/form-group/form-group.twig' with { label: main_label|default('Message') } %}
     {% block control %}
--- a/src/components/comment-form/comment-form.twig
+++ b/src/components/comment-form/comment-form.twig
@@ -1,3 +1,13 @@
+{% set help_text %}
+  <p {% if help_text_id %}id="{{ help_text_id }}"{% endif %}>
+      Please be kind, courteous and constructive.
+      You may use simple HTML or
+      <a href="https://en.support.wordpress.com/markdown-quick-reference">Markdown</a>
+      in your comments.
+      All fields are required.
+  </p>
+{% endset %}
+
 <form
   class="c-comment-form{% if class %} {{class}}{% endif %}"
   {% if heading_id %}aria-labelledby="{{ heading_id }}"{% endif %}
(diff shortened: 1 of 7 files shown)


src/components/comment-form/comment-form.twig
```diff
@@ -12,13 +22,16 @@
...
+
+    {% if wrap_help_text_in_alert %}
+      {% embed '@cloudfour/components/alert/alert.twig' %}
+        {% block content %}
+          {{help_text}}
+        {% endblock %}
+      {% endembed %}
+    {% else %}
+      {{help_text}}
+    {% endif %}
```
tylersticka, line 34: Would this be simpler if we used blocks instead of a property?

```suggestion
    {% block help_text %}
      {% if help_text is not empty %}
        {% embed '@cloudfour/components/alert/alert.twig' %}
          {% block content %}
            {{help_text}}
          {% endblock %}
        {% endembed %}
      {% endif %}
    {% endblock %}
```

src/components/comment/comment.scss
```diff
@@ -50,6 +51,10 @@
 
 .c-comment.is-replying .c-comment__reply-form {
   display: block;
+  /// The top margin matches the grid gap spacing applied to the comment form.
+  /// Otherwise, the form's help text appears visually closer to the comment
+  /// you're responding to than to the form.
+  margin-top: ms.step(2, 1rem);
```
tylersticka, line 57: Did you consider using what's used for the replies themselves to inherit any rhythm object spacing?

```
margin-top: var(--rhythm, #{size.$rhythm-default});
```

src/components/comment/comment.twig
```diff
@@ -125,6 +125,7 @@
           heading_text: "Reply to #{comment.author.name}",
           heading_class: 'u-hidden-visually',
           main_label: 'Reply',
+          wrap_help_text_in_alert: true,
```
tylersticka, line 128: See earlier comment about using a block instead of a boolean property.

src/components/comment/comment.scss
Paul-Hebert: This is what I was doing previously, but the rhythm spacing between replies is smaller than the spacing within the comment form. 

This led to the help text being closer to the comment above it than to the rest of the form. I worried that made it less clear what the help text was connected to.

Though maybe if the rhythm spacing used on the WP site is larger that won't be an issue?

src/components/comment-form/comment-form.twig
Paul-Hebert: I'm not sure. This component is used in two different places:

- Leaving a comment on a post
- Responding to a comment on a post

I think the following things are the same in both use cases:

- The help text content
- The `aria-describedby` link between the help text and the `textarea`

I think the following is different:

- Whether the help text should be wrapped in an alert.

If I'm understanding your suggestion correctly, we would need to pass in the help text content, and maintain the `aria-describedby` link manually when we didn't want the help text in an alert. That seems like more of a maintenance burden than adding a new property, but I may be misunderstanding the suggestion.

tylersticka: When do we not want the help text in an alert?

src/components/comment/comment.scss
tylersticka: In that case, can we use whatever token the form is using instead of a new number? (And if it's not using a token, maybe it should be?)

Paul-Hebert: Sure, I can switch the form to using a token and use it here as well.

src/components/comment-form/comment-form.twig
Paul-Hebert: Ahh, perhaps I had misunderstood. My impression was we didn't want to use the alert in the main "Leave a comment" form. 

image

If that's not the case then you're right that this new prop is over-complicating things. I can switch to using an alert everywhere.

tylersticka: Yeah, let's keep it consistent!

Paul-Hebert:
Thanks @tylersticka ! Based on your feedback I made a couple changes:

- The Comment Form always wraps the help text in an alert so we no longer need a new prop https://github.com/cloudfour/cloudfour.com-patterns/pull/1434/commits/bd3428c90acd19e7519c0ff70c3dcdf5451b9310
- The Comment uses tokens for its gaps. The vertical gap token is also used for the space between the reply form and the comment above it https://github.com/cloudfour/cloudfour.com-patterns/pull/1434/commits/2249cf9ee109ef2dfe6eb7c7c70bdd188a77f698

Let me know if you'd like any further changes. Thanks!

src/components/comment/comment.scss
```diff
@@ -4,6 +4,7 @@
 @use '../../compiled/tokens/scss/size';
 @use '../../mixins/headings';
 @use '../../mixins/theme';
+@use "../../mixins/ms";
```
tylersticka, line 7: Nit: Is this still necessary?

tylersticka (approved):

src/components/comment/comment.scss
Paul-Hebert: Nope :+1: Removed

Paul-Hebert:
@tylersticka removing the `ms` mixin dismissed your review. Mind re-approving?

tylersticka (approved):

PR #1439, opened 2021-07-23
Switch Comment Badging to a Slot

### LINKED ISSUE #1429: Finalize Comment Template Properties
The Comment pattern has a number of properties. 

The primary property is an object matching a [Timber Comment](https://timber.github.io/docs/reference/timber-comment/). In addition to that it exposes a few other properties:

- `allow_replies`: A boolean property that controls whether to show a reply button and form
- `logged_in_user`: user object
- `log_out_url`: URL used for log out link.
- `source`: An optional object containing a url and name for the comment source.
- `demo_post_author`: Whether the commenter is the author
- `demo_cloud_four_member`: Whether the commenter is a team member.

Right now there's no way to pass these properties down to child comments (replies).

I think most of these probably are fine to not pass down, but the last two (`demo_post_author`, `demo_cloud_four_member`) will need to be passed down. We should explore how we'll integrate these in Wordpress, rename the properties, and make any necessary changes to the patterns.

Paul-Hebert:
## Overview

This PR makes a change to how author/member badges can be applied to comments.

Instead of using properties for this, the template now accepts an `author_meta` slot.

This allows the consuming app to use whatever logic makes sense to determine what badge to apply.

## Screenshots

(No visual changes)

## Testing

1. Review the [preview deploy docs page](https://deploy-preview-1439--cloudfour-patterns.netlify.app/?path=/docs/components-comment--single)

---

- Fixes #1429

### DIFF
--- a/.changeset/short-walls-hear.md
+++ b/.changeset/short-walls-hear.md
@@ -2,4 +2,4 @@
 '@cloudfour/patterns': minor
 ---
 
-Switch comment badging to use a slot
+Increase comment template flexibility using blocks.
--- a/.changeset/short-walls-hear.md
+++ b/.changeset/short-walls-hear.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Switch comment badging to use a slot
--- a/src/components/comment/comment.stories.mdx
+++ b/src/components/comment/comment.stories.mdx
@@ -288,8 +288,6 @@ While it is theoretically possible to infinitely nest `children`, it's recommend
 
 ## Template Properties
 
-Note: These template properties are not finalized, and may change in the future.
-
 - `comment`: an object matching the structure of a [Timber comment](https://timber.github.io/docs/reference/timber-comment/)
 - `allow_replies`: A boolean property that controls whether to show a reply button and form
 - `logged_in_user`: [user object](https://timber.github.io/docs/reference/timber-user/#properties) of the type:
@@ -301,6 +299,13 @@ Note: These template properties are not finalized, and may change in the future.
   ```
 - `log_out_url`: URL used for log out link.
 - `source`: An optional object containing a `url` and `name` for the comment source.
+
+## Template Blocks
+
+- `header_content`: Use to over-write the content of the comment header
+- `heading_content`: Use to over-write the content of the comment heading
+- `author_title`: Use to over-write a portion of the comment heading content. Will be followed by a visually hidden span reading either `replied` or `said`. (Valid values could be `John Doe` or `John Doe (Post Author)`.)
+- `replies`: Use to pass in your own replies markup. This is useful if you need to modify other blocks within replies.
 
 ## JavaScript Instructions
 
--- a/src/components/comment/comment.stories.mdx
+++ b/src/components/comment/comment.stories.mdx
@@ -5,6 +5,18 @@ import { initCommentReplyForm } from './comment.ts';
 import { createElasticTextArea } from '../input/elastic-textarea.ts';
 import { useEffect } from '@storybook/client-api';
 import { makeTwigInclude } from '../../make-twig-include';
+import authorDemo from './demo/author.twig';
+import memberDemo from './demo/member.twig';
+// See: https://github.com/webpack-contrib/raw-loader#examples
+// For now, it seems likely Storybook is pretty tied to Webpack, therefore, we are
+// okay with the following Webpack-specific raw loader syntax. It's better to leave
+// the ESLint rule enabled globally, and only thoughtfully disable as needed (e.g.
+// within a Storybook docs page and not within an actual component).
+// This can be revisited in the future if Storybook no longer relies on Webpack.
+// eslint-disable-next-line @cloudfour/import/no-webpack-loader-syntax
+import authorDemoSource from '!!raw-loader!./demo/author.twig';
+// eslint-disable-next-line @cloudfour/import/no-webpack-loader-syntax
+import memberDemoSource from '!!raw-loader!./demo/member.twig';
 const tyler = {
   name: 'Tyler Sticka',
   link: 'https://cloudfour.com/is/tyler',
@@ -41,12 +53,6 @@ const initCommentReplyForms = () => {
 # Comment
 
 Displays a single comment in a comment thread, optionally with replies. Multiple comments can be displayed within [a Rhythm layout object](/docs/objects-rhythm--default-story).
-
-## Status
-
-This component is still a work in progress. The following features are still in development:
-
-- Finalizing this pattern's blocks and properties for theme integration.
 
 ## Single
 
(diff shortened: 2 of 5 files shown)


.changeset/short-walls-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
```
Paul-Hebert, line 2: This is technically major, but the breaking change affects a feature we're not using anywhere yet?

src/components/comment/comment.stories.mdx
```diff
@@ -85,27 +91,24 @@ This information may be passed to the component as a `comment` object matching t
   </Story>
 </Canvas>
 
-## Role badges
+## Author Meta Info
+
+You can use the `author_meta` block to display author meta information.
```
Paul-Hebert, line 96: @tylersticka I know there was some discussion of whether the entire heading content should be a block, or if just the meta info should be a block. I thought it made more sense to only affect the meta info, but I could easily be convinced otherwise.

tylersticka: @Paul-Hebert I think I'd like to convince you otherwise. 🙂 

The approach you've taken here means a developer can append content to the heading, and that's it. They can't modify the heading content, they can't prepend content, etc. without creating additional blocks. And because there's no `<div class="author-meta">` or anything wrapping it, we haven't saved them the trouble of rewriting markup. It's just kind of a very specific block that goes unused by default.

If you instead did the following:

```
{%block heading %}
  {{comment.author.name}}
{% endblock %}
```

Now we have a block with a very defined relationship to its containing element, and developers have the option of replacing it entirely, prepending content, or appending content, all without the need to create additional blocks.

I feel strongly that this would be a better approach.

Paul-Hebert: Sure, I've been convinced 🙂 

I can switch that over. (Currently looking into some biz-dev stuff. Will circle back before EOD)

Paul-Hebert: @tylersticka , following up on this. Here's the current heading structure (without the meta stuff)

```twig
<h{{_heading_depth}} class="c-comment__heading">
  {{comment.author.name}}

  <span class="u-hidden-visually">
    {% if comment.is_child %}replied{% else %}said{% endif %}:
  </span>
</h{{_heading_depth}}>
```

I think the block should be inside the `h{{_heading_depth}}` tag, so perhaps it should be called `heading_content`?

```twig
<h{{_heading_depth}} class="c-comment__heading">
  {% block heading_content %}
    {{comment.author.name}}
  {% endblock %}

  <span class="u-hidden-visually">
    {% if comment.is_child %}replied{% else %}said{% endif %}:
  </span>
</h{{_heading_depth}}>
```

But this block still isn't the full heading content, since there's the visually hidden text. The visually hidden text is following some special logic right now, and it doesn't seem like the component consumer should need to recreate that. But, whatever they put in our new block needs to work with the visually hidden text following it. Maybe the block should be called `author_title` or something along those lines?

```twig
<h{{_heading_depth}} class="c-comment__heading">
  {% block author_title %}
    {{comment.author.name}}
  {% endblock %}

  <span class="u-hidden-visually">
    {% if comment.is_child %}replied{% else %}said{% endif %}:
  </span>
</h{{_heading_depth}}>
```

I went back and forth between title and name, since the block kind of does both. What do you think?

tylersticka: I don't think it should be called `author_title` because that dictates what the content is. I think `heading_content` would be fine.

Paul-Hebert: Okay, do you think `heading_content` should also contain the visually hidden text? 

Or should we just be explicit in the documentation that whatever you put in `heading_content` should make sense followed by a visually hidden `replied`/`said`?

tylersticka: More documentation never hurt.

Something I try to do when naming stuff is imagine what will happen if I need more levels of depth.

To be clear, **I'm not suggesting you actually do this**, but if we imagine that at some point we might _need_ to modify more aspects of this, I could see that shaping up like...

```twig
{% block header %}
  <h3 class="c-comment__heading">
    {% block heading %}
      {% block heading_content %}
        {{comment.author.name}}
      {% endblock %}
      <span class="u-hidden-visually">
        {% if comment.is_child %}replied{% else %}said{% endif %}:
      </span>
    {% endblock %}
  </h3>
{% endblock %}
```

Which seems fine to me.

Paul-Hebert: Yeah that makes sense. 

Side note: My hesitation with some of those names is the blocks are named for what they're _inside_ not what they expect to be passed. For example if you pass a heading (`h1`-`h6`) into the `heading` block, you're not going to get what you expect. 

But, the underlying point stands that multiple levels of blocks can provide more flexibility.

---

As we chat about this more I'm realizing my current solution does not fully solve Wordpress integration, since there's no way to pass the `heading_content` block into child comments. So maybe we also need to make a `replies` block as you mentioned [in the issue](https://github.com/cloudfour/cloudfour.com-patterns/issues/1429)

My concern then becomes that a lot of the a11y stuff the pattern currently handles (heading levels, visually hidden text) becomes the responsibility of the component consumer. At that point they need to understand the inner workings of the component so they can know the correct heading levels to use and such. This may not be a big deal, and I may be overthinking it.

I'm going to take a step back and experiment to see what makes sense.

tylersticka: > Side note: My hesitation with some of those names is the blocks are named for what they're _inside_ not what they expect to be passed. For example if you pass a heading (`h1`-`h6`) into the `heading` block, you're not going to get what you expect.

I guess the way I think about this is that blocks maybe shouldn't be named based on their content if that content can change? So if the `heading` block _wraps_ the `<h3>`, and you pass in a `<table>`, now it's no longer a heading.

I dunno, I could be wrong.

Paul-Hebert: Yea, that's fair. Names are hard!

Paul-Hebert: @tylersticka how do you feel about these blocks?

```twig
<header class="c-comment__header">
  {% block header_content %}
    <h{{_heading_depth}} class="c-comment__heading">
      {% block heading_content %}
        {% block author_title %}
          {{comment.author.name}}
        {% endblock %}

        <span class="u-hidden-visually">
          {% if comment.is_child %}replied{% else %}said{% endif %}:
        </span>
      {% endblock %}
    </h{{_heading_depth}}>
  {% endblock %}
</header>
```

For our existing use case (passing in a badge) you would pass in an `author_title`. If you wanted to replace all of the heading content, you could do so with `heading_content`. If you want to replace the heading, you could pass in `header_content`.

(`header_content` may be premature?)

If this still feels off maybe we can find some time to chat with voices?

tylersticka: Sure, that's fine with me

tylersticka: (the suggested blocks, not the proposal to chat with voices... I mean, I'm fine chatting with voices, I just don't know that I need to for this!)

Paul-Hebert: Cool, thanks Tyler! Implementing this now

Paul-Hebert:
@tylersticka I adjusted the blocks based on our discussion, added a `replies` block,  and added documentation around these blocks. Let me know what you think. If possible I'd love to try integrating this into Wordpress before I leave for vacation to confirm it all works as expected.

tylersticka (commented):
It looks like maybe the documentation needs updating to match the new block names?

Paul-Hebert:
Good catch, done!

tylersticka (approved):

PR #1440, opened 2021-07-23
Fix issue with Comment `replies` block not working

Paul-Hebert:
## Overview

The Comment template has a `replies` block allowing devs to customize how replies are output.

This block was not working correctly because it was nested inside another embed.

This PR uses `set` to define the replies outside of the embed and then pass them in.

### DIFF
--- a/src/components/comment/comment.twig
+++ b/src/components/comment/comment.twig
@@ -113,18 +113,27 @@
       <h{{_section_heading_depth}} class="u-hidden-visually">
         Replies to {{comment.author.name}}
       </h{{_section_heading_depth}}>
+
+      {#
+        This is defined here, so we can pass it into the block of a nested embed.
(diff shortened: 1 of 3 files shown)


src/components/comment/comment.twig
```diff
@@ -113,18 +113,27 @@
...
+      {% set repliesContent %}
+        {% block replies %}
+          {% for child in comment.children %}
+            {% include '@cloudfour/components/comment/comment.twig' with {
+              comment: child,
+              heading_depth: _child_heading_depth,
+            } only %}
+          {% endfor %}
+        {% endblock %}
+      {% endset %}
```
Paul-Hebert, line 130: I'm open to other ideas if there's a less yucky way to handle this. This needs to be defined after `_child_heading_depth` is set

tylersticka: I think this is fine 👍

tylersticka (approved):

Issue #1418, opened 2021-07-19
Prototype team pages


tylersticka:
Our site's "team" section consists of two templates:

- The listing of team members: https://cloudfour.com/is/
- The individual team member pages: https://cloudfour.com/is/gerardo/
  - We will no longer be supporting the optional "trivia" feature
  - It would be cool to also show the person's talks (if any)
  - We should try to make the article listing below the bio consistent with other article listings to simplify implementation (showing 4 articles on the first page and more on subsequent pages was _really_ complicated!)

We should prototype these pages to identify any missing pattern work necessary to implement them.

You may also make smaller tweaks along the way to improve the polish of this page's design!

PR #1457, opened 2021-07-29
Prototypes/team

AriannaChau:
## Overview

This PR consists of 3 different pages:

1. The overview of the Team page. Where a list of all team members lives
2. Individual bio page
3. What a second page would look like for articles

In the future, I'd break something like this up into separate PR's. I didn't expect it to be as big as it turned out to be 😅

### Team overview

**Small screens:**
localhost_6006_iframe html_id=prototypes-team-page--list args= viewMode=story

**Large screens:**
localhost_6006_iframe html_id=prototypes-team-page--list args= viewMode=story copy

### Individual bio

**Small screens:**
localhost_6006_iframe html_id=prototypes-team-page--individual-bio args= viewMode=story (1)

**Large screens:**
localhost_6006_iframe html_id=prototypes-team-page--individual-bio args= viewMode=story

### Articles, page 2

**Small screens:**
localhost_6006_iframe html_id=prototypes-team-page--articles-page-2 args= viewMode=story copy

**Large screens:**
localhost_6006_iframe html_id=prototypes-team-page--articles-page-2 args= viewMode=story

## Testing

1. [Deploy link](https://deploy-preview-1457--cloudfour-patterns.netlify.app/?path=/story/prototypes-team-page--list)

---

- Fixes https://github.com/cloudfour/cloudfour.com-patterns/issues/1418


tylersticka (commented):
Overall, I'm very happy with where these landed!

I have some nitpicky feedback on the individual prototypes.

## Team Page - List

1. Can the hover effect be more consistent with that of cards?

## Team Page - Individual

1. The top bio section feels like it could use more bottom padding/whitespace.
2. For the article listing, are you thinking we'd limit the list to three? If so, we'll want to replace the pagination with a link more like "all articles," because it's not within the current scope to have the first page have a different number of articles than subsequent pages.
3. I don't think it'll be within the current scope to paginate talks, unfortunately.

## Team Page - Articles Page 2

1. Instead of the abbreviated "continued" text in the heading, can we include the actual page number?

tylersticka (commented):
Remaining feedback is specific to the listing:

1. I'd still prefer if the hover effect on individual team members on the list page worked more like cards, where the scale transform is constrained to the image _within_ the container. (I try to limit the scaling effects to constrained elements like buttons.)
2. It seems like the team listing is now maxing out at 5 columns in this iteration? I think this looks a little more awkward than the 4-column max used previously.

tylersticka (approved):

PR #1465, opened 2021-08-04
Support b and strong from c-heading--light

tylersticka:
## Overview

I found a case where I wanted to **bold** some words inside of `c-heading--light`, but the `font-weight` only increased one level. This adds overt `b` and `strong` styles to that modifier to make this possible. It also updates the demo to showcase this feature.

## Screenshots

Screen Shot 2021-08-04 at 2 37 20 PM


Paul-Hebert (approved):
LGTM!

PR #1471, opened 2021-08-09
Two o-rhythm improvements

tylersticka:
## Overview

While providing feedback for @AriannaChau's article prototype, I became aware of two issues with `o-rhythm`:

- There was no way to revert to the default rhythm in nested rhythm embeds. So in this PR, I've added the `o-rhythm--default` modifier for this purpose.
- The `--generous-headings` modifier was ineffective on `c-heading` elements. In this PR, I've added a `[class*='heading']` selector to the vertical spacing mixin to serve this purpose in an extensible way.

## Screenshots

These screenshots both show an `o-rhythm--generous-headings` element with a plain `h2` and a `c-heading` of the same level. The spacing before each should be equal.

### Before

Screen Shot 2021-08-09 at 1 57 50 PM

### After

Screen Shot 2021-08-09 at 1 57 02 PM

### DIFF
--- a/src/mixins/_spacing.scss
+++ b/src/mixins/_spacing.scss
@@ -22,7 +22,8 @@
     & > * + h3,
     & > * + h4,
     & > * + h5,
-    & > * + h6 {
+    & > * + h6,
+    & > * + [class*='heading'] {
       margin-top: var(--rhythm-headings);
     }
   }
--- a/src/mixins/_spacing.scss
+++ b/src/mixins/_spacing.scss
@@ -15,7 +15,11 @@
   }
 
   // Where custom properties are supported, also allow headings to receive their
-  // own rhythm.
+  // own rhythm. Note that we also apply this selector to classes containing
+  // `heading`, which covers cases like the `c-heading` component or element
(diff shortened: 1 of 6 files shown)


AriannaChau (dismissed):
It looks good to me.

src/mixins/_spacing.scss
```diff
@@ -22,7 +22,8 @@
     & > * + h3,
     & > * + h4,
     & > * + h5,
-    & > * + h6 {
+    & > * + h6,
+    & > * + [class*='heading'] {
```
calebeby, line 26: @tylersticka Are there use cases for classes other than `.c-heading`?

tylersticka: I noticed that it's common for us to use class names like `{component-name}__heading`. I thought structuring the selector like this felt more indicative of what I was trying to accomplish than the specific component class name. But if others object, I could be convinced otherwise.

calebeby: @tylersticka 👍 Can you put that example in a comment?

calebeby (approved):

PR #1481, opened 2021-08-13
Add responsive variations of spacing utilities

### LINKED ISSUE #1475: Increase vertical padding for container object
I noticed a lot while prototyping that I had to add extra vertical padding to the container object.

Without the extra padding:
Screen Shot 2021-08-11 at 10 38 38 AM

With the extra padding:
Screen Shot 2021-08-11 at 10 38 58 AM

We should increase the vertical padding for the existing object. I'm currently using step 6 on the modular scale in my prototypes.

tylersticka:
## Overview

The team page prototypes called for increasing the vertical padding of `o-container`, but in practice this had some unintentional consequences. I also noticed that this would only solve for one element's unique spacing requirements, when the same need befell several elements per prototype.

So this PR expands our existing spacing utilities to include responsive variations, as that was the only capability of the existing prototype CSS that wasn't possible before.

## Screenshots

responsive-spacing

---

- Fixes #1475 

/CC @AriannaChau


gerardo-rodriguez (approved):
LGTM!

PR #1489, opened 2021-08-17
Add subheading support to Heading

### LINKED ISSUE #1464: Heading with subtext
On the speaking event prototype, there is a heading with smaller subtext. On small screens the text appears beneath the heading. On larger screens, the text next to the heading. We need a way to place the smaller text next to the heading.

**Small Screens**
Screen Shot 2021-08-02 at 4 24 09 PM

**Large Screens**
Screen Shot 2021-08-02 at 4 23 51 PM

tylersticka:
## Overview

This PR refactors our heading component class and template to accept a `subheading` option.

It differs slightly from the prototype in that the subheading has a `medium` font weight instead of `normal` because I thought that made the two segments of the heading appear more closely related.

The reasons for the refactor are as follows:

- The previous structure made it challenging to correctly align a permalink if a subheading is present.
- The previous template assumed that permalinks were the sole reason the complexity of the markup might change, but subheadings complicated this.
- The previous stories, which each had their own separate demo file, made testing various combinations of options challenging, leading me to rewrite the stories using controls.
- Because the `content` and `subheading` options may be properties or blocks, I found it necessary to add some additional `makeTwig*` utility functions for defining the `templateSource` intelligently.

I marked this as a `minor` change because the template syntax has not changed.

See the docs and code comments for specific implementation details.

## Screenshots

Screenshot 2021-08-17 at 12-48-17 Storybook

## Testing

[On the deploy preview...](https://deploy-preview-1489--cloudfour-patterns.netlify.app/)

1. Verify that Heading stories work as intended
2. Check that the Heading story code samples change accurately depending on controls
3. Verify that the Speaking Event prototype's "Watch" heading and subheading work as intended

---

- Fixes #1464 

/CC @dromo77 @calebeby


src/components/heading/heading.stories.mdx
```diff
@@ -1,121 +1,196 @@
-import { Story, Canvas, Meta } from '@storybook/addon-docs/blocks';
-import levelsDemo from './demo/levels.twig';
-import permalinksDemo from './demo/permalinks.twig';
-import weightsDemo from './demo/weights.twig';
-
-<Meta title="Components/Heading" />
+import { Story, Canvas, Meta, ArgsTable } from '@storybook/addon-docs';
+import { makeTwigEmbedIfHtml } from '../../make-twig-include.js';
```
calebeby, line 2:
```suggestion
import { makeTwigEmbedIfHtml } from '../../make-twig-include';
```

src/make-twig-include.js
```diff
@@ -1,13 +1,80 @@
...
+  return `{% embed '${path}'${argsString} only %}${blocks.join('\n')}
+{% endembed %}`;
+};
+
+/**
+ * @param {string} path
+ * @param {any} args
+ * @param {string[]} blockNames
+ */
+export const makeTwigEmbedIfHtml = (path, args = {}, blockNames = []) => {
```
calebeby, line 66: @tylersticka Are there use cases where passing a specific array of block names is required, or could we change this to loop through all the keys of args to detect which args should be blocks?

calebeby (commented):
@tylersticka The source code generation thing is so cool! I love how it switches between embed and include!

src/make-twig-include.js
tylersticka: > loop through all the keys of args to detect which args should be blocks?

Do you know a way we could do this?

calebeby:
```js
  for (const [key, value] of Object.keys(args)) {
    if (
      typeof value === 'string' &&
      /<\/?[a-z][\S\s]*>/i.test(value)
    ) {
      filteredBlockNames.push(value);
    }
  }
```

tylersticka: The potential problem with that is that it assumes every property has an equivalent block. But most templates do not. And sometimes the block and the property aren't equivalent. For example:

```twig
<header>
  {% block header %}
    <h1>{{heading}}</h1>
  {% endblock %}
</header>
```

calebeby: I'm not sure I understand how properties and blocks work together 🤔 

If a property has HTML in it, are there cases where it will need to just be passed as a prop and not as a block?

tylersticka: Props and blocks are different things. It's sort of like with web components the difference between an attribute and a slot.

In this case, they happen to be equivalent because I wanted someone to be able to use a simple property _or_ use a block for markup...

```twig
{# The template is written like this: #}
{% block content %}
  {{content}}
{% endblock %}

{# Which means these two things are equivalent: #}
{% include 'path/to/example.twig' with { content: 'foo' } %}
{% embed 'path/to/example.twig' %}{% block content %}foo{% endblock %}{% endembed %}
```

But there'd be nothing to stop me decoupling them completely:

```twig
<p>{{content}}</p>
<button>
  {% block content %}
    Hey there, {{firstName}} {{lastName}}!
  {% endblock %}
</button>
```

In which case those two examples would have entirely different results.

calebeby (approved):

PR #1491, opened 2021-08-20
Add o-container__fill elements

### LINKED ISSUE #1474: Give cards the option to be full width at small screens:
For one of our prototypes, the contained card used for the bio typically has padding and rounded corners on all screen sizes:

Screen Shot 2021-08-11 at 10 33 26 AM

However, it would be nice if we could have the option for the contained cards to be full width on small screens, then becoming more contained on larger screens:

Screen Shot 2021-08-11 at 10 34 21 AM

Screen Shot 2021-08-11 at 10 34 43 AM

tylersticka:
## Overview

This adds two elements to `o-container` to allow child elements to fill that element's inline padding and optionally re-apply that padding. This allows certain elements to "fill" the available space while aligning with adjacent content.

I was originally tackling this as a modifier to `c-card`, but realized that the styles had more to do with the container object than cards, and that it might be nice to have access to this without always relying on `c-card`. But I'm open to feedback on that.

## Screenshots

Screen Shot 2021-08-20 at 9 48 36 AM

Screen Shot 2021-08-20 at 9 48 56 AM

Screen Shot 2021-08-20 at 9 49 09 AM

## Testing

On the deploy preview...

1. Review Objects ▸ Container
1. Review Prototypes ▸ Single Article bio element

---

- Fixes #1474 

/CC @AriannaChau


gerardo-rodriguez:
@tylersticka For my own understanding, these two appear the same visually to me. That _is_ expected. Correct? 🔍 🤔 😅 🤷 

Screen Shot 2021-08-20 at 3 34 43 PM

gerardo-rodriguez:
> I was originally tackling this as a modifier to `c-card`, but realized that the styles had more to do with the container object than cards, and that it might be nice to have access to this without always relying on `c-card`. But I'm open to feedback on that.

@tylersticka Based on my understanding, your proposed solution makes sense. 😉

tylersticka:
> For my own understanding, these two appear the same visually to me. That _is_ expected. Correct? 🔍 🤔 😅 🤷

@gerardo-rodriguez So the difference is that the first card has its default padding, whereas the second card has the padding of the container. There are viewport sizes where the amount of padding will be the same, but below or above that they will differ.

It's easier to see if you resize the browser and compare the way the content is behaving to what's outside of it. Toggling the Outlines plugin in the canvas toolbar can also help:

Screen Shot 2021-08-20 at 3 40 22 PM

gerardo-rodriguez:
@tylersticka It is expected that there be rounded corners when the container is still touching the horizontal edges, is this correct? 🔍 👀 

Screen Shot 2021-08-20 at 3 40 57 PM

tylersticka:
> It is expected that there be rounded corners when the container is still touching the horizontal edges, is this correct? 🔍 👀

@gerardo-rodriguez So, here was my journey with this...

We had a discussion _very_ early on in this project where we decided to apply the padding to a different element than the content width, because designers didn't like setting a content width only for padding to be subtracted from it.

Separately, we also embraced fluid values for certain padding on elements, so narrow screens would have minimal padding and larger screens would have larger amounts of padding.

One unfortunate side effect of this, though, is that it means if you actually need a breakpoint that applies to when a _padded container_ is narrower than the viewport, the math breaks my brain. If you add the smallest padding value to the media query, you just delay the rounded corners popping in. If you add the largest padding value, which doesn't take effect until wider viewports, then the rounded corners pop in too late.

I couldn't think of a good solution that didn't involve brain-breaking math, and decided it wasn't a big deal. But that might be my cold brain talking... let me know what you think.

gerardo-rodriguez:
> I couldn't think of a good solution that didn't involve brain-breaking math, and decided it wasn't a big deal.

Works for me, thanks, @tylersticka!

gerardo-rodriguez (approved):
I wasn't able to fully wrap my head around the CSS but it LGTM! 🙂

tylersticka:
> I wasn't able to fully wrap my head around the CSS but it LGTM.

@gerardo-rodriguez Are those portions something you'd like to see more comments on? If so, can you highlight the sections that are most confusing?

If you find it confusing, there's a good chance others might, too, y'know?

src/mixins/_fluid.scss
```diff
@@ -108,6 +108,23 @@
...
+  @media (min-width: $min-width) {
+    margin-inline-end: fluid-calc($min-width, $max-width, $min, $max);
+    margin-inline-start: fluid-calc($min-width, $max-width, $min, $max);
+  }
+
+  @media (min-width: $max-width) {
+    margin-inline-end: $max;
+    margin-inline-start: $max;
+  }
+}
```
gerardo-rodriguez, line 126: > can you highlight the sections that are most confusing?

Some extra comments here would be helpful. The rest is okay without more comments. 👍 

Thanks, @tylersticka!

tylersticka: Gotcha. Since this mixin is almost exactly the same as all the others in this file, I might check to see if any of the others are already commented first before singling this one out. (I wish Sass let you use string interpolation on property names, then we wouldn't need property-specific versions of these mixins.)

tylersticka: @gerardo-rodriguez I unfortunately don't think that I have any additional comments to provide other than those at the top of the file: https://github.com/cloudfour/cloudfour.com-patterns/blob/a2d9e84b77619bfd71cc7cdd946c4a2769141990/src/mixins/_fluid.scss#L3-L14

tylersticka:
> I couldn't think of a good solution that didn't involve brain-breaking math

@gerardo-rodriguez FYI, I was over-complicating this. Instead of doing the math, I just manually adjusted the breakpoint dimensions until the issue was resolved, and added a comment explaining why there were magic numbers in use. The latest iteration may not be _pixel-perfect_, but you should see a lot less dissonance between the `max-width` and the rounded corners being restored.

src/mixins/_fluid.scss
gerardo-rodriguez: Thanks, @tylersticka, I totally missed this comment! 👍

gerardo-rodriguez (approved):
I approve, thanks, @tylersticka! 🎉

PR #1498, opened 2021-08-23
Add options to Comments and Comment Forms to allow submission

Paul-Hebert:
## Overview

This makes some changes to our components to allow the Wordpress site to properly set up comment form submission.

I'm like 90% sure this will be all we need. There's a chance we'll need to do a follow-up PR if things don't work as expected in WP land.

### Comment Form

The comment form got a new `hidden_inputs` block which allows you to pass in `<input type="hidden">` elements, which Wordpress requires for comment submission. (We technically could have done this already with the `prompt` block, but that felt kinda gross and I had to make other changes.)

### Comment

This component required more changes so that it can configure its nested Comment Form:

#### Properties

- `reply_form_action`: the action for the reply form
- `reply_form_method`: the method for the reply form

#### Blocks

- `reply_form_hidden_inputs`: This allows you to pass in content to the form's new `hidden_inputs` block
- `reply_form`: This allows you to replace the whole reply form. We shouldn't currently need this but it seemed like a useful escape hatch.

## Screenshots

No visual changes

---

(Hiding whitespace changes will simplify review)

### DIFF
--- a/.changeset/mean-spies-glow.md
+++ b/.changeset/mean-spies-glow.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add new options to Comment and Comment Form components to support comment submission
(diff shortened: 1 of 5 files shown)


gerardo-rodriguez (approved):
🎉

PR #1499, opened 2021-08-24
Feature/text utilities

### LINKED ISSUE #1436: Text utilities
The [speaking prototype](http://localhost:6006/iframe.html?id=prototypes-speaking-listing--example&args=&viewMode=story) includes cards that showcase past talks. The description limits the number of visible events to 3. If there are more events, there is an additional bit of text that says "+ X more". I'm not sure if this needs to be it's own pattern, or if text utilities would be best. For example, `u-textBrandPrimary` and `u-textNoWrap`.

**Example**
Screen Shot 2021-07-22 at 2 11 29 PM

gerardo-rodriguez:
## Overview

Please include a summary of the change and which issue is fixed.
Please also include relevant motivation and context.
List any dependencies that are required for this change.

## Screenshots

Visual changes should include a screenshot.

## Testing

1. instructions for reviewers
1. to test your changes

---

- Closes #1436

### DIFF
--- a/src/utilities/text/text.scss
+++ b/src/utilities/text/text.scss
@@ -1,0 +1,9 @@
+@use "../../compiled/tokens/scss/color";
+
+.u-textNoWrap {
+  white-space: nowrap !important;
+}
+
+.u-textColorBrandPrimary {
+  color: color.$brand-primary !important;
+}
--- a/src/utilities/text/demo/color.twig
+++ b/src/utilities/text/demo/color.twig
@@ -1,0 +1,1 @@
+<p class="u-textColorBrandPrimary">This show an example text that has the utility class applied.</p>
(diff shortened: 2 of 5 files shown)


src/utilities/text/text.scss
```diff
@@ -0,0 +1,9 @@
+@use "../../compiled/tokens/scss/color";
+
+.u-textNoWrap {
+  white-space: nowrap !important;
+}
+
+.u-textColorBrandPrimary {
+  color: color.$brand-primary !important;
```
tylersticka, line 8: How does this work when themes are applied?

tylersticka (changes requested):
Please use BEMIT naming conventions rather than SUIT. (Kebab-case instead of camel-case.)

src/utilities/text/text.scss
gerardo-rodriguez: Great question! Hmm...I'll have to think about this. 🤔

Issue #1473, opened 2021-08-10
Full-page prototypes should be simpler to initialize with working header/footer


AriannaChau:
While prototyping several pages and including the Sky Nav, I thought I had an issue with the Menu not collapsing on small screens, see here:
<img width="501" alt="Screen Shot 2021-08-10 at 1 36 18 PM" src="https://user-images.githubusercontent.com/24928337/128931427-0068058d-e82a-4759-8871-ab527712e33c.png">

It turns out you need to use JavaScript to initialize the Sky Nac to make it work as intended. I think there's room for improvement in making this implementation easier.

PR #1503, opened 2021-08-25
Initialize sky nav JS for all prototypes that use the sky nav

calebeby:
## Overview

Closes https://github.com/cloudfour/cloudfour.com-patterns/issues/1473

I made a custom hook `useSkyNav` that initializes the sky nav. In prototypes that have the sky nav, the hook needs to be called:

```js
export const SomePrototypeStory = () => {
  useSkyNav()
  return someTwigFile()
}
```

## Testing

Make sure that all the prototypes that use the sky nav initialize the JS correctly for small screens

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@cloudfour/patterns",
-  "version": "3.1.1",
+  "version": "4.1.0",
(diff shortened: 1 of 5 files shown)


tylersticka (approved):
Nice and simple! Works great

PR #1516, opened 2021-08-30
Button Swap component

### LINKED ISSUE #1508: Create more accessible notifications subscription pattern
This issue was created in response to further exploration done for #1094. 

The goal is to create a more accessible notifications subscription pattern.

Use the [following](https://cloudfour.slack.com/archives/C04AEEVK2/p1630088630056200?thread_ts=1630014315.009200&cid=C04AEEVK2) as a reference:
> Here is one without aria-live that uses focus: https://codepen.io/calebeby/pen/xxrGXWb
> 
> Here is one with aria-live that doesn’t use focus:
https://codepen.io/gerardorodriguez/pen/dyRoZWZ

Ensure the UX is inclusive (including VoiceOver + keyboard UX).

gerardo-rodriguez:
## Overview

This PR adds a Button Swap component with the intent to be used to subscribe/unsubscribe from push notifications on article pages.

## Screenshots

screencapture-localhost-6006-2021-09-03-13_28_00

## Testing

1. Review the component docs: https://deploy-preview-1516--cloudfour-patterns.netlify.app/?path=/docs/components-button-swap--default-story

---

- Closes #1508


src/components/button-swap/button-swap.test.ts
```diff
@@ -0,0 +1,185 @@
...
+);
+
+test(
+  'Swap UI state when clicked',
+  withBrowser(async ({ utils, screen, user }) => {
+    await utils.injectHTML(await componentMarkup());
+    await loadGlobalCSS(utils);
+    // I'd like to avoid using a test ID, but I couldn't figure out a different way.
+    // @todo Can this be done without at test ID?
+    await initJS(utils, await screen.getByTestId('test-id'));
```
gerardo-rodriguez, line 65: @cloudfour/dev Any thoughts? Suggestions?

src/components/button-swap/button-swap.stories.mdx
```diff
@@ -0,0 +1,163 @@
...
+
+<Canvas>
+  <Story
+    name="Default"
+    inline={false}
+    args={{
+      first_button_visually_hidden_text: 'Notifications have been turned off.',
+      first_button_label: 'Give me all the notifications!',
+      second_button_visually_hidden_text: 'Notifications have been turned on.',
+      second_button_label: 'Just kidding, stop notifications, please.',
```
tylersticka, line 54: These arguments are kind of a mouthful! I wonder if they could be simplified? Idea:

```
initial_label
initial_visual_label
swapped_label
swapped_visual_label
```

src/components/button-swap/button-swap.stories.mdx
```diff
@@ -0,0 +1,163 @@
...
+
+```js
+initButtonSwap(buttonSwapEl, options);
+```
+
+### Parameters
+
+#### `buttonSwapEl`
+
+The `.js-c-button-swap` Button Swap wrapper element.
```
tylersticka, line 101: The `c-` component is not necessary for `js-` selectors. See elastic text areas for a comparison: https://v-next--cloudfour-patterns.netlify.app/?path=/docs/components-input--elastic-textarea#javascript-instructions

src/components/button-swap/button-swap.stories.mdx
```diff
@@ -0,0 +1,163 @@
...
+
+#### `options` (optional)
+
+An object which allows you to optionally pass `firstBtnCallback` and `secondBtnCallback` functions to be called when each button is clicked.
+
+```js
+options?: {
+  firstBtnCallback?: (event: Event) => void;
+  secondBtnCallback?: (event: Event) => void;
```
tylersticka, line 110: I think `initialCallback` and `swappedCallback` might be more intuitive? Maybe not, though. I don't know why "first" and "second" are messing with my head.

tylersticka (commented):
Some small comments left inline.

One question I have for @cloudfour/dev is whether or not this would make sense to be packaged as a small web component since it seems to depend on JavaScript as-is?

gerardo-rodriguez:
> One question I have for @cloudfour/dev is whether or not this would make sense to be packaged as a small web component since it seems to depend on JavaScript as-is?

Interesting idea, I hadn't considered this! I can definitely look into this as an option. It would be fun for me write a small web component. 🙂

calebeby:
> One question I have for @cloudfour/dev is whether or not this would make sense to be packaged as a small web component since it seems to depend on JavaScript as-is?

We might be able to do this if we render the child components with slots (otherwise we'd have problems with the children buttons not being able to be statically rendered and we'd have to load twig client-side):

```html
<button-swap>
  {% include '@cloudfour/..../button' with ... %}
  {% include '@cloudfour/..../button' with ... %}
</button-swap>
```

Benefits:
- Don't have to initialize JS per-instance, just once initially, plus initialize listeners on the children if needed
- Slightly less verbose than `{% include `

Downsides:
- Probably more JS than the current solution

src/components/button-swap/button-swap.twig
```diff
@@ -0,0 +1,28 @@
+{#
+  This pattern currently only supports a start icon.
+  End icon support can be added in the future if needed.
+#}
+<section role="region" data-testid="test-id" class="c-button-swap js-c-button-swap{% if class %} {{ class }}{% endif %}">
```
Paul-Hebert, line 5: Is this role necessary? My impression was `section` implicitly got the region role?

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Region_role

src/components/button-swap/button-swap.stories.mdx
```diff
@@ -0,0 +1,163 @@
...
+    }}
+  >
+    {(args) => buttonSwapStory(args)}
+  </Story>
+</Canvas>
+
+## Template Properties
+
+- `class` (string): Appends a CSS class(es) to the root element.
+- `content_start_icon` (string, default `'bell'`): The button [icon](/docs/design-icons--page) (the same icon is applied to both buttons).
```
Paul-Hebert, line 81: Would it make more sense to default to no icon? Or is an icon required? 🤔

src/components/button-swap/button-swap.stories.mdx
Paul-Hebert: I like initial/swapped

src/components/button-swap/button-swap.test.ts
Paul-Hebert: Could we use `getByRole('region')` or something to grab the `section`?

src/components/button-swap/button-swap.ts
```diff
@@ -0,0 +1,86 @@
...
+    firstBtnWrapper.hidden = false;
+
+    const statusMsg = getStatusMessage(firstBtnWrapper);
+    statusMsg.setAttribute('role', 'alert');
+    statusMsg.focus();
+
+    if (secondBtnCallback) {
+      secondBtnCallback(event);
+    }
+  };
```
Paul-Hebert, line 72: I wonder if both buttons could call the same method and pass in some vars? Something like this?

```js
  const onButtonClick = ({event, hiddenWrapper, visibleWrapper, callback}) => {
    hiddenWrapper.hidden = true;
    visibleWrapper.hidden = false;

    const statusMsg = getStatusMessage(visibleWrapper);
    statusMsg.setAttribute('role', 'alert');
    statusMsg.focus();

    if (callback) {
      callback(event);
    }
  };
  
firstBtn.addEventListener('click', (event) => {
  onButtonClick({
    event,
    hiddenWrapper: firstBtnWrapper,
    visibleWrapper: secondBtnWrapper,
    callback: firstBtnCallback
  });
});
```

That might dry things up a bit, and reduce the amount of duplicated code we're shipping

Paul-Hebert: (Though I suppose it would make the `destroy` logic a bit more complicated.)

src/components/button-swap/button-swap.twig
```diff
@@ -0,0 +1,28 @@
...
+  <div class="c-button-swap__subscribe js-c-button-swap__first-btn-wrapper">
+    <span tabindex="-1" role="status" class="u-hidden-visually js-c-button-swap__message">
+      {{ first_button_visually_hidden_text|default('Unsubscribed from notifications') }}
```
Paul-Hebert, line 8: Nit: I worry a bit that someone might mishear this as "Unsubscribe" instead of "Unsubscribed"

I wonder if adding "Currently" to the stat of these messages might make them clearer?

src/components/button-swap/button-swap.twig
```diff
@@ -0,0 +1,28 @@
+{#
+  This pattern currently only supports a start icon.
+  End icon support can be added in the future if needed.
+#}
+<section role="region" data-testid="test-id" class="c-button-swap js-c-button-swap{% if class %} {{ class }}{% endif %}">
+  <div class="c-button-swap__subscribe js-c-button-swap__first-btn-wrapper">
+    <span tabindex="-1" role="status" class="u-hidden-visually js-c-button-swap__message">
```
Paul-Hebert, line 7: I'm curious about the reasoning for adding `tabindex` to these. Could this lead to keyboard users focusing a hidden element they can't see?

Paul-Hebert (commented):
This is looking great @gerardo-rodriguez ! I left some thoughts and questions inline.

> One question I have for @cloudfour/dev is whether or not this would make sense to be packaged as a small web component since it seems to depend on JavaScript as-is?

This is a great question. I agree with @calebeby 's takeaways overall. I'm not sure how much value it being a web component would add if we only have one location we're currently planning to use it but I like the idea of us starting to use more little web components for progressive enhancement.

src/components/button-swap/button-swap.twig
calebeby: Only if it has an accessible name: https://w3c.github.io/html-aam/index.html#el-section

Paul-Hebert: Ahhh good call. So this gives it a role without a name? Should we give it a name instead? I wonder how much benefit an unnamed region adds here?

src/components/button-swap/button-swap.twig
Paul-Hebert: Ignore this... I was confusing `-1` and `0` tabindexes

calebeby: https://webaim.org/techniques/keyboard/tabindex#negative-one

The -1 value makes it so that you can focus it from javascript, but it is skipped in the natural tab order

src/components/button-swap/button-swap.test.ts
calebeby: If you end up adding an accessible name to the section, the name can be used to filter as well:

```js
await screen.getByRole('region', { name: /something/i })
```

src/components/button-swap/button-swap.stories.mdx
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1516/commits/3da46ecee1dbf6b294f0c9113681ee7ef8f4942b

src/components/button-swap/button-swap.test.ts
gerardo-rodriguez: > Could we use `getByRole('region')` or something to grab the `section`?

@calebeby @Paul-Hebert I tried the following initially:
```js
await initJS(utils, await screen.getByRole('region'));
```
But that would result in this failed test:
```
Unable to find an accessible element with the role "region"

    Here are the accessible roles:

      document:

      Name "":
      <body>
      <section data-testid="test-id" class="c-button-swap js-c-button-swap">[...]</section>


    </body>

      --------------------------------------------------
      status:

      Name "":
      <span
      tabindex="-1"
      role="status"
      class="u-hidden-visually js-c-button-swap__message"
    >

          Unsubscribed from notifications

    </span>

      --------------------------------------------------
      button:

      Name "Get notifications":
      <button class="c-button js-c-button-swap__button">[...]</button>

      --------------------------------------------------

    Within: #document
```

gerardo-rodriguez: Oh, maybe I need to give it a name:
> Using the `<section>` element will automatically communicate a section has a role of `region` if it is given an accessible name. Developers should always prefer using the correct semantic HTML element, in this case `<section>`, over using ARIA.

Source: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Region_role#description

src/components/button-swap/button-swap.stories.mdx
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1516/commits/47177615cbb93df03875f2b536808393242d5cc7

src/components/button-swap/button-swap.twig
gerardo-rodriguez: I didn't know an accessible name was required for the `section` element to have a role of `region`! 🎓

src/components/button-swap/button-swap.twig
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1516/commits/32da47af9c44fcc0927c62fd0ea699bb379c9b72

src/components/button-swap/button-swap.stories.mdx
gerardo-rodriguez: Hmm...good question. I'll have to think about that. 🤔

src/components/button-swap/button-swap.stories.mdx
gerardo-rodriguez: Changed! https://github.com/cloudfour/cloudfour.com-patterns/pull/1516/commits/c3ca0f3d0fcb5f979d18fae517a2e528735ba450

src/components/button-swap/button-swap.ts
gerardo-rodriguez: @Paul-Hebert Something like the following?

https://github.com/cloudfour/cloudfour.com-patterns/pull/1516/commits/714e30c29dbc0f5fe5fef9102623790ecb464a30

gerardo-rodriguez:
> > One question I have for @cloudfour/dev is whether or not this would make sense to be packaged as a small web component since it seems to depend on JavaScript as-is?
> 
> This is a great question. I agree with @calebeby 's takeaways overall. I'm not sure how much value it being a web component would add if we only have one location we're currently planning to use it but I like the idea of us starting to use more little web components for progressive enhancement.

Thanks for all your input! Regarding whether or not to make this a web component, my takeaways are:
- Yes, let's start thinking about using more web components!
- This particular pattern doesn't seem like the best use case, though.

If I am understanding correctly, then I will continue addressing PR feedback and leave this implementation as-is (no web component). Do we all agree @tylersticka, @calebeby, @Paul-Hebert? 

Thanks!

tylersticka:
@gerardo-rodriguez I threw out the idea to make sure it was considered, but I trust the team to decide if it's the right option.

src/components/button-swap/button-swap.twig
```diff
@@ -0,0 +1,29 @@
+{#
+  This pattern currently only supports a start icon.
+  End icon support can be added in the future if needed.
+#}
+<section aria-labelledby="some-cool-id" class="c-button-swap js-button-swap{% if class %} {{ class }}{% endif %}">
+  <h3 class="u-hidden-visually" id="blah">THis is the name</h3>
```
gerardo-rodriguez, line 6:
```suggestion
  <h3 class="u-hidden-visually" id="some-cool-id">THis is the name</h3>
```

src/components/button-swap/button-swap.twig
```diff
@@ -0,0 +1,29 @@
+{#
+  This pattern currently only supports a start icon.
+  End icon support can be added in the future if needed.
+#}
+<section aria-labelledby="some-cool-id" class="c-button-swap js-button-swap{% if class %} {{ class }}{% endif %}">
+  <h3 class="u-hidden-visually" id="some-cool-id">THis is the name</h3>
```
gerardo-rodriguez, line 6: @cloudfour/dev For the `<section>` element to have a role of `region`, I need to give it an accessible name via an `aria` attribute _or_ explicitly set the `role` to `"region"`.

I'm curious what your thoughts are on which solution feels better. 

Original implementation:
- No heading existed
- The `<section>` element had an explicit `role="region"` attribute

New implementation:
- The `<section>` element has an `aria-labelledby` attribute but I'm unclear what the `aria-labelledby` value should point to.

Questions

1. Should the pattern include a visually hidden heading with a matching `id`?
    - Its seems the heading tag (e.g. `h3` or `h2`), `id` and heading text values should be properties that can be passed in.
2. If I add a visually hidden heading, how do I ensure the `id` is unique? I'm not sure. 🤔 
3. Should the pattern not have a visually hidden heading and instead require you pass in a value for the `aria-labelledby` putting the responsibility on the pattern consumer to create the element (with `id`) that this pattern should point to?
4. Should I avoid all of these challenges and explicitly set a `role="region"` value on the `<section>` element?  
    - While this _technically_ should work, it doesn't seem like the best UX when using an assistive technology like a screen reader. It'd be better if the region had an accessible name. So now that I think about it more, I feel this is not an acceptable solution.

Thoughts?

tylersticka: Do the answers to any of these questions differ when we consider this within the context of the newsletter + notification sign-up UI? Since that is currently the only context for this component, just want to make sure we're considering that whole chunk of UI before we dive _too_ deeply into the questions of each individual pattern.

(If we're not sure, we may want to pause on developing individual components so we can prototype the whole experience of that element, _then_ carve off which piece or pieces deserve to be patterns.)

src/components/button-swap/button-swap.ts
Paul-Hebert: Exactly!

src/components/button-swap/button-swap.twig
Paul-Hebert: I wonder if the `section` is actually adding helpful context here. If we can't think of a name that describes the section, then is it helpful to flag it as a region?

Going off of what Tyler said above, maybe this makes more sense when it is used as part of a larger form? In which case this component maybe doesn't need its own semantic role?

Paul-Hebert:
>If I am understanding correctly, then I will continue addressing PR feedback and leave this implementation as-is (no web component). Do we all agree @tylersticka, @calebeby, @Paul-Hebert?

Makes sense to me :+1:

src/components/button-swap/button-swap.twig
gerardo-rodriguez: > Going off of what Tyler said above, maybe this makes more sense when it is used as part of a larger form? In which case this component maybe doesn't need its own semantic role?

Fair! Thanks! 🙂

src/components/button-swap/button-swap.stories.mdx
gerardo-rodriguez: Thanks for the suggestion, @Paul-Hebert!

After chatting with you on a call and realizing making this change means potentially modifying Buttom template logic, I agree, this doesn't feel like a blocker. Our only use case for this pattern expects a bell icon. I don't see any use cases in the immediate future that would exclude the icon, therefore, I'll leave it as-is for now.

If we _do_ find a new use case where we need this pattern without an icon, we can add in the feature at that point. 😉

gerardo-rodriguez:
Thanks for the feedback everyone! This PR is ready for another review. 😄

src/components/button-swap/button-swap.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    await loadGlobalCSS(utils);
+
+    // Visually hidden text for a more inclusive UX
+    const statusMsg = await screen.getByRole('status');
+    await expect(statusMsg).toHaveTextContent(
+      /^currently unsubscribed from notifications$/i
+    );
+
+    const statusMsgs = await screen.getAllByRole('status');
+    expect(statusMsgs.length).toBe(1);
```
calebeby, line 43: You can use just `getByRole` since `getBy` matchers will throw if there are 0 results or more than one result, it will only pass if there is exactly 1: https://testing-library.com/docs/dom-testing-library/cheatsheet#queries

src/components/button-swap/button-swap.test.ts
```diff
@@ -0,0 +1,177 @@
...
+        initial_label: 'Unsubscribed',
+        swapped_label: 'Subscribed',
+      })
+    );
+    await loadGlobalCSS(utils);
+    await initJS(utils);
+
+    // Visually hidden text for a more inclusive UX
+    let statusMsg = await screen.getByRole('status');
+    await expect(statusMsg).toHaveTextContent(/^unsubscribed$/i);
```
calebeby, line 134: Nit: It might make sense to use a string directly instead of a regex here since the static value is written directly in this test

src/components/button-swap/button-swap.stories.mdx
Paul-Hebert: Sounds good to me :+1:

src/components/button-swap/button-swap.test.ts
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1516/commits/dc3cca71b226578e8799fb575d1115a646a5fe87

src/components/button-swap/button-swap.test.ts
```diff
@@ -0,0 +1,177 @@
...
+  })
+);
+
+test(
+  'Swap UI state when clicked',
+  withBrowser(async ({ utils, screen, user }) => {
+    await utils.injectHTML(await componentMarkup());
+    await loadGlobalCSS(utils);
+    // I'd like to avoid using a test ID, but I couldn't figure out a different way.
+    // @todo Can this be done without at test ID?
```
Paul-Hebert, line 63: Is this comment still relevant?

Paul-Hebert (approved):
This is looking great! (And sounds even better 😉 )

This looks good to go besides Caleb's feedback and one nit I left inline.

src/components/button-swap/button-swap.test.ts
gerardo-rodriguez: Oops, great catch! This comment is no longer relevant. 🙂

gerardo-rodriguez: Deleted! https://github.com/cloudfour/cloudfour.com-patterns/pull/1516/commits/a64e04b59a7b0ca0768d68d02661cb6d653ebd93

src/components/button-swap/button-swap.test.ts
gerardo-rodriguez: Thank you! 🎓

gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1516/commits/2ddbb739661f82016db66aad11fe6d30b269d2bc

gerardo-rodriguez:
Thanks, @Paul-Hebert! 🎉 

@calebeby Thanks for your suggestions, they have been addressed. Ready for another review. 😉

Paul-Hebert (approved):
🚢 🎉

src/components/button-swap/button-swap.ts
```diff
@@ -0,0 +1,98 @@
...
+  wrapper.querySelector('.js-button-swap__message') as HTMLElement;
+
+/**
+ * Performs click actions
+ */
+const onClick = ({
+  event,
+  hiddenWrapper,
+  visibleWrapper,
+  callback,
```
calebeby, line 20: Nit: make this required and remove the if statement below?

calebeby (approved):
Looks great!

src/components/button-swap/button-swap.ts
gerardo-rodriguez: Sure! 👍

gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1516/commits/a7a49c435ee8d4b801cecf5b133241f520a3b1b5

gerardo-rodriguez:
@Paul-Hebert @calebeby My last pushed commit dismissed your approvals. Can one of you approve so I can merge, please? 🙂

gerardo-rodriguez:
> @Paul-Hebert @calebeby My last pushed commit dismissed your approvals. Can one of you approve so I can merge, please? 🙂

Oops, hold up, CI failed. 🔍 👀

gerardo-rodriguez:
> > @Paul-Hebert @calebeby My last pushed commit dismissed your approvals. Can one of you approve so I can merge, please? 🙂
> 
> Oops, hold up, CI failed. 🔍 👀

@Paul-Hebert @calebeby Okay, fixed! (Yay for type checking!)

If one of you can approve, I'll go ahead and merge. Thanks! 😉

calebeby (approved):

Issue #1514, opened 2021-08-30
Sky Nav logo feels too large on small screens


tylersticka:
On large screens, the Sky Nav logo feels nicely balanced (IMO):

<img width="1700" alt="Screen Shot 2021-08-30 at 9 40 15 AM" src="https://user-images.githubusercontent.com/69633/131373916-14818817-5bf1-49b1-b7c3-5f088b406ea1.png">

But on smaller screens, it feels a bit confined:

<img width="358" alt="Screen Shot 2021-08-30 at 9 41 02 AM" src="https://user-images.githubusercontent.com/69633/131374004-164f5573-0606-435d-b364-b8ba378f86c6.png">

Ideally it'd be a bit more balanced with the size of the button (this is just a quick mockup):

<img width="358" alt="Screen Shot 2021-08-30 at 9 41 02 AM adjusted" src="https://user-images.githubusercontent.com/69633/131374198-ef4bdb4b-8560-4bd0-b1d6-ed9fe4b86666.png">

This issue is labeled `design` because it involves design thinking but `dev` because I believe the adjustment can be made directly to the Sky Nav component without a separate prototyping step.

calebeby:
Oops forgot to close this 🤦

PR #1517, opened 2021-08-31
Scale down logo more for small screens

calebeby:
## Overview

https://github.com/cloudfour/cloudfour.com-patterns/issues/1514

## Screenshots

Screen Shot 2021-08-31 at 9 19 52 AM

## Testing

https://deploy-preview-1517--cloudfour-patterns.netlify.app/iframe.html?id=components-sky-nav--dark&viewMode=story&args=&globals=

### DIFF
--- a/src/components/sky-nav/sky-nav.scss
+++ b/src/components/sky-nav/sky-nav.scss
@@ -182,8 +182,14 @@ $_masthead-height-sm: ms.step(7);
   margin: (ms.step(-6) * -1); /* 2 */
   transition: opacity transition.$slow ease.$out,
     transform transition.$quick ease.$out;
-  width: ms.step(8);
-
+  width: fluid.fluid-calc(breakpoint.$s, breakpoint.$l, ms.step(7), ms.step(9));
+  @media (min-width: breakpoint.$l) {
+    width: ms.step(8);
+  }
+  @media (max-width: breakpoint.$s) {
+    width: ms.step(7);
+  }
+  
   /**
    * When the containing logo is hovered, scale up this element. Scaling this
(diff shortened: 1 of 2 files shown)


src/components/sky-nav/sky-nav.scss
```diff
@@ -182,7 +182,13 @@ $_masthead-height-sm: ms.step(7);
   margin: (ms.step(-6) * -1); /* 2 */
   transition: opacity transition.$slow ease.$out,
     transform transition.$quick ease.$out;
-  width: ms.step(8);
+  width: fluid.fluid-calc(breakpoint.$s, breakpoint.$l, ms.step(7), ms.step(9));
```
calebeby, line 185: @tylersticka Any idea why the last number has to be 9? If I change it to 8, at the large breakpoint it is very discontinuous and there is a large jump. I tried to figure out what was going wrong but I couldn't figure it out. It might have something to do with the units?

tylersticka: I'm honestly not exactly sure, the math I stole from CSS-Tricks articles. I've been considering whether long-term we should replace this logic with something simpler based on math functions like `min` and `max`, which I didn't realize until recently are [startlingly well-supported](https://caniuse.com/css-math-functions).

Can you add a comment explaining why it's set to `9` here?

tylersticka (commented):
This feels so much better. Will approve once we add a Sass comment so future generations understand why we used `ms.step(9)` (always good to explain magic numbers, y'know?), but no issues at all with the end result.

tylersticka (approved):

tylersticka (approved):

PR #1530, opened 2021-09-09
Use aria-labelledby to add accessible names to articles

Paul-Hebert:
## Overview

Recently @gerardo-rodriguez pointed out:

> While testing a PR with VoiceOver, I noticed that the rotor lists all comments as "article". It seems we might be able to enhance the experience by adding accessible names to the <article> elements? (e.g. aria-labelledby)

article

This PR adds `aria-labelledby` to link the existing comment headings to the comment article.

You're gonna wanna hide whitespace changes while reviewing. There was an indentation issue in the file that was fixed as well.

## Screenshots

comment names

## Testing

1. Review  `/?path=/story/components-comment--with-reply-thread` on the preview deploy in VoiceOver. Check the rotor and note the articles have accessible names.

### DIFF
--- a/.changeset/rude-pears-breathe.md
+++ b/.changeset/rude-pears-breathe.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Add accessible names to comments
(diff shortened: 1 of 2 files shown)


gerardo-rodriguez (approved):
🎉

Issue #1521, opened 2021-09-03
Refine blog post block quote styling


Paul-Hebert:
We've had two passes over the Gutenberg quote block styles, but they still don't look quite right: https://v-next--cloudfour-patterns.netlify.app/?path=/docs/vendor-wordpress-core-blocks--quote#pullquote

We should refine these styles to work better in blog posts (and comments! Though maybe those are two different styles? I'm not sure. What do you think?)

We should also consider where those CSS styles are applied. There are a few places we may want to apply this?

- The [Gutenberg block](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/vendor-wordpress-core-blocks--quote#pullquote)
- block quotes within Wordpress markdown content. (something like this):
  ```
  .wp-block-jetpack-markdown blockquote {
    /* blockquote styles */
  }
  ```
- Within Wordpress comments? I'm not sure of the best way to handle this... should our `.c-comment__content` class style its children? Should we make a new "CMS Content" class we can apply wherever we expect CMS content? 

Note: we do not want to style `blockquote` directly since that's used other places like testimonials

The primary task here is to refine the visual styles. If whoever takes on the task is confident in applying these styles in the correct places go for it! Alternately, feel free to reach out to team mates for advice, or close out this card and make a new card for applying those styles.

Paul-Hebert:
The styles we applied to block quotes in CMS content in the old theme may be a good starting point: https://cloudfour-patterns.netlify.app/patterns/components/cms-content.html

PR #1551, opened 2021-09-23
Add Quotes Prototype

dromo77:
## Overview

This PR adds a prototype with pull quote and block quote styles.

## Screenshots

Screen Shot 2021-09-23 at 10 50 11 AM

Screen Shot 2021-09-23 at 10 50 35 AM

## Testing
1. [Deploy Preview](https://deploy-preview-1551--cloudfour-patterns.netlify.app/?path=/story/prototypes-quotes--example)

---

-  #1521


Paul-Hebert:
@dromo77 I wonder if we could remove the italics from block quotes as well so we don't lose the semantic meaning of italics in the text we quote? What do you think?

tylersticka:
> I wonder if we could remove the italics from block quotes as well so we don't lose the semantic meaning of italics in the text we quote? What do you think?

I agree, I don't love the italics.

dromo77:
@Paul-Hebert @tylersticka Sure thing
Screen Shot 2021-09-23 at 12 34 12 PM
!

Paul-Hebert:
This looks good to me, but I'll defer final design approval to @tylersticka

tylersticka (approved):

PR #1553, opened 2021-09-23
Remove duplicate CI step from test script

Paul-Hebert:
## Overview

(The build script runs preprocess, so running preprocess manually was redundant)

## Testing

I believe we just need to confirm the CI passes.


Paul-Hebert:
Huh, maybe the preprocess is required... weird! I'll dig into this more after lunch

Paul-Hebert:
Okay, I don't get why preprocess is required, but evidently it is 🤷

PR #1556, opened 2021-09-30
Update quote styles

### LINKED ISSUE #1521: Refine blog post block quote styling

dromo77:
## Overview

This PR updates the pull quote and block quotes styles. It uses the Gutenberg block classes to apply styles so that the `blockquote` element remains fairly unopinionated. It also applies those same styles to all `blockquote` elements within the `c-comment` component.

## Screenshots

Screen Shot 2021-09-30 at 11 46 26 AM

## Testing

1. [Deploy preview](https://deploy-preview-1556--cloudfour-patterns.netlify.app/?path=/docs/vendor-wordpress-core-blocks--pullquote)

---

- Fixes #1521


src/components/comment/comment.scss
```diff
@@ -83,6 +83,23 @@
...
+  color: color.$base-gray-dark;
+}
+
+.c-comment__content blockquote cite {
+  font-style: normal;
+}
+
+.c-comment__content blockquote cite::before {
+  content: '\2014';
+}
```
tylersticka, line 101: Since these styles are repeated here and in our WordPress vendor styles, should we create a Sass mixin and/or design tokens for them?

src/components/comment/comment.scss
```diff
@@ -83,6 +83,23 @@
   margin-top: math.div(size.$rhythm-condensed, -4);
 }
 
+// We added default styles to the blockquote element using the `wp-block-quote`
+// class. We're applying those same styles here, so comment blockquotes look
+// the similar.
+
+.c-comment__content blockquote {
+  border-left: size.$edge-medium solid color.$base-gray-light;
+  color: color.$base-gray-dark;
```
tylersticka, line 92: We should include styles for our dark theme as well.

src/vendor/wordpress/core-blocks.stories.mdx
```diff
@@ -466,10 +466,12 @@ applied only if the solid color background is toggled on.
...
 <Canvas>
   <Story name="Pullquote">
-    {`<figure class="wp-block-pullquote has-background is-style-solid-color" style="background-color:#dcd7ca">
+    {`<figure class="wp-block-pullquote has-background is-style-solid-color">
         <blockquote class="has-text-color has-accent-color">
           <p>One of the hardest things to do in technology is disrupt yourself.</p>
-          <cite>Matt Mullenweg</cite>
+          <footer>
+            <cite>Matt Mullenweg</cite>
+          </footer>
```
tylersticka, line 474: Did you verify that WordPress includes the `footer` element? If not, then we probably can't modify this markup.

dromo77: I didn't, I'll double check that.

src/components/comment/comment.scss
dromo77: That's a good point, I think tokens would work well here.

dromo77: @tylersticka I added tokens for most of these styles. However, I'm having trouble using a token for `font-style: normal`. I added a token in `tokens/font/styles.json`, but I get an error when I try to use it.

src/components/comment/comment.scss
```diff
@@ -2,9 +2,24 @@
...
+
+@include theme.props() {
+  --theme-color-border-blockquote: #{color.$border-blockquote-light};
+  --theme-color-text-blockquote: #{color.$text-blockquote-light};
+}
+
+@include theme.props(dark) {
+  --theme-color-border-blockquote: #{color.$border-blockquote-dark};
+  --theme-color-text-blockquote: #{color.$text-blockquote-dark};
+}
```
tylersticka, line 21: Reading this now, I think for the text you might be able to use `--theme-color-text-muted` instead of defining new properties and tokens? See `src/base/_themes.scss`

I could see adding a new border property in the base themes as well, esp. since we're wanting to coordinate that color with comment threads.

src/components/comment/comment.scss
```diff
@@ -83,6 +98,24 @@
...
+    var(--theme-color-border-blockquote);
+  color: var(--theme-color-text-blockquote);
+}
+
+.c-comment__content blockquote cite {
+  font-style: normal;
+}
+
+.c-comment__content blockquote cite::before {
+  @include emdash.content;
```
tylersticka, line 116: Just curious if we could simplify by using an em dash directly instead of needing the `\` code?

```
content: '—';
```

src/components/comment/comment.scss
dromo77: Cool! I didn't know we had theme tokens. I updated the color and border color properties so that we're using existing tokens.

src/components/comment/comment.scss
```diff
@@ -25,6 +26,7 @@
 /// those comments visually with their parent.
 .c-comment.is-replying::after,
 .c-comment--thread::after {
+  background-color: var(--theme-color-border-kbd);
```
tylersticka, line 29: Can we create a border property specifically for this? `kbd` refers to the `kbd` element, so I'm not sure it's the most intuitive thing to use here. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd

If it happens to be the right color, maybe we just need to rename it wherever it's referenced to something more general.

dromo77: > I'm not sure it's the most intuitive thing to use here

Haha probably not 😆. I added a property specifically for the border color.

tylersticka (approved):
Great work, @dromo77! Thanks for weathering the rounds of feedback.

Issue #1547, opened 2021-09-16
Offline variation of alert pattern


tylersticka:
Our live site includes an alert when the user has gone offline. We do this so it won't be a surprise if they attempt to interact with something that still requires connectivity, such as a contact form or an embedded element (video, CodePen, etc.)

![Screen Shot 2021-09-16 at 08 49 06](https://user-images.githubusercontent.com/69633/133644256-5adf1178-7322-4b1c-9d5a-1220e780f584.png)

An updated Alert pattern was designed, but we had some difficulty landing on an accessible "offline" variation. You can see the default variation here: https://v-next--cloudfour-patterns.netlify.app/?path=/docs/components-alert--full-width#alert

This task involves designing a modifier class that will style the alert well for notifying the user they are offline. It should be prominent but not alarming (since the user didn't do anything wrong), and it should look nice against the blue of our Cloud Cover pattern. And unlike the current offline bar, it should pass our accessibility guidelines.

It's also okay to explore additions for this scenario, such as an icon, but that's not strictly necessary.

AriannaChau:
I'm assigning myself to this because I'm interested in designing it. It's not my next task and if I dev wants to also assign themselves and we could possibly go back and forth, please feel free!

tylersticka:
Updated assignment so @dromo77 can pick up where Arianna left off. 🙂

PR #1557, opened 2021-10-04
Add new alert modifiers and new alert variation prototype

AriannaChau:
## Overview

This PR adds two new modifiers for the alert pattern: Adding an icon and adding a floating style.
There's also an alert added to a prototype to get a feel for the intended use of the offline alert.

A couple of other issues can come from this:

- Add more icon background colors
- Adding white background as its own modifier?
- Add positioning for specific offline alert
- Create a new offline/wifi icon

### Icon modifier: 
Screen Shot 2021-10-04 at 3 00 51 PM

### Floating modifier:
Screen Shot 2021-10-04 at 3 01 01 PM

### Alert in practice: 
Screen Shot 2021-10-04 at 3 00 11 PM

## Testing

1. [Deploy link (icon)](https://deploy-preview-1557--cloudfour-patterns.netlify.app/?path=/docs/components-alert--basic#icon)
2. [Deploy link (floating)](https://deploy-preview-1557--cloudfour-patterns.netlify.app/?path=/docs/components-alert--basic#floating)

---

- Fixes https://github.com/cloudfour/cloudfour.com-patterns/issues/1547


.changeset/two-forks-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add 2 new modifiers for the alert pattern
```
tylersticka, line 5: Let's be more specific about this message...

```suggestion
Add optional icon and floating modifier to the Alert component
```

src/components/alert/alert.scss
```diff
@@ -89,3 +89,24 @@
...
+.c-alert__icon {
+  align-items: center;
+  background-color: lighten(color.$base-orange, 30%);
+  border-bottom-left-radius: size.$border-radius-medium;
+  border-top-left-radius: size.$border-radius-medium;
+  color: color.$base-orange;
+  display: flex;
+  grid-area: icon;
+  justify-content: center;
+  width: ms.step(5);
```
tylersticka, line 102: We should add a comment explaining how we came to this size since it's sort of a magic number. For example, if this was chosen based on the usual single-line height of the pattern, that's a good thing to make note of.

src/components/alert/alert.scss
```diff
@@ -89,3 +89,24 @@
     transform: scale(scale.$effect-shrink);
   }
 }
+
+.c-alert__icon {
+  align-items: center;
+  background-color: lighten(color.$base-orange, 30%);
+  border-bottom-left-radius: size.$border-radius-medium;
+  border-top-left-radius: size.$border-radius-medium;
+  color: color.$base-orange;
```
tylersticka, line 98: Should we include colors for when this is contained in a dark theme?

src/components/alert/alert.scss
```diff
@@ -89,3 +89,24 @@
...
+  display: flex;
+  grid-area: icon;
+  justify-content: center;
+  width: ms.step(5);
+}
+
+// Adds a box shadow to add the appearance of the alert floating
+// above the containers below. Floating has a white background for higher
+// contrast between the background and the shadow.
+.c-alert__floating {
```
tylersticka, line 108: This class name is using the element notation (`__`) of the modifier notation (`--`). I think it's meant to be a modifier, in which case the class should be renamed and this style should be moved closer to `c-alert` so its closer to the thing it's modifying.

src/components/alert/alert.scss
```diff
@@ -89,3 +89,24 @@
...
+  background-color: white;
+  box-shadow: 0 0 0 2px hsla(224, 65%, 16%, 0.2),
+    4px 4px 0 2px hsla(224, 65%, 16%, 0.1);
```
tylersticka, line 111: Where did these color values come from? If they're based on an existing color value, we should reference that.

The sizes should also be based on tokens.

src/components/alert/alert.stories.mdx
```diff
@@ -68,11 +68,47 @@ By default, alerts assume they'll display within page content. For alerts meant
   </Story>
 </Canvas>
 
+## Icon
+
+An option to add an icon to the alert. You can add any icon name from our library in the `icon_name` property. Right now, the color is hardcoded to orange, but it would be nice to add more color options in the future.
```
tylersticka, line 73: We should mention _why_ it's hard-coded to orange. (I believe the reason is "to satisfy the current use case of our offline notification"?)

src/components/alert/alert.stories.mdx
```diff
@@ -68,11 +68,47 @@ By default, alerts assume they'll display within page content. For alerts meant
...
+      icon_name: 'check',
+    }}
+  >
+    {(args) => template(args)}
+  </Story>
+</Canvas>
+
+## Floating
+
+By adding `floating: true`, a light border and shadow will be added to the alert for a floating effect. The background is white to improve the contrast between the border, shadow and background. In the future, we could make it a possibilty to change the background color by itself.
```
tylersticka, line 89: Typo:

```suggestion
By adding `floating: true`, a light border and shadow will be added to the alert for a floating effect. The background is white to improve the contrast between the border, shadow and background. In the future, we could make it a possibility to change the background color by itself.
```

src/components/alert/alert.twig
```diff
@@ -1,7 +1,19 @@
 {% set tag_name = tag_name|default('div') %}
 
-<{{ tag_name }} class="c-alert{% if full_width %} c-alert--full-width{% endif %}{% if class %} {{ class }}{% endif %}">
+<{{ tag_name }} class="c-alert{% if full_width %} c-alert--full-width{% endif %} {% if floating %} c-alert__floating {% endif %}{% if class %} {{ class }}{% endif %}">
```
tylersticka, line 3: Nit: Extra space and wrong separator for modifier (see earlier comments on class name)

```suggestion
<{{ tag_name }} class="c-alert{% if full_width %} c-alert--full-width{% endif %}{% if floating %} c-alert--floating {% endif %}{% if class %} {{ class }}{% endif %}">
```

src/prototypes/single-page/example/example.scss
```diff
@@ -1,9 +1,46 @@
...
+      justify-content: center;
+      align-items: center;
+      @media (min-width: breakpoint.$l) {
+        width: 60px;
+      }
+    }
+
+    .text {
+      margin-left: 3em;
+    }
```
tylersticka, line 44: I would not expect these styles to still be necessary in this prototype if you're using the component.

tylersticka (commented):
This is looking great, and I think the list of future issues is spot on.

I left some comments inline. I also notice that when I apply the dark theme to the story, the text disappears.

You may want to consider theming the alert for light and dark containing themes, but then supporting adding a theme class (such as `t-light`) directly so you'll get the text, link and background colors where desired.

Screen Shot 2021-10-04 at 3 32 26 PM

dromo77:
@tylersticka I'm picking up where Arianna left off with this PR. It looks like she addressed almost all the feedback. I resolved a couple merge conflicts and addressed a small piece of feedback that looked like it hadn't been updated yet. Would you mind taking a look again?

src/components/alert/alert.twig
```diff
@@ -1,7 +1,19 @@
 {% set tag_name = tag_name|default('div') %}
 
-<{{ tag_name }} class="c-alert{% if class %} {{ class }}{% endif %}">
+<{{ tag_name }} class="c-alert{% if full_width %} c-alert--full-width{% endif %}{% if floating %} c-alert--floating{% endif %}{% if class %} {{ class }}{% endif %}">
   <div class="c-alert__inner">
+    {% if block('icon')|trim or icon_name %}
+      <span class="c-alert--icon">
```
tylersticka, line 6: This appears to be using modifier syntax (`.c-alert--icon`) instead of element syntax (`.c-alert__icon`).

Since this element is actually an icon _container_, we may want to call it `.c-alert__extra` for consistency with the Button component.

static/media/offline.svg
```diff
@@ -0,0 +1,19 @@
...
+                    <path d="M1.59465814,0 L0.531552715,0 C0.237952904,0 0,0.237952904 0,0.531552715 L0,1.59465814 C0,1.88825796 0.237952904,2.12621086 0.531552715,2.12621086 L1.59465814,2.12621086 C9.52102272,2.12621086 16.0174551,8.62264328 16.0174551,16.5490079 L16.0174551,17.6121133 C16.0174551,17.9057131 16.255408,18.143666 16.5490079,18.143666 L17.6121133,18.143666 C17.9057131,18.143666 18.143666,17.9057131 18.143666,17.6121133 L18.143666,16.5490079 C18.143666,7.44838242 10.6952836,0 1.59465814,0 Z" id="Path"></path>
+                </g>
+                <path d="M15.4049994,31.5599989 C23.9129457,31.5599989 30.8099989,24.6629457 30.8099989,16.1549994 C30.8099989,7.64705318 23.9129457,0.75 15.4049994,0.75 C6.89705318,0.75 -2.13162821e-14,7.64705318 -2.13162821e-14,16.1549994 C-2.13162821e-14,24.6629457 6.89705318,31.5599989 15.4049994,31.5599989 Z M26.373692,6.49794837 L3.94869287,25.8029484" id="Combined-Shape" stroke="#f27041" stroke-width="1.94999993"></path>
+            </g>
+        </g>
+    </g>
+</svg>
```
tylersticka, line 19: We would ideally introduce a new "offline" icon as part of [our glyph set](https://v-next--cloudfour-patterns.netlify.app/?path=/story/design-icons--page) instead of adding an icon to the `static` directory.

This should match the weight, appearance, etc. of our other icons and be included in the Illustrator file in our shared Drive for consistency.

src/components/alert/alert.stories.mdx
```diff
@@ -66,10 +66,67 @@ Alerts work well even when displayed at full width. Their contents will constrai
...
+    {(args) => template(args)}
+  </Story>
+</Canvas>
+
 ## Template Properties
 
 - `class` (string)
+- `icon_name` (string)
 - `dismissable` (boolean, default `false`)
+- `full_width` (boolean, default `false`)
```
tylersticka, line 128: `full_width` is no longer supported, see #1559

src/components/alert/alert.stories.mdx
```diff
@@ -66,10 +66,67 @@ Alerts work well even when displayed at full width. Their contents will constrai
...
+    }}
+  >
+    {(args) => template(args)}
+  </Story>
+</Canvas>
+
 ## Template Properties
 
 - `class` (string)
+- `icon_name` (string)
```
tylersticka, line 126: Any reason not to call this simply `icon`? (Property names and block names don't conflict.)

tylersticka (commented):
Something looks amiss when the alert is full-width at very wide sizes: https://deploy-preview-1557--cloudfour-patterns.netlify.app/?path=/story/components-alert--icon

Screen Shot 2021-11-15 at 3 42 35 PM

dromo77:
> Something looks amiss when the alert is full-width at very wide sizes

@tylersticka `c-alert__inner` has a max-width equal to our `size.$max-width-spread` token. When the alert only has text, the alert content lines up with the page content. But it's funky when there's an icon. I'm not sure what the best solution. We could remove the max-width with a modifier when there is an icon present. The downside of this is that some alert messages will be constrained and others won't. We could remove the max-width altogether, but that would allow alert content to span the full width of its container. We might also consider changing the style of the icon a bit on very large screens so that it works with the max-width. This is a quick example, but maybe something like this?

Screen Shot 2021-11-16 at 11 59 18 AM

tylersticka:
> This is a quick example, but maybe something like this?

@dromo77 I like that idea, it feels the most flexible and responsive. It might look even more purposeful and "on-brand" if the container became a circle past that max-width size?

src/prototypes/single-page/example/example.twig
```diff
@@ -39,6 +39,21 @@
   {% embed '@cloudfour/objects/container/container.twig' with {
     class: 'o-container--prose o-container--pad main-container'} %}
     {% block content %}
+
+
+      {% embed '@cloudfour/components/alert/alert.twig' with {
+        "dismissable": true,
```
tylersticka, line 45: Nit: One property name in quotes

src/components/alert/alert.twig
```diff
@@ -1,7 +1,19 @@
...
+        {% block icon %}
+          {% if icon %}
+            {% include '@cloudfour/components/icon/icon.twig' with {
+              name: icon,
+              aria_hidden: true
+            } only %}
+          {% endif %}
+        {% endblock %}
+      </span>
+    {% endif %}
```
tylersticka, line 16: In a separate PR I've been finding a way to simplify checks like this from what's being done in patterns like our button.

Earlier in the template, we can define our block but capture its output using `set`:

```twig
{% set _icon_block %}
  {% block icon %}
    {% if icon %}
      {% include '@cloudfour/components/icon/icon.twig' with {
        name: icon,
        aria_hidden: true
      } only %}
    {% endif %}
  {% endblock %}
{% endset %}
```

Then later in the template we can do this:

```twig
{% if _icon_block|trim is not empty %}
  <span class="c-alert__extra">
    {{_icon_block}}
  </span>
{% endif %}
```

Slightly less code, and to my eye makes the template _slightly_ easier to understand. But, that may be a matter of opinion... push back if you disagree!

src/assets/icons/offline.svg
```diff
@@ -0,0 +1,14 @@
...
+    c0.2,0,0.3-0.1,0.4-0.2l0.5-0.5c1.5-1.5,4.1-1.5,5.6,0l0.5,0.5c0.2,0.2,0.6,0.2,0.9,0l0.5-0.5c0.1-0.1,0.2-0.3,0.2-0.4
+    C16.9,15.1,16.9,14.9,16.8,14.8z"/>
+  <path d="M18.9,12.1l-0.5-0.5c-3.5-3.5-9.2-3.5-12.7,0l-0.5,0.5c-0.1,0.1-0.2,0.3-0.2,0.4c0,0.2,0.1,0.3,0.2,0.4l0.5,0.5
+    c0.2,0.2,0.7,0.2,0.9,0L7.1,13c1.3-1.3,3.1-2,4.9-2s3.6,0.7,4.9,2l0.5,0.5c0.1,0.1,0.3,0.2,0.4,0.2c0,0,0,0,0,0
+    c0.2,0,0.3-0.1,0.4-0.2l0.5-0.5C19.1,12.8,19.1,12.4,18.9,12.1z"/>
+  <path d="M21,9.5L20.5,9c-4.7-4.7-12.3-4.7-17,0L3,9.5C2.8,9.6,2.8,9.8,2.8,9.9c0,0.2,0.1,0.3,0.2,0.4l0.5,0.5c0.2,0.2,0.6,0.2,0.9,0
+    l0.5-0.5c3.9-3.9,10.3-3.9,14.2,0l0.5,0.5c0.1,0.1,0.3,0.2,0.4,0.2s0.3-0.1,0.4-0.2l0.5-0.5C21.3,10.1,21.3,9.7,21,9.5z"/>
+  <path d="M12,23c6.1,0,11-4.9,11-11S18.1,1,12,1S1,5.9,1,12S5.9,23,12,23z M19.8,5.1l-16,13.7" fill="none" stroke-width="2"/>
+</svg>
+
```
tylersticka, line 14: This icon feels like it could be more consistent with other elements of our patterns.

I think this is especially clear if you compare it to the way our `bell` icon looks inside of a slashed button:

Screen Shot 2021-11-16 at 2 10 06 PM
Screen Shot 2021-11-16 at 2 10 18 PM

When searching the Noun Project for "offline," I notice that a common visual is a cloud with a slash through it. Have we missed an opportunity to use a cloud where it makes symbolic sense? 🤔

tylersticka (commented):
The responsiveness feels great now! Some more feedback left inline, reach out if you have questions or aren't sure where to go next.

src/assets/icons/offline.svg
dromo77: I like the idea of using a cloud. What do you think of this @tylersticka?

Screen Shot 2021-11-17 at 8 47 20 AM

tylersticka: I like that better. Is it possible to make the cloud larger in relation to the slash?

dromo77: It's subtle, but here's an update version with the cloud a little larger in proportion to the slash.

Screen Shot 2021-11-17 at 9 42 12 AM

tylersticka: I like that better!

Can the icon itself be larger within its container? I think I'm reacting to its size relative to the capital "Y" in "You"

dromo77: Here it is a little bigger:

Screen Shot 2021-11-17 at 10 25 03 AM

tylersticka: I like that! How do you feel about it?

dromo77: I like it too!

tylersticka (commented):
Something I just noticed: In general, we want to name our icons based on what they are, not what they're for. So instead of `offline`, we might name it `cloud-slash`.

That might seem silly, but in other projects we've run into issues where we name an icon after one feature, it ends up being used for something else, and then later we stop using it for the original feature, so it's sort of in limbo. Better to just name it what it is.

Otherwise, this is great!

tylersticka (approved):

PR #1561, opened 2021-10-15
Add No Results prototype

### LINKED ISSUE #1543: Improving the "No Posts" experience
We added a search feature to the new version of the site. For some search terms there are no results. Right now it just says "No Posts Found." Is there a better experience we could provide?

dromo77:
## Overview

This PR adds a new prototype for the search results page when there are no results.

## Screenshots

Screen Shot 2021-10-15 at 12 37 58 PM

## Testing
1. https://deploy-preview-1561--cloudfour-patterns.netlify.app/?path=/story/prototypes-no-results--example

---

- Fixes #1543

### DIFF
--- a/src/prototypes/no-results/example/example.scss
+++ b/src/prototypes/no-results/example/example.scss
@@ -1,4 +1,12 @@
+@use "../../../compiled/tokens/scss/color";
+@use "../../../mixins/ms";
+
 #no-results {
+  .large-icon {
+    color: #{color.$base-gray-light};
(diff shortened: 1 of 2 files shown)


dromo77:
@tylersticka I see what you mean about the hierarchy, I struggled to size the heading. I'll switch gears and explore a solution with a CTA instead.

tylersticka:
@dromo77 I like it, but it might benefit from an image or icon as well?

Issue #1520, opened 2021-09-03
Style <hr>s


Paul-Hebert:
Right now our `<hr>` elements are using default styles:

![image](https://user-images.githubusercontent.com/5798536/132068118-16e5fb8c-9de7-4efc-b9aa-67824de77c4d.png)

We should make these fit our brand better and use our brand colors. I recently styled a divider for Footnote patterns which might be a useful starting point for colors/thickness (though I'm open to other ideas!)

![image](https://user-images.githubusercontent.com/5798536/132068164-095a3df4-4c2f-4eef-a300-ac1ebbc45635.png)

Paul-Hebert:
@tylersticka it seems like could just style `hr` directly? Is there any reason that's a bad idea and this should be constrained to a CMS Content class or something like that?

tylersticka:
@Paul-Hebert Styling directly sounds fine!

Paul-Hebert:
It looks like we did take a pass at styling the Gutenberg divider: https://v-next--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-core-blocks--separator

@tylersticka should we standardize around this design? Or give it another pass (and apply those styles to the Gutenberg block)?

tylersticka:
@Paul-Hebert I think it could use another pass.

PR #1569, opened 2021-11-05
Misc. Storybook chores

tylersticka:
## Overview

I started on #1520 and realized there wasn't a great place for default elements. I saw a comment on our figure example mentioning this issue, so I decided to convert our "Typography" section into a "Defaults" section. But in the process of moving those things, I saw a deluge of deprecation warnings in the console. So I decided before I add anything new, I would just go ahead and make those changes to set myself up for success.

Summary of changes:

- The `Design/Typography` stories have been replaced with `Design/Defaults`. I think this is more in the spirit of what the page was documenting, and it no longer feels awkward having other default elements documented there.
    - I was finding the process of updating the demos pretty arduous, so I moved them out of Twig templates and into strings. I think this also keeps us more honest, since if any of the demos relied on a lot of Twig stuff they are probably _not_ examples of base elements. This actually helped me identify an issue where the code block wasn't _really_ demonstrating the default element, for example.
- Doc blocks are now imported directly from `addon-docs` instead of `addon-docs/blocks`. ([More info](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-scoped-blocks-imports))
- The `defaultValue` settings for `argTypes` have been moved to default arguments. ([More info](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#no-longer-inferring-default-values-of-args))
- `argTypes` options that were defined inside of a `control` object have been moved out to the top of their argument definition. ([More info](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-controloptions))
- I noticed that any `argTypes` with an `enum` type seemed to not be working: They would always show an empty text field. Replacing this with `string` seemed to fix this issue.

## Screenshots

Screenshot 2021-11-05 at 12-23-35 Storybook


src/components/calendar-date/calendar-date.stories.mdx
```diff
@@ -19,10 +18,15 @@ Formats a `<time>` element so that it looks like a calendar icon. Intended to ac
...
-  <Story name="Basic" argTypes={argTypes}>
+  <Story name="Basic" args={{ datetime: Date.now() }} argTypes={argTypes}>
     {({ datetime, note }) => template({ datetime: new Date(datetime), note })}
   </Story>
-  <Story name="With Note" argTypes={argTypes} args={{ note: '3-day event' }}>
+  <Story
+    name="With Note"
+    args={{ datetime: Date.now() }}
+    argTypes={argTypes}
+    args={{ note: '3-day event' }}
```
Paul-Hebert, line 28: Looks like args is declared twice here?

Paul-Hebert (approved):
Looks good, thanks @tylersticka ! One minor nit inline.

Paul-Hebert (approved):

PR #1575, opened 2021-11-08
Style horizontal rules and separators

### LINKED ISSUE #1520: Style <hr>s

tylersticka:
## Overview

We lacked a default style for `hr`. This changes that. It uses the top border of `footnote-group` for basic appearance, but takes more color cues from the left comment border so that it will have slightly better contrast (and better compatibility with secondary themes). (I resisted the urge to style the footnote group to match because that feels like a visual distinction that does not necessarily warrant greater contrast.)

It also extends those styles to the WordPress "Separator" block.

## Screenshots

Screen Shot 2021-11-08 at 10 32 54 AM
Screen Shot 2021-11-08 at 10 33 04 AM
Screen Shot 2021-11-08 at 10 33 18 AM
Screen Shot 2021-11-08 at 10 33 23 AM
Screen Shot 2021-11-08 at 10 33 30 AM

---

- Fixes #1520

### DIFF
--- a/src/base/_defaults.scss
+++ b/src/base/_defaults.scss
@@ -1,3 +1,4 @@
+@use 'sass:math';
 @use '../compiled/tokens/scss/color';
 @use '../compiled/tokens/scss/size';
 @use '../mixins/border-radius';
@@ -267,4 +268,18 @@ figcaption {
 
 address {
   font-style: normal;
+}
+
+/**
+ * Horizontal rules
+ *
+ * We use border styles and `currentColor` for greater compatibility with
+ * the WordPress Gutenberg "Separator" block.
+ */
+
+hr {
+  border-color: currentColor;
+  border-style: solid;
+  border-width: math.div(size.$edge-medium, 2) 0;
+  color: var(--theme-color-border-text-group);
 }
--- a/src/base/_defaults.scss
+++ b/src/base/_defaults.scss
@@ -275,11 +275,15 @@ address {
  *
  * We use border styles and `currentColor` for greater compatibility with
  * the WordPress Gutenberg "Separator" block.
+ *
+ * 1. Our goal is for the borders to add together to form the total desired
+ *    width. We do this because that's what WordPress's Gutenberg block does
+ *    in order to support weird border+background style combinations.
  */
 
 hr {
   border-color: currentColor;
   border-style: solid;
-  border-width: math.div(size.$edge-medium, 2) 0;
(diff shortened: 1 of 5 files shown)


src/base/_defaults.scss
```diff
@@ -268,3 +269,17 @@ figcaption {
...
+ * Horizontal rules
+ *
+ * We use border styles and `currentColor` for greater compatibility with
+ * the WordPress Gutenberg "Separator" block.
+ */
+
+hr {
+  border-color: currentColor;
+  border-style: solid;
+  border-width: math.div(size.$edge-medium, 2) 0;
```
spaceninja, line 283: I trust you have a good reason, so I'm still going to approve, but perhaps you could add a comment here explaining why this is set to 1/2 of `$edge-medium`, rather than just using `$edge-small`?

spaceninja (dismissed):

src/vendor/wordpress/styles/_core-blocks.scss
```diff
@@ -281,3 +281,16 @@ $wp-button-gap: size.$spacing-gap-button-group-default;
...
+/**
+ * Gutenberg block: Separator
+ *
+ * The default separator is intended to be partial-width, so we apply those
+ * styles selectively here rather than un-applying them to the other modifiers.
+ */
+
+.wp-block-separator.is-style-default {
+  margin-left: auto;
+  margin-right: auto;
```
spaceninja, line 294: Is our browser support good enough to use `margin-inline`?

src/base/_defaults.scss
tylersticka: @spaceninja I've pushed up an additional comment that I hope explains it better.

src/vendor/wordpress/styles/_core-blocks.scss
tylersticka: @spaceninja Good question. It looks like we currently have 104 occurrences of traditional margin properties in the project already, so I think I'll create a separate issue about investigating this. Maybe also look into a linting error depending on where that lands?

tylersticka: @spaceninja Bam: #1576 1576

spaceninja (approved):

PR #1577, opened 2021-11-09
Add heading_id option to card template

### LINKED ISSUE #1539: Cards should support `aria-labelledby`
Right now cards use the `article` tag by default. These will show up in the "Articles" section of the rotor. However, since they're not specifically labeled they'll all be labeled 'article'.

Cards have headings we could use for labels. Should we set up support for `aria-labelledby`? We wouldn't necessarily want this to be used if we were using a tag other than article. Maybe we just add the attributes to support it and component consumers can choose whether to set it up?

articles

tylersticka:
## Overview

@Paul-Hebert made a good suggestion to add support for the `aria-labelledby` property on articles to improve the experience of navigating between them via the VoiceOver rotor. After some research in [the web-a11y Slack](https://www.tpgi.com/anybody-can-be-an-a11y-slacker/), I couldn't see any downside to adding this as an option.

I originally implemented the `heading_id` and a separate `aria_labelledby` property, but this felt a bit silly since they should always be the same. Instead, I simplified it down to a single `heading_id` option. If the `heading_id` is specified and the `heading` block isn't empty, it's automatically output to `aria-labelledby` on the root element as well. I wasn't sure if the label had to be on the _heading_ or the immediate container of the heading text (in which case the link would need it, if present), but thankfully after some testing it appears the heading is always the most appropriate element.

Along the way, I found myself struggling a bit to update all of the static code samples, so I also replaced those with a `templateSource` function that builds the examples automatically. This should make the documentation for cards easier to maintain going forward.

## Screenshots

Screen Shot 2021-11-09 at 1 19 23 PM

## Testing

1. On the deploy preview, try some of the Card examples in the VoiceOver rotor's "Articles" listing. You should see the article heading instead of a generic "Article."
2. Review the docs page to verify that the source examples do not contain any unexpected regressions compared to what's live.

---

- Fixes #1539

### DIFF
--- a/src/components/card/card.stories.mdx
+++ b/src/components/card/card.stories.mdx
@@ -1,16 +1,13 @@
 import { Story, Canvas, Meta } from '@storybook/addon-docs';
+import { uniqueId } from 'lodash';
 import singleDemo from './demo/single.twig';
-const singleDemoStory = (args) => {
+const singleDemoProps = (args) => {
   const props = {
+    heading_id: args.heading_id || uniqueId('card-heading-'),
     href: args.href,
   };
   const classNames = [];
   const modifiers = [];
-  if (args.show.length > 0) {
-    for (const block of args.show) {
-      props[`show_${block}`] = true;
-    }
-  }
   if (args.horizontal !== 'none') {
     modifiers.push(`horizontal${args.horizontal}`);
   }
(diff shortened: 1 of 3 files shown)


gerardo-rodriguez:
> 2\. Review the docs page to verify that the source examples do not contain any unexpected regressions compared to what's live.

@tylersticka Can you help me understand what you are referring to when you say "live"? Thank you! 🙂

tylersticka:
> Can you help me understand what you are referring to when you say "live"? Thank you! 🙂

What's already been merged: https://v-next--cloudfour-patterns.netlify.app/

spaceninja (dismissed):
Code looks reasonable, and I can confirm that in the VO Rotor under "Articles" the generic "article" has been replaced with the title of the article. I defer to @gerardo-rodriguez for full approval on the a11y improvement, but this PR does what it says on the tin.

src/components/card/card.stories.mdx
```diff
@@ -65,20 +99,13 @@ const singleDemoStory = (args) => {
 
 A card contains content and optionally an action summarizing a single piece of content, for example an article, talk or case study.
 
+If your card represents an `article` and includes a heading, it's a good idea to provide a `heading_id` as well. This allows us to use `aria-labelledby` to improve the experience of navigating to individual cards via the [VoiceOver rotor](https://support.apple.com/en-us/HT204783) (and potentially other assistive interfaces as well).
```
gerardo-rodriguez, line 102: Thanks for including a link, @tylersticka! 

I hadn't actually tried accessing the "articles" rotor menu on an iOS device and I was surprised in two ways:

1. The "articles" menu is not on by default on iOS VoiceOver (it _is_ on by default on macOS VoiceOver)
2. When I navigated by articles via iOS VoiceOver on our live [site](https://cloudfour.com/thinks/), it _does_ read out the heading already. 🤯 This experience is _different_ from macOS VoiceOver. macOS VoiceOver _does not_ read the heading and instead only lists "article" as the [original issue](https://github.com/cloudfour/cloudfour.com-patterns/issues/1539) described.

This made me wonder if the linked article didn't quite make sense since the experience was already good on iOS VoiceOver.

I thought, "I'll find a macOS VoiceOver 'articles menu' rotor link to suggest instead!" but I can't find one. 🤔 

That's a lot of words, I ended with these questions:
1. This is macOS VoiceOver rotor-specific, so maybe we mention that?
2. Can we find a macOS VO-specific article, otherwise, don't link to anything?

Let me know if I didn't make sense. 🙃

src/components/card/card.stories.mdx
```diff
@@ -257,18 +198,19 @@ When the card is a focal point and more contrast is desired, you may also attach
 ## Template Properties
 
 - `class` (string): Append a class to the root element.
-- `tag_name` (string, default `'article'`): The root element tag.
-- `header_tag_name` (string, default `'header'`): The header element tag.
-- `footer_tag_name` (string, default `'footer'`): The footer element tag.
+- `heading_id` (string): Add this `id` to the heading element (if present), as well as the `aria-labelledby` property of the card element (unless `aria_labelledby` is already specified).
```
gerardo-rodriguez, line 201: I had to read this a couple of times for it to register in my brain. 😅 

Maybe modifying it a bit?

```suggestion
- `heading_id` (string): Adds this `id` to the heading element (if a heading element is present); this also adds the `aria-labelledby` property of the card element (unless `aria_labelledby` is already specified).
```

src/components/card/card.stories.mdx
tylersticka: @gerardo-rodriguez Is this better? 26c1e7c

gerardo-rodriguez: That works great, @tylersticka, thanks! 😄

gerardo-rodriguez (approved):
Great work, @tylersticka! 🎉

PR #1578, opened 2021-11-09
Simplify fluid values using clamp

### LINKED ISSUE #1560: Simplify fluid calculations
We have [a number of fluid mixins and functions](https://github.com/cloudfour/cloudfour.com-patterns/blob/v-next/src/mixins/_fluid.scss) designed to gradually transition from a minimum to maximum size based on start and end breakpoints. This is based on some rather complicated math, covered in depth here:

- https://blog.typekit.com/2016/08/17/flexible-typography-with-css-locks/
- https://betterwebtype.com/articles/2019/05/14/the-state-of-fluid-web-typography/
- https://github.com/twbs/rfs

I suspect that these styles could be simplified using [a CSS `clamp()` function](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp()), which is [well supported now](https://caniuse.com/css-math-functions). But it's possible I'm misunderstanding the scope of that function, in which case this issue can be closed.

tylersticka:
## Overview

Replaces our property-specific "fluid" mixins with a more versatile `fluid-clamp` function based on an example by @Paul-Hebert. The output is greatly simplified by replacing media queries with [the `clamp` function](https://caniuse.com/css-math-functions).

I marked this as a breaking change since it slightly reduces browser support, it removes mixins that were previous available, and it changes the argument order of the remaining functions (what we had previously wasn't very intuitive). But I'm open to feedback if that feels too aggressive.

## Testing

Compare various components, objects and prototypes on the deploy preview and compare to the `v-next` version for regressions.

---

- Fixes #1560

### DIFF
--- a/.changeset/ninety-hotels-jump.md
+++ b/.changeset/ninety-hotels-jump.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': major
+---
+
+Use `clamp` instead of media queries to simplify fluid logic.
--- a/src/base/_typography.scss
+++ b/src/base/_typography.scss
@@ -16,11 +16,11 @@
  */
 
 html {
-  @include fluid.font-size(
-    breakpoint.$xs,
-    breakpoint.$xl,
+  font-size: fluid.fluid-clamp(
     ms.step(0, 1rem),
-    ms.step(1, 1rem)
+    ms.step(1, 1rem),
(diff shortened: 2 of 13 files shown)


spaceninja (approved):
Code looks good, and a quick review of the deploy environment didn't show any obvious regressions.

PR #1579, opened 2021-11-11
Add Media Group object

### LINKED ISSUE #1390: Column content alignment pattern
For the [case study listings page](https://v-next--cloudfour-patterns.netlify.app/iframe.html?id=prototypes-case-study-listings--example&args=&viewMode=story), there are some columns that are in different sections but are still aligned vertically, particularly the book is aligned with the icons below, and the text for each one is also aligned. I'm not exactly sure what the solution for this would be...feel free to reach out if you need more info.

Screen Shot 2021-07-12 at 10 19 05 AM

dromo77:
## Overview

This PR adds a Media Group object. It should be helpful for aligning images/content on pages like the [Case Study Listing](https://v-next--cloudfour-patterns.netlify.app/iframe.html?id=prototypes-case-study-listings--example&args=&viewMode=story) page.

## Screenshots

Screen Shot 2021-11-11 at 11 12 16 AM

## Testing

1.

---

- Fixes #1390


tylersticka:
@dromo77 Thanks for tackling this challenging layout problem.

I'm not sure I understand what `o-media-group__item` does that `o-media` does not, or what `o-media-group` does that `o-deck` does not. I assume I'm missing something between the way the two interact?

dromo77:
> I'm not sure I understand what `o-media-group__item` does that `o-media` does not, or what `o-media-group` does that `o-deck` does not. I assume I'm missing something between the way the two interact?

@tylersticka I ran into a couple issues trying to use `o-media` and `o-deck` as a solution. In the prototype the content splits into two columns at the `l` breakpoint.  `o-deck` automatically creates columns, so there were multiple columns before the `l` breakpoint. The content looked a little squished.

`o-media` doesn't size the object and also uses `flex`. I thought about some sort of modifier that uses CSS Grid. This would help size of the object and could also be used to align objects vertically, similar to what the `o-media-group__item` is doing. However, the alignment and grid size of the object are different on small screens versus large screens. I wasn't sure how to handle the different sizes/alignment needed. 

I'm open to other solutions, but wanted to give some context to the problems I ran into using existing patterns.

tylersticka:
> `o-deck` automatically creates columns, so there were multiple columns before the `l` breakpoint. The content looked a little squished.

I assume that was true even with the `o-deck--2-column@l`modifier. Perhaps we should add `1-column` modifiers to `o-deck`? Would that solve the issue?

> `o-media` doesn't size the object and also uses `flex`. I thought about some sort of modifier that uses CSS Grid. This would help size the object and could also be used to align objects vertically, similar to what the `o-media-group__item` is doing. However, the alignment and grid size of the object are different on small screens versus large screens. I wasn't sure how to handle the different sizes/alignment needed.

Reading the comment in the `media.scss` file, it seems like I used Flexbox to write it just because it was simple. But that's not a very compelling reason to _keep_ using Flexbox. If there's a good reason to use CSS Grid instead, we should totally do that.

From there, there are probably a few ways of handling those case study image widths. Perhaps we could make one or more `o-media` modifiers that enforce a certain object column width?

Or, honestly, I don't mind the different media columns being different as much as I used to. If we find that's taking an outsized amount of effort to componentize, maybe it isn't completely necessary after all? What do you think?

dromo77:
>  If we find that's taking an outsized amount of effort to componentize, maybe it isn't completely necessary after all?

I'll try using our existing patterns with the changes you mentioned. I don't think it will take too much time to add `o-media` modifiers, but if I find otherwise we can consider different sized columns.

dromo77:
@tylersticka I made several changes:

1. I updated the [Case Study Listing](https://deploy-preview-1579--cloudfour-patterns.netlify.app/?path=/story/prototypes-case-study-listings--example) prototype to use patterns so we can see what it looks like in action.
2. I updated the `o-deck` column modifier so you can specify when you want 1 column.
3. I updated `o-media` to use CSS grid when supported. I _think_ I did it correctly. I haven't noticed any regressions, but would appreciate another pair of eyes on it.
4.  I added a modifier to specify the size of `o-media__object` relative to `o-media__content`. I want to make sure the modifier name and documentation makes sense.
5.  I added another modifier to `o-media` to fix some alignment issues I noticed in the prototype.

.changeset/polite-gorillas-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Update Deck columns modifier. Update Media to use CSS Grid and add modifier to size content.
```
tylersticka, line 5: 1. Can we add these as two separate changelog updates?
2. Can we be more specific about what modifiers were added?
3. Can we add a bit of "why" information?

src/objects/media/media.scss
```diff
@@ -50,3 +52,49 @@
...
+@supports (display: grid) {
+  .o-media {
+    align-items: center;
+    display: grid;
+    grid-gap: size.$spacing-media-gap;
+    grid-template-columns: minmax(0, auto) 1fr;
+  }
+
+  @for $i from 1 through 4 {
+    .o-media--1by#{$i} {
```
tylersticka, line 69: I like this class name, but can we add hyphens so it's `--1-by-X` for consistency with our other class names?

src/objects/media/media.scss
```diff
@@ -50,3 +52,49 @@
...
+  /**
+   * Reset fallback styles, we use `grid-gap` when CSS Grid is supported
+   */
+  .o-media__object {
+    margin-right: 0;
+  }
+
+  .o-media--reverse .o-media__object {
+    margin-left: 0;
+  }
```
tylersticka, line 89: I don't think these are necessary since Grid is so well supported now.

src/objects/media/media.scss
```diff
@@ -50,3 +52,49 @@
 .o-media__content {
   flex: 1;
 }
+
+/**
+ * We use CSS grid when supported by the browser.
+ */
+
+@supports (display: grid) {
```
tylersticka, line 60: Since IE11 is the only browser that doesn't support Grid these days, I wonder if having a fallback is still necessary?

src/objects/media/media.scss
```diff
@@ -50,3 +52,49 @@
...
+  }
+}
+
+/**
+ *  Modifier: Align Start
+ *
+ * Items remain top aligned
+ */
+
+.o-media--alignStart {
```
tylersticka, line 98: For consistency with our BEMIT class names, this should be `--align-start`

src/objects/media/media.stories.mdx
```diff
@@ -93,6 +94,44 @@ The order of `o-media__object` and `o-media__content` will not affect this patte
   </Story>
 </Canvas>
 
+## Alignment
+
+By default, content will appear middle-aligned with the object. If the content exceeds the object's height, they will appear top-aligned. You can change the alignment so the content and object remain top aligned using the modifier `o-media--alignStart`.
```
tylersticka, line 99: It isn't very clear to me _why_ we would do this? 🤔

tylersticka (commented):
It looks like some of the prototype images have been copied from `src/prototypes/case-study-listings/images` to `static/media`. That's fine, but can we remove the duplicates so they are only in one place?

src/objects/media/media.stories.mdx
dromo77: When I updated the Case Study Listing prototype I noticed some of the media objects and content were top-aligned and some were middle-aligned. In a group like this, it looks off when the alignment is different.

Screen Shot 2021-11-12 at 1 18 14 PM

tylersticka: Ah, interesting. I wonder if that's better solved on the media object or on the deck by changing `align-items` or `align-content` via a modifier?

dromo77: Hmm, that's a good question. I don't have a strong opinion on whether we add a modifier to media or the deck. What do you think?

tylersticka: I think it makes more sense on `o-deck` for a couple of reasons:

- The `o-media` object isn't really causing this alignment issue on its own: It's because the Deck is causing it to stretch to fill the column.
- I think `o-deck--align-start` is easier to understand than `o-media--align-start`, because to most people I think they'd look at `o-media` and think "isn't that already aligned to the start?" Whereas if you were to have a deck of cards I think it would be easier to see the difference.
- It's a little nicer to put the modifier on one container instead of every child element.

tylersticka (commented):
Apparently I wrote a comment on Friday that I never actually posted. Sorry about that. 🤦

src/objects/deck/deck.stories.mdx
```diff
@@ -83,11 +84,25 @@ By default, the `o-deck` class will use CSS Grid Layout to arrange child element
...
+
+<Canvas>
+  <Story
+    name="Alignment"
+    height="400px"
+    args={{ class: 'o-deck--align-start' }}
+  >
+    {articlesStory.bind({})}
+  </Story>
+</Canvas>
```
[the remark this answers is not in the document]
dromo77: @tylersticka I updated the example, let me know if it makes more sense 🙂

tylersticka (approved):

PR #1586, opened 2021-11-16
Add `--icon-size` property, use within `c-button`

### LINKED ISSUE #1540: Add ability for search input's icon to use an increased font size
The [article listing prototype](https://v-next--cloudfour-patterns.netlify.app/?path=/story/prototypes-article-listing--example) includes a search form with an icon button. That icon has `1.25em` font size. I can't figure out how to do this with the existing components. I tried using the `c-button__extra` class on my button content but that also affected margins. 

Should we make a test utility? A button/icon modifier? Something else? Am I missing an existing feature that handles this?

Screen Shot 2021-09-13 at 12 38 23 PM

tylersticka:
## Overview

Addresses an issue where icons were inconsistently sized if included as the primary button content, as one might expect for an icon-only button.

Previously, button icon sizes were controlled via `font-size` on the icon container, `c-button__extra`. When an icon was included as the primary content, `c-button__content`, there was no such `font-size` declaration, and so the icon would appear smaller by comparison.

I considered adding some utility classes and/or icon modifiers to solve this problem, but I realized when futzing with the component that there were other potential issues with relying on `font-size` on a containing element. For example, it's impossible to reliably negate the gap between the `__extra` and `__content` elements since the `font-size` for the icon increases the computed size of that gap to a spot between two of our modular scale steps.

Instead, I added an optional `--icon-size` property to the `c-icon` component. This allows us to set the desired size of the icon on `c-button`, which cascades to _all_ elements below. It also means the computed impact of the icon's size is not effective until the icon itself, so any margins or other relative values on its container will be more consistent with adjacent elements.

Along the way, I noticed that the `c-button__extra` styles were being unnecessarily output a second time within the context of WordPress buttons. I separated those styles into their own mixins to prevent this.

(This PR originally included other changes to `c-button` to streamline its template code, but I realized partway in that the actual solution to the issue was a smaller subset of that, which I've separated into this PR.)

### Before

Screen Shot 2021-11-16 at 2 33 54 PM

### After

Screen Shot 2021-11-16 at 2 40 35 PM

## Testing

On the deploy preview…

1. Review Button component for regressions
2. Review Icon component for regressions
3. Review Article Listing prototype search form for regressions

---

- Fixes #1540

### DIFF
--- a/.changeset/new-pets-care.md
+++ b/.changeset/new-pets-care.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add `--icon-size` CSS custom property to the Icon component to allow sizing via CSS without impacting `font-size` or selector depth.
--- a/.changeset/odd-sloths-allow.md
+++ b/.changeset/odd-sloths-allow.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Remove extra `.c-button__extra` selectors from WordPress block button styles.
--- a/.changeset/tough-books-sing.md
+++ b/.changeset/tough-books-sing.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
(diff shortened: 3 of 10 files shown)


spaceninja (approved):
:shipit:

PR #1591, opened 2021-11-18
Add min-width to Page contents

### LINKED ISSUE #1532: Some blog posts are extending past their containers
@gerardo-rodriguez pointed this out when reviewing a Wordpress PR (https://github.com/cloudfour/cloudfour.com-wp/pull/609) but I think the root of the problem is in our pattern library CSS

Screen Shot 2021-09-09 at 3 48 19 PM

The issue seems to crop up when a post has wide code blocks. These code blocks are meant to max out at 100% width and scroll if they have horizontal overflow but that doesn't seem to be happening.

Weirdly, our `pre` elements _do_ have `overflow: auto` so it seems like they should work. I dug into this a bit and it seems like it may be related to our page object? If I remove the `o-page` class from the `body` or set `o-page__content` to `max-width: 100vw` the problem fixes itself.

It's not totally clear to me what the correct fix is. We should update the single article prototype to reproduce the issue, and then fix it in our pattern library CSS.

spaceninja:
## Overview

This PR fixes a bug where blog posts that contained lengthy code samples were breaking layout. The bug wasn't actually caused by the code samples, which have the proper `overflow:auto` property set. Instead, this is caused by an oddity of Grid layouts, where the Grid items by default have `min-width:auto`, which can prevent them from reducing width the way you might expect. The fix is to tell the Grid layout the minimum width for Grid items, and then it will shrink the way you'd expect.

Learn more: https://css-tricks.com/you-want-minmax10px-1fr-not-1fr/

## Testing

Unfortunately, the bug was not visible in the pattern library, because we weren't applying the Page object to the Single Article prototype. This PR applies that change so we can test the fix.

1. visit the [Single Article prototype](https://deploy-preview-1591--cloudfour-patterns.netlify.app/?path=/story/prototypes-single-article--example) on the deploy preview
2. Using devtools, disable the `grid-template-columns` rule on `.o-page`.
3. Reduce your browser to small screen size, and you should see the layout break, with a horizontal scrollbar on the entire page.
4. Restore the `grid-template-columns` rule.
5. The layout should now be fixes.

---

- Fixes #1532

### DIFF
--- a/src/objects/page/page.scss
+++ b/src/objects/page/page.scss
@@ -1,16 +1,17 @@
 /**
  * Page container
  *
- * Attempts to fill the available area (assuming `height: 100%` is correctly
- * applied to all parent elements) and keeps the footer at the bottom of that
- * area.
- *
- * @see https://css-tricks.com/couple-takes-sticky-footer/#there-is-grid
+ * 1. Avoid grid blowouts by specifying the `min-width` of the column.
+ *    @see https://css-tricks.com/you-want-minmax10px-1fr-not-1fr/
+ * 2. Attempts to fill the available area (assuming `height: 100%` is applied
+ *    to all parent elements) & keep the footer at the bottom of that area.
+ *    @see https://css-tricks.com/couple-takes-sticky-footer/#there-is-grid
  */
 
 .o-page {
   display: grid;
-  grid-template-rows: 1fr auto;
+  grid-template-columns: minmax(0, 1fr); // 1
+  grid-template-rows: 1fr auto; // 2
   min-height: 100%;
 }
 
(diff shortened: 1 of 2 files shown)


src/objects/page/page.scss
```diff
@@ -18,12 +18,19 @@
...
  */
 
 .o-page__content {
   grid-row: 1;
+  min-width: 0; // 1
 }
 
 .o-page__footer {
   grid-row: 2;
+  min-width: 0; // 1
```
tylersticka, line 35: Out of curiosity, would setting `grid-template-columns: minmax(0, 1fr);` on the `o-page` element have the same effect as setting `min-width: 0` on the child elements?

https://css-tricks.com/you-want-minmax10px-1fr-not-1fr/

gerardo-rodriguez: FWIW, I tried out @tylersticka's suggestion in the browser via DevTools and it seems to work. I'm am going to read the article to better understand why. 📖 🔍 👀

gerardo-rodriguez: Okay, I'm convinced in that we should explore `grid-template-columns: minmax(0, 1fr);` as @tylersticka suggested! 🙂 

This article helped explain a bit more for me (linked in the original shared article): https://css-tricks.com/preventing-a-grid-blowout/

spaceninja: Good call, thanks @tylersticka!

Updated.

gerardo-rodriguez (approved):
LGTM! 🎉

Issue #1576, opened 2021-11-08
Make use of logical properties consistently


tylersticka:
## Overview

Waaaaaay back in #651, we had a lot of internal discussion about [logical properties](https://www.w3.org/TR/css-logical-1/), ultimately deciding to [base our utilities on them](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/utilities-spacing--logical-properties). But since then, only our utilities and [our Container object](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/objects-container--basic) use logical properties: We appear to overwhelmingly be using traditional, directional property names like `margin-left` over equivalents like `margin-inline-start`. 

At the time, some of our misgivings may have been a factor of [confusing support](https://caniuse.com/css-logical-props). Chrome versions ≤ 86 and Safari versions ≤ 14.0 did not accurately support shorthand properties like `margin-inline` or `padding-block`. Thankfully, these issues have been resolved in more recent versions.

Another factor may be that we focused too little on this change in CSS compared to utility classes. Our spacing and padding utility classes have proven useful, with ~ 45 occurrences in our documentation and prototypes, but that's nothing compared to the 100+ occurrences of `margin` in our stylesheets.

Perhaps this is also a difficult change to make without some automated hand-holding. There are a few competing stylelint plugins for this, the most popular appearing a bit stale. This alternative seems a bit better maintained: https://www.npmjs.com/package/stylelint-use-logical-spec

## To-dos

- [x] Research and confirm there are no major "gotchas" preventing us from going all-in on logical properties
- [x] Set up linting to enforce use of logical properties in CSS
- [x] Update existing CSS rules so they comply with these new rules
- [x] Test for regressions

---

Thanks @spaceninja for pointing this out [during a PR review](https://github.com/cloudfour/cloudfour.com-patterns/pull/1575#discussion_r745134533)

spaceninja:
I was recently updating my stylelint config and considered adding https://github.com/Jordan-Hall/stylelint-use-logical-spec 

In the end I opted not to because it felt like a big change. But I agree that we should definitely handle this via linting when the time comes.

tylersticka:
We've had some internal discussion around whether or not it makes sense to embrace logical properties in this way or not. To summarize:

- [Most logical properties are well-supported at this point](https://caniuse.com/css-logical-props)
[...]

PR #1592, opened 2021-11-20
Improve themes toolbar, add text flow toolbar

tylersticka:
## Overview

While working on #1576, I realized I didn't have a great method for actually testing that logical properties were applying well across different `direction` and `writing-mode` settings. The Storybook add-ons related to this seem pretty far out of date, but then I saw that this was one of the use cases for [the global toolbar settings](https://storybook.js.org/docs/react/essentials/toolbars-and-globals) we're already using for themes. I found managing separate toolbars for `direction` and `writing-mode` a bit unwieldy, so instead I consolidated these into a "text flow" item.

Along the way, I noticed that our Theme toolbar did not include all of our themes. So I modified it to support `t-alternate`.

I set `showName` to `true` for both of our toolbar items to make them more obvious.

Once I made these changes, I noticed that the logical properties demo for our spacing utilities wasn't quite as useful as seeing how the different settings directly impacted stories. So I removed it.

## Screenshots

Screen Shot 2021-11-19 at 4 37 07 PM

Screen Shot 2021-11-19 at 4 37 20 PM


calebeby:
@tylersticka This is probably not related to this PR, but I noticed that in the spacing stories, the text is white-on-white when the theme is set to dark or dark-alt:

screenshot

calebeby (approved):

tylersticka:
> This is probably not related to this PR, but I noticed that in the spacing stories, the text is white-on-white when the theme is set to dark or dark-alt:

Yeah, those demos aren't really using any of our components, and the bespoke demo CSS isn't themed. So not related to this PR, but thanks for calling out!

PR #1595, opened 2021-11-22
Use logical properties consistently

### LINKED ISSUE #1576: Make use of logical properties consistently

tylersticka:
## Overview

This PR installs a plugin for linting logical properties and applies those fixes across all existing stylesheets.

Please note that the intent of this PR is to enforce consistent property usage, _not_ to optimize each and every component for different text directions and writing modes.

## Screenshots

Screen Shot 2021-11-22 at 3 02 16 PM

## Testing

1. View a few patterns with "text flow" set to something other than the default to verify that properties have been updated.
2. Observe basically every pattern in the deploy preview for regressions using the "default" text flow. I would most appreciate 👀 on Chrome, Edge and/or Safari (as Firefox is my primary browser and probably has disproportionate coverage).

---

- Fixes #1576

### DIFF
--- a/.changeset/quick-pets-yell.md
+++ b/.changeset/quick-pets-yell.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Use logical properties across existing styles to improve future-friendliness of patterns.
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+  plugins: ['stylelint-use-logical-spec'],
   extends: ['stylelint-config-cloudfour', 'stylelint-config-prettier'],
   rules: {
     // disable stylelint-scss rules that conflict with prettier
@@ -38,5 +39,12 @@ module.exports = {
     'suitcss/custom-property-no-outside-root': null,
     // we want to be able to compose :root for theme selectors (#1056)
     'suitcss/selector-root-no-composition': null,
+    'liberty/use-logical-spec': [
+      'always',
+      {
+        // Until Safari 15 adoption is higher
+        except: ['clear', 'float', /^border-.+-radius$/i],
+      },
+    ],
   },
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -76,6 +76,7 @@
         "stylelint": "14.1.0",
         "stylelint-config-cloudfour": "5.0.2",
         "stylelint-config-prettier": "9.0.3",
+        "stylelint-use-logical-spec": "^3.2.2",
         "through2": "4.0.2",
         "tiny-glob": "0.2.9",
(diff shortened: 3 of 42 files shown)


spaceninja:
Code looks good!

Just finished testing for regressions in Edge. I'll test Safari next.

I should note that I'm not super familiar with these patterns, so while I don't see any obvious problems, it might be worth tapping Danielle or another designer to do a regression pass as well.

spaceninja (approved):
Safari looks good too.

I noticed two unrelated bugs in Safari (that are also visible in v-next):

1) Large amount of white space below Overview blocks (visible on the Overview page and the Article Listing prototype).

2) Hovering an avatar on the Team List page turns the avatar square, in a weird flickering way that looks broken.

I'll file bugs for those.

I'm going to approve this PR, but I'd still appreciate a designer who's more familiar with these patterns doing a quick regression sweep.

dromo77:
I tested in Chrome since it looks like the other browsers have been covered. I didn't find any regressions, going to approve from a design standpoint.

dromo77 (approved):

PR #1607, opened 2021-11-30
Speaker pattern

### LINKED ISSUE #1463: Speaker pattern
On the [speaking event prototype](https://v-next--cloudfour-patterns.netlify.app/?path=/story/prototypes-speaking-event--example), there is a pattern similar to `Author`. It includes the speaker name and title. The `Author` pattern only supports a name and date field. We could update `Author` to include an optional text field for titles or we could add a new component.

Screen Shot 2021-08-02 at 4 11 56 PM

gerardo-rodriguez:
## Overview

This PR refactors the Author component so that it can be used for both article and presentation bylines.

## Screenshots

screencapture-localhost-6006-2021-12-03-10_18_33

## Testing

1. Review the refactored [Author pattern](https://deploy-preview-1607--cloudfour-patterns.netlify.app/?path=/docs/components-author--basic-usage)
2. Also review with VoiceOver to ensure the date isn't repeated twice

---

- Closes #1463


tylersticka (commented):
@gerardo-rodriguez It seems strange to have a pattern that so closely matches [the existing Author component](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/components-author--default-story).

Did you consider updating the Author component instead, replacing `date` with a more flexible/generic option?

gerardo-rodriguez:
@tylersticka Hmm...good question. They are _very_ similar. 

I can take a closer look! 🔍 👀

src/components/author/author.twig
```diff
@@ -7,17 +7,30 @@
...
+
+    {% if date is defined %}
+      <p>
+        <span class="u-hidden-visually">
+          Published on {{ date|date('F jS, Y') }}
+        </span>
+        <time aria-hidden="true" datetime="{{ date|date('Y-m-d') }}">
+          {{ date|date('F jS, Y') }}
+        </time>
+      </p>
```
gerardo-rodriguez, line 27: When testing with macOS VoiceOver, the date would get read twice before, which was a confusing experience.

A bit of research led me to [this tweet](https://twitter.com/LeonieWatson/status/1333078194925264898) stating:
> AFAIK `<time datetime>` is not well accessibility supported by browsers/screen readers.

With this in mind, I decided to make the screen reader experience better while still keeping the `<time>` but hiding it from the screen reader. I kept the `<time>` element because per the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time), it allows for better search engine results (though, admittingly, I don't know much about this):
> The `<time>` HTML element represents a specific period in time. It may include the `datetime` attribute to translate dates into machine-readable format, allowing for better search engine results or custom features such as reminders.

gerardo-rodriguez:
@tylersticka This PR is ready for another review, appreciate your feedback, thanks! 🙂

src/components/author/author.stories.mdx
```diff
@@ -40,30 +58,87 @@ const megan = {
...
+        authors: authors(args.count),
+        date: new Date('March 31, 2021'),
+      })
+    }
+  </Story>
+</Canvas>
+
+### Without `date` defined
+
+If the `date` is _not_ defined, no additional information is rendered.
```
tylersticka, line 117: Is there a reason this is specific to multiple authors?

Maybe we should reformat these stories to start with the simplest use case (no date or title), then build up from there by adding a date, and multiple authors, etc., mentioning any limitations along the way?

src/components/author/author.twig
tylersticka: Your context is great, but I wish more of it would be included as a Twig comment within the template itself so this is as clear in the future as it is while reviewing this PR.

tylersticka (commented):
Thanks for continuing to revise this, @gerardo-rodriguez!

src/components/author/author.stories.mdx
gerardo-rodriguez: > Is there a reason this is specific to multiple authors?

Great question, @tylersticka!

Perhaps I've misunderstood the requirements?

Before, the Author component would _always_ render a publish date. 

The Speaker requirements were to [include a job title for single speakers and drop the title for multiple speakers](https://github.com/cloudfour/cloudfour.com-patterns/issues/1463#issuecomment-934891778).

Merging those two, I landed at:
  - If 1 author and `date` is defined: Render author avatar, linked name, and date
  - If 1 author and no `date` is defined: Render author avatar, linked name, and `job_title`
  - If multiple authors and `date` is defined: Render all author avatars, linked names, and date
  - If multiple authors and no `date` is defined: Render all author avatars and linked names

> Maybe we should reformat these stories to start with the simplest use case (no date or title)

I didn't know this was a use case! 😅 If it is, the template needs to be updated to handle this use case. Currently, an empty `<p>` tag is rendered. 🔍 👀 

I can make this change, though! 🏗️

src/components/author/author.twig
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1607/commits/63349fd2f8c6230fc1e3de84181068f9a0517ac1

src/components/author/author.stories.mdx
tylersticka: It's important to remember that the patterns and components in a design system are meant to _enable_ feature requirements at a minimum, but ideally they should be flexible enough to anticipate _future_ requirements.

So our Author component that we're starting from was designed for blog author bylines, which had a specific set of requirements:

1. They must display 1 or more avatars.
2. They must display 1 or more names.
3. They must include a publication date.

But that doesn't quite gel with the speaker requirements you mentioned:

> include a job title for single speakers and drop the title for multiple speakers

So now, we have three options for how to proceed:

1. We can design a whole new component, which was what the first draft of this PR contained. But that seems a bit unnecessary: There's way more that's _similar_ about speaker listings and author listings than there is different.
2. We can _append_ these requirements to the existing component, which is kind of where we are now. That will keep things pretty DRY and will absolutely meet our immediate needs, but it sort of repeats the mistakes that got us here in the first place: We're thinking so specifically about our _current_ needs, that we aren't thinking about how to meet our _future_ needs.
3. We re-think the original requirements to see if we can satisfy both use cases.

I think option 3 is the way to go. And if we think about it from that perspective, here's how I'd revise those original Author requirements:

1. They must display 1 or more avatars.
2. They must display 1 or more names.
3. They ~must~ **may optionally** include a ~publication~ date **or additional text content to display on a second line**.

Not only is that a simpler set of requirements that will be easier for designers and developers to understand, it broadens the potential application of this pattern to more than blog posts and single events.

Let me know if we need to discuss more with voices!

gerardo-rodriguez: Thanks, @tylersticka, that makes sense! 

What happens if both a date and a job title exist at the same time? Or are you suggesting the job title or date is passed into a generic "second line" property?

The current template props are:
- `authors` array where the `author` object contains the job title
- `date`

Are you suggesting to remove the job title from the author object and instead have:
- `authors` array that only includes avatar, name, link
- `second_line_content` (but a better name) that handles a date, or a job title, or any other string?

Thank you! 😄

tylersticka: I'm suggesting that the `date` would be kind of a shorthand property for the `second_line_content` (which I might call `extra` or `meta` or something along those lines).

So if I were to think of this as a Twig template, it might be like this toward the be:

```twig
{% set _extra_content %}
  {% block extra %}
    {% if extra %}
      {{extra}}
    {% elseif date %}
      (Do date stuff here!)
    {% endif %}
  {% endblock %}
{% endset %}
```

Then you can check if `_extra_content` includes anything before outputting the `div`...

```twig
{% if _extra_content|trim %}
  <div>
    {{_extra_content}}
  </div>
{% endif %}
```

Check out the Badge component template if you'd like a similar example.

gerardo-rodriguez: Thanks, @tylersticka!

> Check out the Badge component template if you'd like a similar example.

Ha! I was reviewing the Alert component before I read your suggestion and I forgot about this neat pattern (which is also implemented in Badge).

```twig
{% block content %}
  <p>
    {{message|default('Hello world!')}}
  </p>
{% endblock %}
```
Definitely more flexible. 😉

gerardo-rodriguez: Refactored! https://github.com/cloudfour/cloudfour.com-patterns/pull/1607/commits/e6a86975f1c0a4f1a90bff1443fc84bba849b5c2

gerardo-rodriguez:
Thanks again, @tylersticka! This PR is ready for another review. 🎉

src/components/author/author.twig
```diff
@@ -1,23 +1,76 @@
...
+#}
+{% set _meta_content %}
+  {% block meta_content %}
+    {% if meta %}
+      <p>{{ meta }}</p>
+    {% elseif date %}
+      <p>
+        {# This creates a more accessible, and less confusing, user experience #}
+        <span class="u-hidden-visually">
+          Published on {{ date|date('F jS, Y') }}
```
gerardo-rodriguez, line 20: The "Published on" text should probably be a template property so it can be changed if needed.

src/components/author/author.twig
```diff
@@ -1,23 +1,76 @@
+{#
+  This component accepts the following properties
+  - `meta` (optional, string)
+  - `date` (optional, date object)
+
+  If both the `meta` and `date` properties contain values, the `meta` value has
+  a higher priority and will override the `date` value.
+
+  It is recommened to use the `date` property when wanted to render a date as the
```
tylersticka, line 9: Nit: Typo

src/components/author/author.twig
tylersticka: Yeah, that's a good idea now that the use cases have been broadened. Also a good way to future-proof. But I think this is a great default value!

src/components/author/author.twig
```diff
@@ -1,23 +1,76 @@
...
   {% endembed %}
-  <div class="o-media__content">
-    <div>
-      <span class="u-hidden-visually">Authored by</span>
+
+  {# Author content #}
+  <div class="c-author__content-container o-media__content">
+    {# Author links #}
+    <p class="c-author__author">
+      <span class="u-hidden-visually">By</span>
```
tylersticka, line 61: In the spirit of your date prefix suggestion, should this also be configurable? (But I still think this is a good default!)

tylersticka (commented):
This version's awesome, @gerardo-rodriguez! More features but it actually seems simpler to use when you read the docs, which is the ideal. My comments are really just affirming some of your instincts regarding accessibility prefix options.

src/components/author/author.twig
gerardo-rodriguez: Edited! https://github.com/cloudfour/cloudfour.com-patterns/pull/1607/commits/e0e22d6b45b6dc28ef23f40593e4530788f45ce6

src/components/author/author.twig
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1607/commits/b6e22592b0fd39ad91ead1037f0c1671be00b4d5

src/components/author/author.twig
gerardo-rodriguez: Updated! https://github.com/cloudfour/cloudfour.com-patterns/pull/1607/commits/f4074b232547985fbcecfeb2f456ce72ce0baf40

tylersticka (approved):

PR #1611, opened 2021-12-07
Fix Card cover hover issue in Safari

### LINKED ISSUE #1597: Hovering a card with a cover image causes rounded corners to temporarily flicker in Safari
In Safari, when you mouse over an avatar on the Team List page, it flickers for a second revealing the square shape of the avatar.

Untitled

tylersticka:
## Overview

This is a [depressingly old](https://bugs.webkit.org/show_bug.cgi?id=68196) bug. [This solution](https://stackoverflow.com/a/16681137) seemed to function the best, but I simplified it because the pattern does not use 3D transforms and those properties no longer require vendor prefixes.

## Screenshots

safari-card-cover-fix

## Testing

Review linked cards with covers. Confirm that they no longer break overflow or border radius on hover in Safari.

---

- Fixes #1597

### DIFF
--- a/.changeset/modern-turtles-hear.md
+++ b/.changeset/modern-turtles-hear.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Fix Card covers breaking overflow and border radius rules on hover in Safari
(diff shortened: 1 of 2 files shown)


Paul-Hebert (approved):
LGTM :+1:

PR #1612, opened 2021-12-07
Fix issues related to columned lists in Safari

### LINKED ISSUE #1596: Large amount of white space below Overview blocks in Safari
There's a large amount of whitespace visible in Safari below Overview blocks. This is visible on the Overview page and the Article Listing prototype.

Screen Shot 2021-11-23 at 4 40 06 PM

tylersticka:
## Overview

This PR fixes three Safari issues that were visible in different situations in the article topics listing:

- Any `u-hidden-visually` elements within the list would cause Safari to increase the scrollable height of the window. This was resolved by adding `position: relative` to children of the list. (It was not effective to add this to the parent.)
- Some elements would appear divided between columns. This was resolved by adding `break-inside: avoid` to children of the list. (`page-break-inside` and `-webkit-column-break-inside` are common suggestions, but these are older properties that didn't seem to have any more effect)
- The bottom margin of an item could carry into the next column. This was fixed by swapping `u-space-block-end-0` with `u-pad-block-end-0` in the relevant templates.

### Before

Highlighting visually divided items and extra space...

Screen Shot 2021-12-07 at 1 09 56 PM

### After

Screen Shot 2021-12-07 at 1 09 18 PM

## Testing

The issue was previously symptomatic in "Objects ▸ Overview ▸ Advanced" and "Prototypes ▸ Article Listing ▸ Example," so those are good stories to verify the fix in.

---

- Fixes #1596

### DIFF
--- a/.changeset/ten-candles-guess.md
+++ b/.changeset/ten-candles-guess.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Fix for List objects with columns in Safari that include visually hidden elements or elements with content divided between columns
--- a/src/objects/list/list.scss
+++ b/src/objects/list/list.scss
@@ -4,33 +4,29 @@
 @use '../../mixins/fluid';
 @use '../../mixins/media-query';
 
-/**
- * Override browser defaults. Note that this means it's very important to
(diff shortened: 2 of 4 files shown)


gerardo-rodriguez (approved):
Wow, excellent work, @tylersticka! 🎉

PR #1619, opened 2021-12-23
Update pleasantest and use getAccessibilityTree in tests

calebeby:
This updates Pleasantest to the version that has getAccessibilityTree, and changes the tests which benefit from getAccessibilityTree to use it.


src/components/author/author.test.ts
```diff
@@ -1,62 +1,67 @@
...
-    })
-  );
-});
+    // Confirm the meta value is rendered and the date is not rendered
+    expect(await getAccessibilityTree(body)).toMatchInlineSnapshot(`
+        text "By"
+        link "Shakira Isabel Mebarak Ripoll"
+          text "Shakira Isabel Mebarak Ripoll"
+        text "Singer and songwriter"
+      `);
```
Paul-Hebert, line 65: Nit: I wonder if it would match the original intent of this test better to do something like this?

```js
const a11yTree = await getAccessibilityTree(body));

expect(a11yTree.includes('Singer and songwriter')).toBeTrue();
expect(a11yTree.includes('Published on March 31st, 2021')).toBeFalse();
```

(This is not a blocking comment. Feel free to ignore it.)

src/components/comment/comment.test.ts
```diff
@@ -33,8 +33,23 @@ test(
...
+            text "Test author name"
+            text "said:"
+        text "Test content"
+        contentinfo
+          link "Permalink to Test author name’s Jan 1, 2000 comment"
+            text "Permalink to Test author name’s"
+            text "Jan 1, 2000"
+            text "comment"
+          button "Reply"
+    `);
```
Paul-Hebert, line 52: Nit: I wonder if this test would be easier to read with `includeText: false`?

src/components/comment/comment.test.ts
```diff
@@ -44,17 +59,44 @@ test(
...
 
     // Updated state: reply form is no longer hidden
     await expect(form).toBeVisible();
     // Reply button is hidden
     await expect(replyButton).not.toBeVisible();
-    // The first input or textarea should be focused.
-    // (In practice this will always be a textarea, but if the form is
-    // changed to lead with an input in the future, and that input is selected,
-    // this test should still pass.)
```
Paul-Hebert, line 55: It's less clear to me now what this test is checking for. I wonder if we should add a comment above the snapshot explaining that we're interested in checking that the first focusable element in the form should be focused?

Paul-Hebert (dismissed):
These changes make sense to me and it's cool to see `getAccessibilityTree` used. I left a couple non-blocking nits inline.

src/components/comment/comment.test.ts
calebeby: Yeah sure 👍 . It is also checking that all the parts of the form are rendered with the intended text + roles + structure

src/components/author/author.test.ts
calebeby: I think the snapshot is better because it will be more clear when it fails by showing where it specifically failed, and also it will fail in more cases (for example if things changed order or role, etc), and since it is an inline snapshot it just requires telling Jest to update it if the change is intended instead of having to update it manually.

Paul-Hebert (approved):
LGTM!

PR #1653, opened 2022-02-16
Lint Fix for `currentcolor`

spaceninja:
## Overview

`currentColor` was brought into CSS from SVG, and that's why it kept
the unusual capitalization. The CSS level 4 spec formally changed it
to `currentcolor` and Stylelint recently updated their rules to
enforce that.

See discussion here: https://github.com/stylelint/stylelint/issues/5863

### DIFF
--- a/src/base/_defaults.scss
+++ b/src/base/_defaults.scss
@@ -291,7 +291,7 @@ address {
  */
 
 hr {
-  border-color: currentColor;
+  border-color: currentcolor;
   border-style: solid;
   border-width: math.div(size.$edge-medium, 2) 0; /* 1 */
   color: var(--theme-color-border-text-group);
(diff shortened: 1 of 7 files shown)


spaceninja:
Browser support for this seems to be solid. If you'd prefer, I could update our stylelint config to use the [`camelCaseSvgKeywords: true`](https://stylelint.io/user-guide/rules/list/value-keyword-case/) rule to enforce the `currentColor` casing. That said, my default stance is to avoid diverging from standard Stylelint config unless it's causing problems.

tylersticka (commented):
Please use the `camelCaseSvgKeywords` option to preserve interoperability between the CSS and SVG specs: https://stylelint.io/user-guide/rules/list/value-keyword-case/#camelcasesvgkeywords-true--false-default-false

Thank you!

spaceninja:
Closing this in favor of updating our stylelint config to v5.1.0 which includes the `camelCaseSvgKeywords` rule.

PR #1675, opened 2022-03-12
cancel running CI jobs when a PR is updated

spaceninja:
@see https://github.com/nodejs/node/pull/42017


gerardo-rodriguez (approved):
I don't know how to test but the code LGTM, excellent change! 👍

PR #1679, opened 2022-03-15
"Get Weekly Digests" pattern

### LINKED ISSUE #1541: Get Weekly Digests pattern
This task is a continuation of #1508 where the Button Swap pattern was introduced for the "Get Notifications"/"Turn Off Notifications" UI interactions.

This task focuses on the "Get Weekly Digests" UI interactions which reveals a subscription form. 

image

gerardo-rodriguez:
## Overview

This PR introduces a Subscription Choices component.

- It includes an embedded Button Swap component for the "Get Notifications" button
- It shows/hides the email subscription form

## Screenshots

demo

## Testing

Preview: https://deploy-preview-1679--cloudfour-patterns.netlify.app/?path=/story/components-subscribe--default-story

### Accessibility

- [x] Confirm you can use the keyboard to navigate to and away from the form
- [x] There _should not_ be a form hide delay when tabbing backward from the form to the "Get Weekly Digests" link
- [x] There _should_ be a form hide delay when you tab forward out of the form
- [x] Confirm VoiceOver + Safari experience is accessible

### General

With both keyboard and mouse inputs, clicking/activating the "Get Weekly Digests" link should:
- [x] Reveal the form over the controls
- [x] Set the `:focus` onto the email input field

If the form is visible, hitting <kbd>Escape</kbd> on the keyboard should:
- [x] Hide the form
- [x] Move the `:focus` to the "Get Weekly Digests" link

### Theming 

- [x] Confirm the form background color looks correct in each theme.

## Known issues

~I was unable to figure out how to set the form background color for the "dark alternate" theme. Any ideas @tylersticka?~ (This has been resolved)

Screen Shot 2022-03-22 at 8 26 12 AM

---

- Closes #1541


src/components/subscription-choices/subscription-choices.scss
```diff
@@ -0,0 +1,72 @@
...
+///
+/// Subscription Choices
+///
+
+.c-subscription-choices {
+  position: relative;
+}
+
+.c-subscription-choices__controls-wrapper {
+  display: grid;
```
gerardo-rodriguez, line 35: I wonder if I should use `flex` here instead. It seems a bit much to use `grid` for managing two buttons? 🤔

gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/975e09a519bea456fba87a65488739160e16e779

src/components/subscription-choices/subscription-choices.scss
```diff
@@ -0,0 +1,90 @@
...
+.c-subscription-choices__control {
+  flex-grow: 1;
+}
+
+///
+/// 1. Ensures the form background color matches the current theme
+///
+
+.c-subscription-choices__form {
+  background-color: var(--theme-color-background-subscription-form); // 1
```
tylersticka, line 62: > I was unable to figure out how to set the form background color for the "dark alternate" theme. Any ideas @tylersticka?

You can accomplish this by using the same property as the theme itself:

```suggestion
  background-color: var(--theme-color-background-base); // 1
```

src/components/subscription-choices/subscription-choices.scss
```diff
@@ -0,0 +1,90 @@
...
+
+// Dark theme
+@include theme.props(dark) {
+  --theme-color-background-subscription-form: #{color.$brand-primary};
+}
+
+// Alternate theme
+@include theme.props(alternate) {
+  --theme-color-background-subscription-form: #{color.$base-gray-lighter};
+}
```
tylersticka, line 24: These are not necessary if you use the existing theme properties instead of setting new ones (see my other comment further down).

src/components/subscription-choices/subscription-choices.scss
```diff
@@ -0,0 +1,90 @@
...
+  inset-block-start: 0;
+  inset-inline-start: 0;
+  position: absolute;
+
+  ///
+  /// 1. Allows for the visually hidden form to be keyboard accessible, will be
+  ///    visually hidden when elements within the form are not in focus.
+  ///
+
+  label,
```
tylersticka, line 74: Is it intentional that you're targeting a default `label` tag here?

src/components/subscription-choices/subscription-choices.scss
```diff
@@ -0,0 +1,90 @@
...
+    display: flex;
+  }
+
+  @media (min-width: breakpoint.$s) {
+    flex-direction: row;
+  }
+}
+
+.c-subscription-choices__control {
+  flex-grow: 1;
```
tylersticka, line 54: Most guides recommend using the shorthand because browsers make some other related adjustments based on this...

```suggestion
  flex: 1;
```

src/components/subscription-choices/subscription-choices.scss
```diff
@@ -0,0 +1,90 @@
...
+  gap: 1em;
+  margin: size.$rhythm-default 0;
+
+  .c-button-swap {
+    flex-grow: 1;
+  }
+
+  .c-button-swap__subscribe {
+    display: flex;
+  }
```
tylersticka, line 46: It's not a good idea to target one component class name from another like this. You might consider...

1. Merging `c-button-swap` into this component (since this is the only context for it currently).
2. Adding a unique element class for these things with this component's prefixes so the separation of concerns is evident.
3. Create a new modifier for the `c-button-swap` component so it will work in this type of context.
4. Refactor the button swap component so that these overrides are not necessary.

src/components/subscription-choices/subscription-choices.scss
```diff
@@ -0,0 +1,90 @@
...
+// Alternate theme
+@include theme.props(alternate) {
+  --theme-color-background-subscription-form: #{color.$base-gray-lighter};
+}
+
+///
+/// Subscription Choices
+///
+
+.c-subscription-choices {
```
tylersticka, line 30: Could this be simplified to `.c-subscribe`?

src/components/subscription-choices/subscription-choices.scss
gerardo-rodriguez: Genious! 🎉

src/components/subscription-choices/subscription-choices.scss
spaceninja: +1

This is a common BEM pattern:

```
<div class="button button--modifier">
  <div class="button__avatar avatar avatar--modifier">
      <image class="button__avatar-image avatar__image">
```

This allows you to semantically separate the styles that originate from alpha and beta, while still applying styles from both to the same element.

The only time this becomes a concern is if they're in conflict, then you need to understand the source order of your rendered CSS — but in practice, I've run into this concern very rarely. Typically, as in the example above, you'll be targeting different aspects. Likely, in this example, `button__avatar-image` is dealing with layout concerns like sizing, margins, etc, while `avatar__image` is dealing with things like setting aspect ratio, etc.

src/components/subscription-choices/subscription-choices.scss
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/1362e84a6c29b9dc1c0bdaf05670702503bcf8f8

src/components/subscription-choices/subscription-choices.scss
gerardo-rodriguez: Oops, I should add a CSS class to the label. Good catch!

gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/d808584988c8c3e56bdaaea2fbc56c4b951bc3ed

src/components/subscription-choices/subscription-choices.scss
gerardo-rodriguez: Thanks! #TIL

gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/9d9b917ad2c5bcf6860d998bb8d75207bd7b5c32

spaceninja: For a bit more detail:

`flex: 1` is equivalent to `flex: 1 1 0`, or `flex-grow: 1; flex-shrink: 1; flex-basis: 0;`

When you only specify `flex-grow: 1` you're leaving the default flex values of `flex-shrink: 1;` and `flex-basis: auto;`.

The difference between setting your `flex-basis` to `0` or `auto` will depend on your layout, but since troubleshooting flex layouts can be confusing, it's usually better to be explicit about the expected behavior you want by either using the shorthand, or by explicitly setting all three values.

src/components/subscription-choices/subscription-choices.scss
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/fa3619d19e1869a11dcd8e189a6e3d2fe025c8d5

src/components/subscription-choices/subscription-choices.scss
spaceninja: Technically, BEM says you should rarely, if ever, target an element. Instead, set a class. This serves two purposes: First, it keeps all selectors on the same specificity level, and second, the BEM class names are self-documenting, so it's more clear what is happening to someone reading the markup.

`<label>` is generic. `<label class="subscription-form__label">` makes it clear what BEM module this is part of and where I can find the CSS.

src/components/subscription-choices/subscription-choices.test.ts
```diff
@@ -0,0 +1,255 @@
...
+      // Email input should be in focus
+      const emailInput = await screen.getByRole('textbox', { name: 'Email' });
+      await expect(emailInput).toHaveFocus();
+
+      // Tab again to get to the Submit button
+      await page.keyboard.press('Tab');
+
+      // Submit button should be in focus
+      const submitBtn = await screen.getByRole('button', { name: 'Subscribe' });
+      await expect(submitBtn).toHaveFocus();
```
calebeby, line 97: nit:
```suggestion
      const subscribeBtn = await screen.getByRole('button', { name: 'Subscribe' });
      await expect(subscribeBtn).toHaveFocus();
```

src/components/subscription-choices/subscription-choices.test.ts
```diff
@@ -0,0 +1,255 @@
...
+        formWidth: formEl.clientWidth,
+      });
+
+      // Confirm the form is visually hidden by default
+      const form = await screen.getByRole('form', {
+        name: 'Get Weekly Digests',
+      });
+      let { formHeight, formWidth } = await form.evaluate(getFormDimensions);
+      expect(formHeight).toBeLessThanOrEqual(1);
+      expect(formWidth).toBeLessThanOrEqual(1);
```
calebeby, line 76: @gerardo-rodriguez Is there a reason we can't use [`.toBeVisible()`](https://github.com/testing-library/jest-dom#tobevisible) here?

src/components/subscription-choices/subscription-choices.test.ts
```diff
@@ -0,0 +1,255 @@
...
+      let form = await screen.getByRole('form');
+      let { formAction, emailInputPlaceHolder } = await form.evaluate(
+        (formEl) => ({
+          formAction: formEl.getAttribute('action'),
+          emailInputPlaceHolder: formEl.querySelector('input')?.placeholder,
+        })
+      );
+      expect(formAction).toBe(
+        'https://cloudfour.us13.list-manage.com/subscribe/post?u=ce064f42c86a5982dd218d4de&amp;id=7e505a6a67'
+      );
```
calebeby, line 196: Can we use [`.toHaveAttribute`](https://github.com/testing-library/jest-dom#tohaveattribute) here? (for placeholder too)?

src/components/subscription-choices/subscription-choices.test.ts
gerardo-rodriguez: Great question, @calebeby!

The form is _technically_ visible so that assertion fails. The reason it is still visible is that it uses the [`@include a11y.sr-only` mixin](https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/files#diff-feb19a35856697c41225bad7dda4443f13f7ccab3052b5c9a15bb1d2203cc10fR58) to ensure the form is accessible via keyboard navigation. It is only _visually_ hidden.

src/components/subscription-choices/subscription-choices.test.ts
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/9134022b79a690bc2a1bbda7a0933b522505be5d

src/components/subscription-choices/subscription-choices.test.ts
calebeby: The toBeVisible matcher is supposed to represent visual visibility. So .not.toBeVisible should pass even if for example an element has `opacity: 0` (so it'd still be in the accessibility tree), otherwise there is a bug with .not.toBeVisible. Mind opening an issue?

calebeby:
@gerardo-rodriguez it's possible that this discussion already happened, but would it make sense to allow pressing the escape key in the email form to switch it back to the two buttons option? It feels like there should be a way to get back without un-focusing and waiting.

src/components/subscription-choices/subscription-choices.test.ts
gerardo-rodriguez: @calebeby These are the values for the [`a11y.sr-only` mixin](https://github.com/cloudfour/cloudfour.com-patterns/blob/v-next/src/mixins/_a11y.scss):
```css
  block-size: 1px,
  border: 0,
  clip: rect(0 0 0 0),
  clip-path: polygon(0 0, 0 0, 0 0),
  inline-size: 1px,
  margin: -1px,
  overflow: hidden,
  padding: 0,
  position: absolute,
  white-space: nowrap,
```
I don't think it's a bug, the `opacity` is not set to `0`. The form is _visible_ to assistive technologies like a screen reader. If you tab into the form, the form will be visually visible as well. This is the same behavior as the "Skip to main content" link.

Screen Shot 2022-03-22 at 3 12 11 PM

calebeby: I still think it is a bug. opacity:0 is also still visible to assistive technologies too (at least based on what Chrome devtools shows). My understanding of the toBeVisible matcher is that it is supposed to check that something is visually visible, not that it is present in the accessibility tree.

gerardo-rodriguez:
> @gerardo-rodriguez it's possible that this discussion already happened, but would it make sense to allow pressing the escape key in the email form to switch it back to the two buttons option? It feels like there should be a way to get back without un-focusing and waiting.

@calebeby This has not been discussed yet. It's a great question and I'm not sure the answer. I was unable to find an established accessible pattern that matched our design (you can read a bit more backstory [here](https://github.com/cloudfour/cloudfour.com-patterns/issues/1541#issuecomment-1068598799)). An escape key makes perfect sense if it was a modal, for example. But it's less clear for this UI/UX to me.

From a pure screenreader non-visual perspective, the experience is what you'd expect it to be, you are tabbing through elements on the page. It doesn't matter that the buttons visually hide for this experience.

From a visual perspective, one could argue escape key functionality would be beneficial. 

Hmm... 🤔

src/components/subscription-choices/subscription-choices.test.ts
gerardo-rodriguez: > My understanding of the toBeVisible matcher is that it is supposed to check that something is visually visible, not that it is present in the accessibility tree.

@calebeby Interesting! I hadn't thought about it.

I did a quick search and found [this](https://github.com/testing-library/jest-dom/issues/338#issuecomment-878882097) (off of your GH issue 🙃):

> In testing-library we usually go with "visible" = possibly perceivable + accessible by assistive technology. E.g. display: none is impossible to be perceivable and inaccessible. On the other hand zero-dimensions is impossible to be perceivable but still be considered visible because the element would be exposed to assistive technology.

If we subscribe to that, then `toBeVisible()` is doing what it's supposed to be doing. 🙂

gerardo-rodriguez:
> > @gerardo-rodriguez it's possible that this discussion already happened, but would it make sense to allow pressing the escape key in the email form to switch it back to the two buttons option? It feels like there should be a way to get back without un-focusing and waiting.
> 
> @calebeby This has not been discussed yet. It's a great question and I'm not sure the answer. I was unable to find an established accessible pattern that matched our design (you can read a bit more backstory [here](https://github.com/cloudfour/cloudfour.com-patterns/issues/1541#issuecomment-1068598799)). An escape key makes perfect sense if it was a modal, for example. But it's less clear for this UI/UX to me.
> 
> From a pure screenreader non-visual perspective, the experience is what you'd expect it to be, you are tabbing through elements on the page. It doesn't matter that the buttons visually hide for this experience.
> 
> From a visual perspective, one could argue escape key functionality would be beneficial.
> 
> Hmm... 🤔

@cloudfour/design @cloudfour/dev Thoughts?

tylersticka:
> From a visual perspective, one could argue escape key functionality would be beneficial.

I'm not really hearing a reason _not_ to? Sounds like a good idea assuming it doesn't balloon the scope.

src/components/subscription-choices/subscription-choices.scss
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/5766e69a1b9456c6eee379e8bad03603e49300de

gerardo-rodriguez: I decided to change this to use CSS grid, which avoids this issue completely. https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/afc52b34481e6ad7e0a367573588b0190a93bfac

gerardo-rodriguez:
@tylersticka @calebeby @spaceninja Thanks for your feedback, this PR is ready for another review!

> > From a visual perspective, one could argue escape key functionality would be beneficial.
> 
> I'm not really hearing a reason _not_ to? Sounds like a good idea assuming it doesn't balloon the scope.

[I've added this functionality](https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/8273485bcdbf8aecebe5147db25df6a057b3958e). 😉

src/components/subscription-choices/subscription-choices.test.ts
calebeby: It seems like based on that discussion, the zero-dimensions case is chosen because of the limitations of JSDOM, where elements don't have size.

In pleasantest, we don't have that constraint, we _could_ make the behavior match the zero-opacity case where the element is accessible by assistive technology but not possibly perceivable. In the zero-opacity case, the element is not considered visible by jest-dom. We could make pleasantest also consider the zero-size case to not be visible.

:shrug: Probably a discussion that can happen outside this PR.

In this PR, would it make sense to separate out a `expectElementToHaveSize` function?

```js
await expectElementToHaveSize(form);
```

```js
const expectElementToHaveSize = async () => {
	const { formHeight, formWidth } = await form.evaluate((formEl: HTMLElement) => ({
		formHeight: formEl.clientHeight,
		formWidth: formEl.clientWidth,
	}));
	expect(formHeight).toBeGreaterThan(1);
	expect(formWidth).toBeGreaterThan(1);
}
```

.github/workflows/ci.yml
```diff
@@ -80,10 +80,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Use Node.js 12
+      - name: Use Node.js 16
         uses: actions/setup-node@v3
         with:
-          node-version: 12
+          node-version: 16
```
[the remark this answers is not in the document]
tylersticka: Fine by me since this keeps us in sync with what our `.nvmrc` resolves to 👍

src/components/subscription-choices/subscription-choices.stories.mdx
```diff
@@ -0,0 +1,179 @@
...
+| `weekly_digests_heading`                 | string | `'Get Weekly Digests'`                  | The subscription form heading                                                                                   |
+| `email_input_label`                      | string | `'Email'`                               | The email input label                                                                                           |
+| `email_input_placeholder`                | string | `'Your Email Address'`                  | The email input placeholder text                                                                                |
+| `submit_btn_label`                       | string | `'Subscribe'`                           | The subscription form submit button label                                                                       |
+
+## JavaScript
+
+The Subscription Choices component UX can be progressively enhanced by adding
+a 1 second form hide delay when `:focus` is moved away from the Subscription
+Choices component elements.
```
tylersticka, line 123: Should we also mention the `esc` functionality? (I assume that's handled with JavaScript.)

src/components/subscription-choices/subscription-choices.stories.mdx
```diff
@@ -0,0 +1,179 @@
...
+
+### Syntax
+
+```js
+initSubscriptionChoices(subscriptionChoicesEl);
```
tylersticka, line 128: Does the name of the JS stuff need to be updated to match the new component name?

src/components/subscription-choices/subscription-choices.twig
```diff
@@ -0,0 +1,61 @@
...
+        swapped_label: notifications_btn_swapped_label|default('Notifications have been turned on.'),
+        swapped_visual_label: notifications_btn_swapped_visual_label|default('Turn off notifications'),
+      } only %}
+
+      {% include '@cloudfour/components/button/button.twig' with {
+        href: "#subscription-choices-#{form_id}",
+        class: "c-subscribe__control
+js-subscription-choices__control
+js-subscription-choices__get-weekly-digests-btn
+#{weekly_digests_btn_class|default('')}",
```
tylersticka, line 25: Nit: In your other examples, you're assembling a class by joining an array. I'm fine with either approach, but we should probably be consistent within the component.

src/components/button-swap/button-swap.stories.mdx
```diff
@@ -74,6 +74,7 @@ You have the ability to pass `initialCallback` and `swappedCallback` callback fu
 ## Template Properties
 
 - `class` (string): Appends a CSS class(es) to the root element.
+- `button_class` (string): Appends a CSS class(es) to the button element.
```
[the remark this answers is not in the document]
gerardo-rodriguez: Oh, excellent suggestion, thank you!

src/components/subscription-choices/subscription-choices.scss
```diff
@@ -0,0 +1,62 @@
+@use '../../compiled/tokens/scss/breakpoint';
+@use '../../compiled/tokens/scss/color';
+@use '../../compiled/tokens/scss/size';
+@use '../../mixins/a11y';
+@use '../../mixins/theme';
+
+///
+/// Subscription Choices
```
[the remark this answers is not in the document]
gerardo-rodriguez: Good catch!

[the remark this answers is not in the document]
gerardo-rodriguez:
```suggestion
/// Subscribe component
```

src/components/subscribe/subscribe.test.ts
```diff
@@ -0,0 +1,273 @@
...
+// Helper to initialize the component JS
+const initJS = (utils: PleasantestUtils) =>
+  utils.runJS(`
+    import { initSubscriptionChoices } from './subscription-choices'
+    export default () => initSubscriptionChoices(
+      document.querySelector('.js-subscription-choices')
+    )
+  `);
+
+describe('Subscription Choices', () => {
```
gerardo-rodriguez, line 22:
```suggestion
describe('Subscription', () => {
```

.changeset/little-colts-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add Subscription Choices component
```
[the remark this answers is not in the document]
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/2ab70153d8c4054bb303f242e9c8fe7314507e26

src/components/subscription-choices/subscription-choices.scss
```diff
@@ -0,0 +1,62 @@
...
+  ///
+  /// 1. Allows for the visually hidden form to be keyboard accessible, will be
+  ///    visually hidden when elements within the form are not in focus.
+  ///
+
+  .c-subscribe__form-input-label,
+  .c-subscribe:not(.activate-form) // 1
+    &:not(:target):not(:focus-within) {
+    @include a11y.sr-only;
+  }
```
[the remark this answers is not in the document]
gerardo-rodriguez: @tylersticka I can't make this change, as soon as I do, it resets itself when I save the file. 🤔

src/components/subscription-choices/subscription-choices.twig
```diff
@@ -0,0 +1,61 @@
...
+  <div class="c-subscribe__controls">
+    <{{_heading_tag}}>
+      {{never_miss_article_heading|default('Never miss an article!')}}
+    </{{_heading_tag}}>
+
+    <div class="c-subscribe__controls-wrapper">
+      {% include '@cloudfour/components/button-swap/button-swap.twig' with {
+        button_class: "c-subscribe__control
+js-subscription-choices__control
+#{notifications_btn_class|default('')}",
```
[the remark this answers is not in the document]
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/2a85120d80af06a4f1be3d8b01206d0f04e786ff

src/components/subscription-choices/subscription-choices.twig
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/2a85120d80af06a4f1be3d8b01206d0f04e786ff

src/components/subscription-choices/subscription-choices.stories.mdx
```diff
@@ -0,0 +1,179 @@
...
+  </Story>
+</Canvas>
+
+## Template Properties
+
+### Required
+
+| property  | type   | default | description                                                          |
+| --------- | ------ | ------- | -------------------------------------------------------------------- |
+| `form_id` | string | `''`    | Appends this value to the weekly digest form `id`, should be unique. |
```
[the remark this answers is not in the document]
gerardo-rodriguez: Ah, thanks for the heads up, @tylersticka! 🎉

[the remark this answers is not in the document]
gerardo-rodriguez: Updated! https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/297f7044e981a72637c3bc80f67e8d01aa08b98a

src/components/subscribe/subscribe.test.ts
```diff
@@ -0,0 +1,273 @@
...
+          status
+            text "Notifications have been turned off."
+          button "Get notifications"
+          link "Get Weekly Digests"
+            text "Get Weekly Digests"
+          form "Get Weekly Digests"
+            heading "Get Weekly Digests"
+              text "Get Weekly Digests"
+            text "Email"
+            textbox
```
gerardo-rodriguez, line 44: I switched to using the Input Group  with an Input component and this value changed. Before when I was using a vanilla `<input>`, it read `textbox "Email"`. Now it only reads `textbox`. The "name" value seems to have gone away?

I reviewed both the vanilla and Input Group with the Input component and I was unable to notice a difference in the markup so I'm not sure why this changed. 🤔

src/components/subscription-choices/subscription-choices.scss
```diff
@@ -0,0 +1,62 @@
...
+.c-subscribe__form-input-group {
+  display: flex;
+  flex-direction: column;
+  gap: 1em;
+  margin: size.$rhythm-default 0;
+
+  @media (min-width: breakpoint.$s) {
+    flex-direction: row;
+  }
+}
```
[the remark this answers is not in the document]
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/627307894b1c0df409b57c58dc912946dac2c80c

gerardo-rodriguez:
@tylersticka @spaceninja I'm not sure why CI isn't running after my last commit, but this PR is ready for another review! I have addressed all feedback. 

I did [leave a question for @calebeby](https://github.com/cloudfour/cloudfour.com-patterns/pull/1679#discussion_r833585373), but I don't think that's a blocker for this PR.

src/components/subscribe/subscribe.scss
```diff
@@ -0,0 +1,55 @@
...
+
+///
+/// Subscribe component
+///
+
+.c-subscribe {
+  position: relative;
+}
+
+.c-subscribe__controls-wrapper {
```
tylersticka, line 15: Nit: Would `.c-subscribe__controls` suffice?

src/components/subscription-choices/subscription-choices.test.ts
calebeby: To me this line:
```js
await expectElementToHaveSize(form);
```
is more clear about what it is doing than these lines:
```js
({ formHeight, formWidth } = await form.evaluate(getFormDimensions));
expect(formHeight).toBeGreaterThan(1);
expect(formWidth).toBeGreaterThan(1);
```

I don't feel super strongly about this.

src/objects/input-group/input-group.twig
```diff
@@ -1,3 +1,3 @@
-<div class="o-input-group">
+<div class="{{['o-input-group', class|default('')]|join(' ')}}">
```
tylersticka, line 1: Could we simplify this a bit since this is HTML and not a template argument?

```suggestion
<div class="o-input-group {{class}}">
```

src/components/subscribe/subscribe.twig
```diff
@@ -0,0 +1,76 @@
+{% set _heading_tag = heading_tag|default('h2') %}
+
+<div class="{{['c-subscribe', 'js-subscribe', class|default('')]|join(' ')}}">
```
tylersticka, line 3: We can probably simplify HTML class attributes a bit.

```suggestion
<div class="c-subscribe js-subscribe {{class}}">
```

src/components/subscribe/subscribe.twig
```diff
@@ -0,0 +1,76 @@
...
+        ]|join(' '),
+        label: weekly_digests_btn_label|default('Get Weekly Digests'),
+        content_start_icon: weekly_digests_btn_icon|default('envelope')
+      } only %}
+    </div>
+  </div>
+
+  <form
+    id="subscribe-{{form_id}}"
+    action="{{form_action|default('https://cloudfour.us13.list-manage.com/subscribe/post?u=ce064f42c86a5982dd218d4de&amp;id=7e505a6a67')}}"
```
tylersticka, line 38: I'm not sure I agree with embedding this in the front-end pattern since it's so specific to a single project. I think it's reasonable to expect this to be provided by the host project instead.

src/components/subscribe/subscribe.scss
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/b2401fcd18300929a836d26c0c1ee2f916ecbc72

gerardo-rodriguez: Oops, this change causes a layout regression. There was already a `c-subscribe__controls` element. 🔍 👀

src/components/subscribe/subscribe.twig
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/c9e2af36d0f38ad0c46ae14e6629a96416c2ccba

src/components/subscribe/subscribe.scss
gerardo-rodriguez: I think I can remove a wrapping `<div>`. That would allow this to just be `*__controls`. Looking into it now. 🏗️

gerardo-rodriguez: Fixed! https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/6e7d0dc6b2be8524831ff595f80437753419bc30

gerardo-rodriguez:
@tylersticka Thanks for your feedback! 🎉 

I have addressed the latest round of feedback. 😉

src/components/subscription-choices/subscription-choices.test.ts
gerardo-rodriguez: @calebeby That's a fair argument.

Do I create a separate function for the `toBeLessThanOrEqual(1)` version?

One is checking:
```js
 expect(formHeight).toBeGreaterThan(1);
expect(formWidth).toBeGreaterThan(1);
```
The other is checking:
```js
expect(formHeight).toBeLessThanOrEqual(1);
expect(formWidth).toBeLessThanOrEqual(1);
```
The element is expected to have a size in both cases. (`1` or greater than one). Which this in mind, I'm not sure `expectElementToHaveSize()` would work as nicely.

Hmm... 🤔

tylersticka (commented):
LGTM, will defer approval to @cloudfour/dev since I feel less qualified to review the JS and testing stuff!

src/components/subscription-choices/subscription-choices.test.ts
gerardo-rodriguez: @calebeby I've made changes! https://github.com/cloudfour/cloudfour.com-patterns/pull/1679/commits/c9216fdc855dc8c940c79c983e280325d7b069ff

gerardo-rodriguez:
@calebeby @spaceninja Can either of you help out with a final dev review? Thank you!

spaceninja (approved):
Code looks good, and works as described :shipit:

src/components/subscribe/subscribe.test.ts
calebeby: @gerardo-rodriguez I tried to debug this locally, I logged the markup using:

```js
console.log(await page.evaluate(() => document.body.innerHTML));
```

right before the snapshot.

It looks like the `id`/`for` attributes aren't getting set correctly.

```html
<label class="c-subscribe__form-input-label" for="subscription-email-">
          Email
        </label>
        <div class="o-input-group c-subscribe__form-input-group">


      <input type="email" class="c-input c-subscribe__form-input" id="subscription-email-null" placeholder="Your
 Email Address" autocomplete="home email" required="">
```

Looks like we are missing the `form_id` prop?

gerardo-rodriguez: > I logged the markup using

Great idea! We should add this “recipe” tip somewhere in the Pleasantest docs. It wasn’t obvious I could do that. LOL

> Looks like we are missing the form_id prop?

Oh, odd! Let’s create a GH issue for this. Can you help me with that? If not, no worries, I can do so later this morning. 

Thanks for following up, @calebeby!

calebeby: https://github.com/cloudfour/cloudfour.com-patterns/pull/1689

PR #1685, opened 2022-03-23
Update master to main in docs

gerardo-rodriguez:
## Overview

Figured I'd make these quick changes. 🙂


tylersticka (commented):
@gerardo-rodriguez We were planning to do this once `v-next` is ready to be the new `main`. I'm all for this change, but since we aren't planning to introduce any more updates to the main branch right now, it feels a bit premature.

PR #1689, opened 2022-03-28
Set a default form_id for the subscribe pattern for testing

calebeby:
Conversation here: https://github.com/cloudfour/cloudfour.com-patterns/pull/1679#discussion_r836628459

@gerardo-rodriguez an alternate solution would be to make the twig file itself have its own default value for the form ID, but I don't know if that is the best because it would mean that the form id would be duplicated on the page if you included the subscribe component multiple times on the same page. Do you know if there is a way to make a twig prop required, so that it'll throw an error or something if the prop is missing?

### DIFF
--- a/src/components/subscribe/subscribe.test.ts
+++ b/src/components/subscribe/subscribe.test.ts
@@ -4,9 +4,11 @@ import { withBrowser, getAccessibilityTree } from 'pleasantest';
 import { loadTwigTemplate, loadGlobalCSS } from '../../../test-utils';
 
 // Helper to load the Twig template file
-const componentMarkup = loadTwigTemplate(
-  path.join(__dirname, './subscribe.twig')
-);
+const componentMarkup = (args = {}) =>
+  loadTwigTemplate(path.join(__dirname, './subscribe.twig'))({
+    form_id: 'example-form',
+    ...args,
+  });
 // Helper to load the demo Twig template file
 const demoMarkup = loadTwigTemplate(path.join(__dirname, './demo/demo.twig'));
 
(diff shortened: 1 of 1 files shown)


gerardo-rodriguez:
> an alternate solution would be to make the twig file itself have its own default value for the form ID, but I don't know if that is the best because it would mean that the form id would be duplicated on the page if you included the subscribe component multiple times on the same page.

@calebeby I agree, I'd be concerned the form ID would be duplicated on the page if the component existed multiple times on the same page.

> Do you know if there is a way to make a twig prop required, so that it'll throw an error or something if the prop is missing?

I am not aware, but also Twig isn't my forte so there might be something I'm not aware of.

gerardo-rodriguez (approved):
Thanks, @calebeby!

gerardo-rodriguez:
> > an alternate solution would be to make the twig file itself have its own default value for the form ID, but I don't know if that is the best because it would mean that the form id would be duplicated on the page if you included the subscribe component multiple times on the same page.
> 
> @calebeby I agree, I'd be concerned the form ID would be duplicated on the page if the component existed multiple times on the same page.
> 
> > Do you know if there is a way to make a twig prop required, so that it'll throw an error or something if the prop is missing?
> 
> I am not aware, but also Twig isn't my forte so there might be something I'm not aware of.

@calebeby Maybe you could do something with [`defined`](https://twig.symfony.com/doc/3.x/tests/defined.html)? Maybe render a message instead of the component if a given prop is not defined? 🤔

calebeby:
Oh yeah that might be helpful. IDK if it's worth it, but we could probably make a error pattern that is very visible/flashy and then go through all our patterns and make them render the error pattern if any of their required props are missing

gerardo-rodriguez:
That's an interesting idea! I can create a GH issue and we can leave it in the backlog as a nice-to-have as I imagine it'll be a lower priority through this initial phase of work.

PR #1696, opened 2022-03-31
Add supporting materials to speaking event prototype

### LINKED ISSUE #1694: Add supporting materials to Speaking Event page
The speaking event [prototype](https://v-next--cloudfour-patterns.netlify.app/iframe.html?id=prototypes-speaking-event--single-speaker) is missing an area for supporting materials. We should consider where the materials go and what they look like.

Screen Shot 2022-03-29 at 11 20 11 AM

dromo77:
## Overview

Some of the talks have links to additional resources/supporting materials. This PR updates the prototypes to include an example with one link and one with multiple.

## Screenshots

Screen Shot 2022-03-31 at 2 20 50 PM
Screen Shot 2022-03-31 at 2 20 55 PM
Screen Shot 2022-03-31 at 2 21 06 PM

## Testing

https://deploy-preview-1696--cloudfour-patterns.netlify.app/?path=/story/prototypes-speaking-event--single-speaker

---

- Fixes #1694

### DIFF
--- a/src/prototypes/speaking-event/example/multiple-speakers.twig
+++ b/src/prototypes/speaking-event/example/multiple-speakers.twig
@@ -38,7 +38,7 @@
             "class": "o-embed--wide",
             "src": deck
           } only %}
-          <div class="o-button-group">
+          <div class="o-button-group o-button-group--grow">
             {% include '@cloudfour/components/button/button.twig' with {
               "content_start_icon": "brands/github",
               "label": "GitHub"
--- a/src/prototypes/speaking-event/example/multiple-speakers.twig
+++ b/src/prototypes/speaking-event/example/multiple-speakers.twig
@@ -38,7 +38,7 @@
             "class": "o-embed--wide",
             "src": deck
           } only %}
-          <div class="button-columns">
+          <div class="o-button-group">
             {% include '@cloudfour/components/button/button.twig' with {
               "content_start_icon": "brands/github",
               "label": "GitHub"
--- a/src/prototypes/speaking-event/example/multiple-speakers.twig
+++ b/src/prototypes/speaking-event/example/multiple-speakers.twig
@@ -27,10 +27,27 @@
           <p>
             If you have a website—particularly one that generates revenue for your organization—you need a Progressive Web App. So where do you begin? How do you decide which features of a Progressive Web App make sense for your users? What tools can make the process easier (or harder)? In this practical session, Jason will guide you through the key design decisions you’ll need to make about your Progressive Web App and how those decisions impact the scope of your project. He'll also teach you how to avoid common pitfalls and help you take full advantage of Progressive Web App technology.
           </p>
+          <div class="u-pad-block-start-4">
+            {% include '@cloudfour/components/heading/heading.twig' with {
+              level: 1,
+              tag_name: 'h2',
+              content: 'Resources',
+            } only %}
+          </div>
           {% include '@cloudfour/objects/embed/demo/image.twig' with {
             "class": "o-embed--wide",
             "src": deck
           } only %}
+          <div class="button-columns">
+            {% include '@cloudfour/components/button/button.twig' with {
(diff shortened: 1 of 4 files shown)


dromo77:
> It feels a little funny to me that a single button takes up 50% of the space

I thought about making it full width on large screens, but it feels a little big. What do you think @tylersticka?

Screen Shot 2022-03-31 at 2 56 17 PM

tylersticka:
I agree that it looks a little big that way, but I think it's preferable.

src/prototypes/speaking-event/example/multiple-speakers.twig
```diff
@@ -27,10 +27,27 @@
...
+          <div class="button-columns">
+            {% include '@cloudfour/components/button/button.twig' with {
+              "content_start_icon": "brands/github",
+              "label": "GitHub"
+            } only %}
+            {% include '@cloudfour/components/button/button.twig' with {
+              "content_start_icon": "link",
+              "label": "Progressive Web Apps"
+            } only %}
+          </div>
```
tylersticka, line 50: Could you use a Button Group? https://v-next--cloudfour-patterns.netlify.app/?path=/docs/objects-button-group--grow

dromo77: Here's what it looks like using the Button Group pattern:

Screen Shot 2022-03-31 at 3 58 58 PM

tylersticka: Hmm, I'm not sure if that's intentional or not! Maybe we should make some alterations or a new option or modifier for that pattern?

dromo77: I changed the flex property so that the buttons are the same size on large screens. On small screens they retain their default size. It looks like this:

Screen Shot 2022-03-31 at 4 14 40 PM
Screen Shot 2022-03-31 at 4 14 58 PM

tylersticka: That looks nice!

tylersticka: @dromo77 Maybe! Can you coordinate with @gerardo-rodriguez to double check? This looks like his Object that he made originally, and he can tell us if the behavior you observed was intentional. He may also have an opinion on whether or not this should be baked in or a modifier since he'll know better why it was created in the first place.

dromo77: Thanks @tylersticka! @gerardo-rodriguez I'm trying to use the Button Group object to display the supporting materials on the "Talks" page. However, with the current `o-button-group--grow` modifier the buttons are different sizes. I added some custom CSS to the prototype to make the buttons the same size. I'd love your opinion, should I add a new modifier to support this behavior?

gerardo-rodriguez: @dromo77 I think you found a bug with the Button Group pattern! The intent was to have them be the same size. 🙂

dromo77: Thanks @gerardo-rodriguez! Should that be the default behavior even without the `--grow` modifier?

dromo77:
I updated `o-button-group--grow` so that the buttons are the same size. @tylersticka Would you mind taking another look?

src/prototypes/speaking-event/example/multiple-speakers.twig
gerardo-rodriguez: @dromo77 Mmm...good question. I suppose it doesn't matter as long as we have the ability to have both options? Either way, we need a modifier. So I don't have a preference. We have only _barely_ started using this pattern so if it seems we always add the grow modifier, then maybe that should be the default! 🙂

tylersticka (approved):
This works great! Love it

PR #1697, opened 2022-03-31
Update pleasantest to v2

calebeby:
Most of the changes were automatic changes from `getAccessibilityTree` changes. I also added a configuration file to set the minimum target size for clicks.


jest.setup.js
```diff
@@ -0,0 +1,5 @@
+import { configureDefaults } from 'pleasantest';
+
+configureDefaults({
+  user: { targetSize: 35 /* px */ },
```
gerardo-rodriguez, line 4: Can you add a comment to describe why this default was set? Something like:
```suggestion
  // Design intent doesn't require Level AAA conformance, 
  // but we want to keep a baseline targetSize check
  user: { targetSize: 35 /* px */ },
```

gerardo-rodriguez (approved):
Left one suggestion, LGTM otherwise! 🎉

gerardo-rodriguez (approved):

gerardo-rodriguez (approved):

PR #1713, opened 2022-04-06
Add width/height properties to Avatar component

### LINKED ISSUE #1693: The Avatar component should accept width and height props
We should add the ability to pass [`width` and `height` prop values to the Avatar component](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/components-avatar--empty#template-properties).

Something like:
```twig
{% include '@cloudfour/components/avatar/avatar.twig' with {
  src: avatar.src|resize(75,75),
  width: 75,
  height: 75,
} only %}
```
Should render:
```html
<img src="path/to/image-75x75-c-default.png" width="75" height="75" alt="">
```

tylersticka:
## Overview

Adds `width` and `height` properties to the Avatar component's Twig template. This has no impact on the visual appearance of the element (since it is using an aspect ratio hack for layout) and no performance impact (since modern browsers will use `sizes` to determine the best `srcset` image), but it does provide a means of constraining the size in case the CSS fails to load or the page is viewed using an alternative user agent or older browser.

Because Avatars are circular, you can provide either `width` or `height` and it will be used as the default for the other.

## Testing

These tests involve [viewing an Avatar story's "canvas" tab](https://deploy-preview-1713--cloudfour-patterns.netlify.app/?path=/story/components-avatar--with-image), changing its properties and observing the "HTML" tab to confirm that the markup has changed as expected.

- [x] With no `width` or `height` specified, the `` element should not have a `width` or `height` attribute.
- [x] When either the `width` _or_ the `height` is specified, both properties should be present on the `` element using whichever value was provided.
- [x] When both the `width` _and_ `height` properties are specified and are different, both properties should be present on the the `` element with their distinct values intact.
- [x] None of these properties should result in a visual change to the component.

---

- Fixes #1693

### DIFF
--- a/.changeset/gentle-grapes-camp.md
+++ b/.changeset/gentle-grapes-camp.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add width and height properties to Avatar component
--- a/src/components/avatar/avatar.stories.mdx
+++ b/src/components/avatar/avatar.stories.mdx
@@ -31,6 +31,8 @@ const avatarStory = (args) => {
     srcset: { type: { name: 'string' } },
     sizes: { type: { name: 'string' } },
     alt: { type: { name: 'string' } },
+    width: { type: { name: 'number' } },
+    height: { type: { name: 'number' } },
   }}
 />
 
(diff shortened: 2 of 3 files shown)


Paul-Hebert (approved):
LGTM and works as described. :+1:

PR #1719, opened 2022-04-07
Author avatar prop can be an object

### LINKED ISSUE #1709: The Author component should accept width, height, sizes, and srcset
Currently, there is no way to pass `width`, `height`, `sizes`, or `srcset` to the Avatar component within the Author component. We need to allow the Author component to pass those values along.

tylersticka:
## Overview

This PR modifies the Author component to allow the `avatar` property of an `authors` array item to be an object. If an object containing at least a `src`, it will be passed directly to the Avatar component. Otherwise, it will be passed as the Avatar `src` as it was before.

This will allow developers the option of preparing a more detailed `avatar` object including whatever properties Avatar supports, for example `srcset` and `sizes`.

**Note:** I also considered creating a `block` for avatars, but since the author information is provided as an array, it seemed weird to have the avatars customizable only as a block and not as data. But I'm open to feedback on that!

## Testing

1. [View the multiple authors story in Canvas view](https://deploy-preview-1719--cloudfour-patterns.netlify.app/?path=/story/components-author--multiple-authors)
2. Check the HTML tab
3. Confirm that the second image includes a `width` and `height` in addition to a `src` (other images should be the same as before)

---

- Fixes #1709

### DIFF
--- a/.changeset/beige-books-battle.md
+++ b/.changeset/beige-books-battle.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Avatar values passed to the Author component may now be an object, which will be passed directly to the Avatar component. This allows additional properties like `srcset`, `sizes`, etc. to be set.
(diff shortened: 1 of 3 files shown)


spaceninja (approved):
Looks great! Thanks for doing this!

PR #1720, opened 2022-04-07
Add heading_id property to Event Log

### LINKED ISSUE #1707: Add accessible names to articles in event logs
When reviewing the talks page in the WP site I noticed each item in an event log is an `article` without an accessible name. This makes it harder to scan the page using a screen reader.

We should consider adding `aria-labelledby` pointing to the headings in the articles.

tylersticka:
## Overview

This adds support for a `heading_id` property for individual Event Log items. When provided, it will be used as the `id` of the heading and within the `aria-labelledby` property on the containing article, which can [improve the screen reader experience](https://support.apple.com/guide/voiceover-guide/navigate-using-the-rotor-mchlp2719/web). This also makes it consistent with our Card component.

## Testing

1. [View the Event Log story](https://deploy-preview-1720--cloudfour-patterns.netlify.app/?path=/story/components-event-log--example)
2. Confirm that the markup contains example `id` and `aria-labelledby` attributes
3. Confirm that the experience is improved using Safari VoiceOver.

---

- Fixes #1707

### DIFF
--- a/.changeset/shy-fireants-mix.md
+++ b/.changeset/shy-fireants-mix.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add `heading_id` property to Event Log component
--- a/src/components/event-log/demo/events.json
+++ b/src/components/event-log/demo/events.json
@@ -1,5 +1,6 @@
 [
   {
+    "heading_id": "event-aea-2017-12-11",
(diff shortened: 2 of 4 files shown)


spaceninja (approved):
Changes seem reasonable, and work as described.

Can you make sure there's a followup wordpress ticket to update the talk template to add an ID prop?

PR #1723, opened 2022-04-08
Replace buggy Author avatar object with block

tylersticka:
## Overview

My #1719 feature addition was buggy: The check for `author.avatar.src` only works in very specific situations. This removes that check to avoid regressions.

In its place, it adds support for template blocks to this. This allows the content of the avatars to be manipulated directly. For example:

```twig
{% embed '@cloudfour/components/author.twig' with { authors: authors, date: date } only %}
  {% block avatars %}
    {% for author in authors %}
      {% include '@cloudfour/components/avatar/avatar.twig' with {
        src: 'anything you want',
        srcset: 'you got it',
        sizes: 'anything you need',
        width: 'you got it'
      } only %}
    {% endfor %}
  {% endblock %}
{% endembed %}
```

## Testing

Just to be safe, it's probably best that someone goes into our WordPress theme, pulls these changes into the `node_modules` locally, confirms there are no regressions, and tries using the block.

### DIFF
--- a/.changeset/good-ligers-arrive.md
+++ b/.changeset/good-ligers-arrive.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': major
+---
+
+Replaces the buggy Author `avatar.src` functionality with a template block
--- a/src/components/author/author.stories.mdx
+++ b/src/components/author/author.stories.mdx
@@ -13,7 +13,7 @@ const allAuthors = [
   },
   {
     name: 'Danielle Romo',
-    avatar: { src: danielleImage, width: 64 },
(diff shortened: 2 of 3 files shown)


spaceninja:
This change looks fine. I'll test it on Monday, and if it works in WP, I'll take care of bumping the pattern library and pulling this change into my PR on the WP side.

Paul-Hebert:
Code and deploy preview look good. Testing in WP

Paul-Hebert (approved):
Code looks good!

I tested:

- This works in the pattern library
- This fixes the bug in WP
- This allows us to pass in custom avatar content in WP

PR #1733, opened 2022-04-18
Fix automatic source code previews

calebeby:
Fixes #1695 

You can test patterns that have auto-generated source code previews, for example:
- https://deploy-preview-1733--cloudfour-patterns.netlify.app/?path=/docs/components-alert--basic
- https://deploy-preview-1733--cloudfour-patterns.netlify.app/?path=/docs/components-author--basic-usage

Tyler did most of the figuring out/fixing, and I just added a fallback for `storyContext.initialArgs`.

I did not fix the syntax highlighting, I don't remember how that works and I don't know how to fix it

Issue #1695, opened 2022-03-29
Component code doc example should show "embed"/"include" examples


gerardo-rodriguez:
There seems to have been a regression.

Before code example docs showed `embed`/`include` examples (image pulled from #1607):

![image](https://user-images.githubusercontent.com/459757/160716631-74fd23eb-d595-4318-ac1a-9a6436613761.png)

Instead, they now show HTML markup:

<img width="1277" alt="Screen Shot 2022-03-29 at 3 21 00 PM" src="https://user-images.githubusercontent.com/459757/160716658-59e8bde6-a1af-467d-acc6-2323e55eb739.png">

tylersticka:
I tried taking a look at this today but I'm unable to solve it completely. I thought I'd document what I found.

The first issue is that the `storyContext.getOriginal()` method we were relying on before seems to have been removed. Instead there is just a `storycontext.originalStoryFn` function. Changing that makes the Twig code samples reappear.

But there are two issues with those code samples that weren't present before.

The biggest issue is that many of them are missing arguments. What's perplexing is that when I `console.log` what the function is doing, it's being run multiple times per story, and many times the outcome of it is correct, but the correct version of the snippet with arguments is always overwritten by a version of the snippet where the arguments are `undefined`.

The smaller issue we could live with is that the Twig template code is no longer highlighted correctly. It's highlighted as if it were HTML.

I've pushed my branch up with the `console.log` intact in case anyone else wants to take a swing at fixing this: https://github.com/cloudfour/cloudfour.com-patterns/compare/fix/broken-transform-source

FYI @calebeby @cloudfour/dev

PR #1735, opened 2022-04-18
Fix Twig syntax highlighting

tylersticka:
## Overview

This is a follow-up to #1733 and restores Twig syntax highlighting to code samples. In a previous version of Storybook docs, the language of the source samples was detected automatically, but at some point this seems to have changed to use the base framework (which honestly makes some sense and is probably more performant). This simply sets the default value of the `docs.source.language` parameter to `'twig'`, which resolves the issue. This can still be overridden manually for specific stories or for code blocks with a language defined.

## Before

Screen Shot 2022-04-18 at 9 33 04 AM

## After

Screen Shot 2022-04-18 at 9 32 29 AM

## Testing

1. View deploy preview
2. Confirm that Twig code previews are correctly highlighted

---

- See #1695


calebeby (approved):
🎉

PR #1738, opened 2022-04-18
Fix Card docs performance issue

tylersticka:
## Overview

I assumed that the performance issues on the Card docs page were related to the failing syntax highlighting, but that was a poor assumption since that page is defining its own `transformSource` function. It turns out that the unique ID fallback for `heading_id` was triggering a reevaluation in Storybook, which would then generate a new unique ID, which resulted in an endless loop.

This PR replaces the fallback ID with one using the story's ID, which won't be invalidated. This makes the page usable again.

## Testing

Try viewing the docs tab for the Card component, confirm it no longer causes the browser tab to hang.


Paul-Hebert (approved):
Interesting! LGTM!

PR #1744, opened 2022-04-21
Updated Block Alignment Styles for Deck

### LINKED ISSUE #1743: Adjust alignment styles for new Deck block
The Deck block supports WordPress alignment styles `alignfull` and `alignwide` but they pull the content right to the edge of the viewport or beyond. We should adjust those styles to match `o-container__fill-pad` so the contents have a bit of breathing room against the viewport edges.

spaceninja:
## Overview

This PR updates the Deck object to better handle the WordPress block
alignment styles `alignwide` and `alignfull`. By default, these would
pull the Deck content flush with the edge of the screen and even clip
a bit in certain circumstances. To resolve this, we add some padding
to prevent the content from going full-bleed.

## Screenshots
Screen Shot 2022-04-21 at 4 00 08 PM
Screen Shot 2022-04-21 at 4 00 47 PM

## Testing

1. Check out this branch locally
1. Review the "[Block Alignment](http://localhost:6006/?path=/story/objects-deck--block-alignment)" story in the Deck page. Try changing the alignment to test both "Wide" and "Full" at several breakpoints.
    - The content should not touch the edge of the viewport
    - The content should not get cut off
    - There should be no horizontal scrollbar

Test the following browsers:

- [x] Chrome
- [x] Edge
- [x] Firefox
- [x] Safari
- [x] Mobile Safari

---

- Fixes #1743

### DIFF
--- a/.changeset/heavy-kings-arrive.md
+++ b/.changeset/heavy-kings-arrive.md
@@ -1,5 +1,5 @@
 ---
-"@cloudfour/patterns": minor
+'@cloudfour/patterns': minor
 ---
 
 Updated Block Alignment Styles for Deck
--- a/.changeset/heavy-kings-arrive.md
+++ b/.changeset/heavy-kings-arrive.md
@@ -1,0 +1,5 @@
+---
+"@cloudfour/patterns": minor
+---
+
+Updated Block Alignment Styles for Deck
--- a/.changeset/heavy-kings-arrive.md
+++ b/.changeset/heavy-kings-arrive.md
@@ -2,4 +2,4 @@
 '@cloudfour/patterns': minor
 ---
 
-Updated Block Alignment Styles for Deck
+Update the Deck object to better handle the WordPress block alignment styles `alignwide` and `alignfull`.
--- a/.changeset/heavy-kings-arrive.md
+++ b/.changeset/heavy-kings-arrive.md
@@ -2,4 +2,4 @@
 '@cloudfour/patterns': minor
 ---
 
-Update the Deck object to better handle the WordPress block alignment styles `alignwide` and `alignfull`.
+Update the Deck object to better handle the WordPress block alignment styles `alignwide` and `alignfull`. By default, these would pull the Deck content flush with the edge of the screen and even clip a bit in certain circumstances. To resolve this, we add some padding to prevent the content from going full-bleed.
--- a/src/objects/deck/deck.scss
+++ b/src/objects/deck/deck.scss
@@ -2,6 +2,7 @@
 @use '../../compiled/tokens/scss/size';
 @use '../../mixins/fluid';
 @use '../../mixins/media-query';
+@use '../container/container.scss';
 
 /**
  * 1. If horizontal items are shown at the wrong column count, they will appear
@@ -65,21 +66,12 @@
  * Block alignment adjustments
  *
  * 1. Set inline padding to match `o-container__fill-pad`.
- *    We're using the container tokens here, but ideally we'd make a
- *    token for the whole `fluid-clamp()` formula, but our token system
- *    doesn't support using Sass formulas that way. This should be okay,
- *    since we're still matching the container's tokens.
  * 2. Remove padding once `alignwide` is no longer full-bleed.
  */
 
 .o-deck.alignfull,
 .o-deck.alignwide {
-  padding-inline: fluid.fluid-clamp(
-    size.$padding-container-horizontal-min,
-    size.$padding-container-horizontal-max,
-    breakpoint.$s,
-    breakpoint.$xl
-  ); // 1
+  padding-inline: container.$fluid-pad-inline; // 1
 }
 
 .o-deck.alignwide {
--- a/src/objects/deck/deck.scss
+++ b/src/objects/deck/deck.scss
@@ -65,7 +65,8 @@
 /**
  * Block alignment adjustments
  *
- * 1. Set inline padding to match `o-container__fill-pad`.
+ * 1. Set inline padding to match `o-container__fill-pad` since we don't want the deck to touch 
+ *     the viewport edges even when it's full bleed.
  * 2. Remove padding once `alignwide` is no longer full-bleed.
  */
 
--- a/src/objects/deck/deck.scss
+++ b/src/objects/deck/deck.scss
@@ -69,34 +69,21 @@
  *    token for the whole `fluid-clamp()` formula, but our token system
  *    doesn't support using Sass formulas that way. This should be okay,
  *    since we're still matching the container's tokens.
- * 2. Adjust default `alignfull` margin to subtract padding.
- *    (see next comment for context.)
- * 3. In a perfect world, we'd just set padding here, and let the default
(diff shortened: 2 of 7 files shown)


src/objects/deck/deck.stories.mdx
```diff
@@ -3,6 +3,7 @@ import articles from './demo/articles.json';
 import articlesDemo from './demo/articles.twig';
 import alignmentDemo from './demo/alignment.twig';
 const articlesStory = (args) => articlesDemo({ items: articles, ...args });
+const alignmentStory = (args) => alignmentDemo({ items: articles, ...args });
```
spaceninja, line 6: Updating the alignment story to support args

src/objects/deck/deck.stories.mdx
```diff
@@ -90,8 +100,32 @@ By default, the `o-deck` class will use CSS Grid Layout to arrange child element
...
-  <Story name="Alignment" height="200px">
-    {(args) => alignmentDemo(args)}
+  <Story
+    name="Alignment"
+    height="200px"
+    args={{
+      columns: 3,
+      columnsBreakpoint: '@m',
+      class: 'o-deck--align-start',
+    }}
```
spaceninja, line 110: Updated the alignment story to accept args, and passing in args to preserve the layout that was previously hard-coded.

src/objects/deck/demo/alignment.twig
```diff
@@ -1,45 +1,27 @@
+{% if columns and columns > 1 and columnsBreakpoint and columnsBreakpoint != 'none' %}
```
spaceninja, line 1: This diff looks large, but it just updates this demo from being hard-coded to generating 3 items, and accepting args. There should be no practical difference in the rendered story, but it was bugging me that this was the only story the controls didn't affect.

src/objects/deck/demo/articles.twig
```diff
@@ -1,7 +1,7 @@
 {% if columns and columns > 1 and columnsBreakpoint and columnsBreakpoint != 'none' %}
-  {% set _deck_class = 'o-deck--' ~ columns ~ '-column' ~ columnsBreakpoint %}
+  {% set _deck_class = 'o-deck--' ~ columns ~ '-column' ~ columnsBreakpoint ~ ' ' ~ class ~ ' ' ~ alignment %}
 {% else %}
-  {% set _deck_class = '' %}
+  {% set _deck_class = class ~ ' ' ~ alignment %}
```
spaceninja, line 4: We had a class arg already but it wasn't being rendered. I added it and the new alignment class arg.

src/objects/deck/demo/articles.twig
```diff
@@ -1,43 +1,63 @@
+{% set _deck_class = '' %}
```
spaceninja, line 1: You'll want to use "hide whitespace" changes for this file — I wrapped the entire embed in a wrapper.

.changeset/heavy-kings-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Updated Block Alignment Styles for Deck
```
tylersticka, line 5: How/why?

Also, is the title case intentional?

src/objects/deck/deck.scss
```diff
@@ -60,3 +60,30 @@
...
 }
+
+/**
+ * Block alignment adjustments
+ *
+ * 1. Set inline padding to match `o-container__fill-pad`.
+ *    We're using the container tokens here, but ideally we'd make a
+ *    token for the whole `fluid-clamp()` formula, but our token system
+ *    doesn't support using Sass formulas that way. This should be okay,
+ *    since we're still matching the container's tokens.
```
tylersticka, line 71: You could create a mixin instead if you wanted to keep things DRY.

src/objects/deck/deck.scss
```diff
@@ -60,3 +60,30 @@
...
+    breakpoint.$s,
+    breakpoint.$xl
+  ); // 1
+}
+
+.o-deck.alignwide {
+  @media (width >= breakpoint.$l) {
+    padding-inline: 0; // 2
+  }
+}
```
tylersticka, line 89: Alternatively, you could override the typical `margin-*` behavior with the same that the `o-container__fill` and `__fill-pad` elements use. If there's a specificity issue there, you could apply these styles to the `wp-block-*` classes in our vendor CSS (to better associate this with the conflicting styles).

.changeset/heavy-kings-arrive.md
spaceninja: Changesets just pulls in the title of the PR by default. I wasn't aware we were adding more details here or had a formatting convention. I'll copy the commit message in.

src/objects/deck/deck.scss
spaceninja: Good idea, I'll do that.

src/objects/deck/deck.scss
spaceninja: I'm afraid I don't follow this comment at all. Probably just me being tired. Can we chat with voices tomorrow morning about this?

tylersticka: Re-reading my comment, I think I misunderstood why this style existed. Looking at it with fresh eyes this morning (and no migraine), I now understand that this isn't really 1:1 with `o-container__fill-pad` and that this padding override exists to fill a more unique need.

I do think in the future we may find that these `.alignwide` and `.alignfull` customizations make more sense alongside other WordPress styles (rather than with the components), but I don't feel strongly enough about that to argue more coherently. I think we're fine as-is.

I'm available to chat with voices sometime this afternoon (I have a few other tasks to take care of before a long morning meeting), but I don't think this should block your progress: I'm good with the current solution.

.changeset/heavy-kings-arrive.md
tylersticka: The case isn't as big a deal as the detail. Sometimes the PR title is adequate on its own, but it might not always be. I've been following the guidance in [the changesets documentation](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md#i-am-in-a-single-package-repository):

> While not every changeset is going to need a huge amount of detail, a good idea of what should be in a changeset is:
>
> - WHAT the change is
> - WHY the change was made
> - HOW a consumer should update their code

In this case I get the "what," I don't get the "why," and I'm not sure "how" is important.

I still consider myself a changeset novice, so stop me if I'm off base here!

spaceninja: No, that makes sense (and I'm just as new as you are!).

I'll update my changeset again to focus more on the why, and I'll put this as a topic on the next Dev team meeting, because I don't think there's clarity on the team about what makes a good changeset message as opposed to commit message.

src/objects/deck/deck.scss
spaceninja: 👍🏻

Paul-Hebert (approved):
Looks good Scott! One minor comment nit inline

Paul-Hebert (approved):
:shipit:

PR #1771, opened 2022-05-05
Add u-text-center utility

### LINKED ISSUE #1772: Add u-text-align-center CSS utility
Prototype: https://deploy-preview-1742--cloudfour-patterns.netlify.app/iframe.html?id=prototypes-case-study-listings--example&viewMode=story

The following text is centered. This utility will help with this use case. 

Screen Shot 2022-05-05 at 2 41 31 PM

gerardo-rodriguez:
## Overview

This PR adds a `u-text-center` utility CSS class.

This was introduced to help with the ["More clients we've worked with" heading](https://deploy-preview-1742--cloudfour-patterns.netlify.app/iframe.html?id=prototypes-case-study-listings--example&viewMode=story) (this is a blocker for https://github.com/cloudfour/cloudfour.com-wp/pull/682)

## Screenshots

Screen Shot 2022-05-05 at 2 32 23 PM

## Testing

1. Review https://deploy-preview-1771--cloudfour-patterns.netlify.app/?path=/docs/utilities-text--text-align-center

---

- Closes #1772

### DIFF
--- a/src/utilities/text/text.scss
+++ b/src/utilities/text/text.scss
@@ -7,3 +7,7 @@
 .u-text-action {
   color: var(--theme-color-text-action) !important;
 }
+
+.u-text-align-center {
+  text-align: center !important;
+}
--- a/src/utilities/text/text.scss
+++ b/src/utilities/text/text.scss
@@ -8,6 +8,6 @@
   color: var(--theme-color-text-action) !important;
 }
 
-.u-text-align-center {
+.u-text-center {
   text-align: center !important;
 }
--- a/.changeset/eight-roses-breathe.md
+++ b/.changeset/eight-roses-breathe.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add u-text-align-center utility CSS class
--- a/.changeset/eight-roses-breathe.md
+++ b/.changeset/eight-roses-breathe.md
@@ -2,4 +2,4 @@
 '@cloudfour/patterns': minor
 ---
 
-Add u-text-align-center utility CSS class
+Add u-text-center utility CSS class
--- a/src/utilities/text/demo/text-align.twig
+++ b/src/utilities/text/demo/text-align.twig
@@ -3,6 +3,6 @@
 } only %}
   {% block content %}
     <p>This is default text.</p>
-    <p class="u-text-align-center">This is text that has the utility class applied.</p>
+    <p class="u-text-center">This is text that has the utility class applied.</p>
   {% endblock %}
 {% endembed %}
--- a/src/utilities/text/demo/text-align.twig
+++ b/src/utilities/text/demo/text-align.twig
@@ -1,0 +1,8 @@
+{% embed '@cloudfour/objects/container/container.twig' with {
+  class: 'o-container--pad'
+} only %}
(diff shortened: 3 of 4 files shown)


gerardo-rodriguez:
FWIW, I considered adding a modifier to the Heading component but then realized headings with a subhead are treated differently since they use a flexbox layout. I took a few steps into the rabbit hole only to realize it was going to be much faster for me to just make a utility CSS class. Plus, this utility CSS class can be used across many more elements as needed.

If you all think it's better to have a Heading `--centered` modifier, that's fine. I was trying to _responsibly_ take the path of least resistance.

src/utilities/text/text.scss
```diff
@@ -7,3 +7,7 @@
 .u-text-action {
   color: var(--theme-color-text-action) !important;
 }
+
+.u-text-align-center {
```
tylersticka, line 11: Is the word `align` necessary?

tylersticka:
> If you all think it's better to have a Heading --centered modifier, that's fine. I was trying to responsibly take the path of least resistance.

The context that called for this heading is pretty unique, so I think a utility class is appropriate.

Paul-Hebert (approved):
>FWIW, I considered adding a modifier to the Heading component but then realized headings with a subhead are treated differently since they use a flexbox layout. I took a few steps into the rabbit hole only to realize it was going to be much faster for me to just make a utility CSS class. Plus, this utility CSS class can be used across many more elements as needed.
>
>If you all think it's better to have a Heading --centered modifier, that's fine. I was trying to responsibly take the path of least resistance.

This all makes sense to me. I'm assuming we already have patterns in place for the logo grid layout? (If not, then another option would be to create a new component that has the centered heading and logo grid)

Do we need a utility (or option) to make a heading `font-weight: 400`?

Approving because the code looks good and it's a reasonable utility to have. If we need additional utilities/patterns for other aspects of the design we can always circle back.

Paul-Hebert:
Just saw the Wordpress PR, discussion of heading weight, and our logo group component, so ignore my comment above. LGTM!

src/utilities/text/text.scss
gerardo-rodriguez: I have no preference. 🙃 

I will rename it to `u-text-center`. 😉

gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1771/commits/d4bb66380290358694a6c2178da3c85ada6c72c7

tylersticka (approved):

gerardo-rodriguez:
@Paul-Hebert Can I get another approval? I renamed the CSS class from `u-text-align-center` to `u-text-center`. 😉

tylersticka:
> @Paul-Hebert Can I get another approval?

@gerardo-rodriguez Oh I see how it is, my approval's not good enough for you anymore? 😉

gerardo-rodriguez:
> > @Paul-Hebert Can I get another approval?
> 
> @gerardo-rodriguez Oh I see how it is, my approval's not good enough for you anymore? 😉
> 
> 

LMAO! 🤣 

Github had not refreshed when I submitted my request. 

I'm sorry

PR #1774, opened 2022-05-06
Outdent Code Blocks

spaceninja:
## Overview

This PR updates the styles for Gutenberg Code blocks and fenced code
sections in Markdown blocks to outdent the block so the code is aligned
with the page content.

## Screenshots

Screen Shot 2022-05-06 at 10 34 24 AM

## Testing

1. Review the Code block section of the WordPress Core Blocks page.

### DIFF
--- a/src/objects/deck/deck.scss
+++ b/src/objects/deck/deck.scss
@@ -2,6 +2,7 @@
 @use '../../compiled/tokens/scss/size';
 @use '../../mixins/fluid';
 @use '../../mixins/media-query';
+@use '../../mixins/spacing';
 @use '../container/container';
 
 /**
@@ -65,14 +66,14 @@
 /**
  * Block alignment adjustments
  *
- * 1. Set inline padding to match `o-container__fill-pad` since we don't want
- *    the deck to touch the viewport edges even when it's full bleed.
+ * 1. Set inline padding since we don't want the deck to touch the viewport
+ *    edges even when it's full bleed.
  * 2. Remove padding once `alignwide` is no longer full-bleed.
  */
 
 .o-deck.alignfull,
 .o-deck.alignwide {
-  padding-inline: container.$fluid-pad-inline; // 1
+  @include spacing.fluid-padding-inline; // 1
 }
 
 .o-deck.alignwide {
--- a/src/objects/deck/deck.scss
+++ b/src/objects/deck/deck.scss
@@ -3,7 +3,6 @@
 @use '../../mixins/fluid';
 @use '../../mixins/media-query';
 @use '../../mixins/spacing';
-@use '../container/container';
 
 /**
  * 1. If horizontal items are shown at the wrong column count, they will appear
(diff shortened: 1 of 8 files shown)


src/objects/deck/deck.scss
```diff
@@ -2,7 +2,7 @@
 @use '../../compiled/tokens/scss/size';
 @use '../../mixins/fluid';
 @use '../../mixins/media-query';
-@use '../container/container.scss';
+@use '../container/container';
```
spaceninja, line 5: The changes in this file are just formatting changes I noticed were needed while copying some code.

src/vendor/wordpress/core-blocks.stories.mdx
```diff
@@ -1,5 +1,6 @@
 import { Story, Canvas, Meta, ArgsTable } from '@storybook/addon-docs';
 import blockImageDemo from './demo/image.twig';
+import blockCodeDemo from './demo/code.twig';
```
spaceninja, line 3: Switched to an included code sample for easier formatting.

src/vendor/wordpress/styles/_core-blocks.scss
```diff
@@ -9,6 +9,7 @@
 @use '../../../mixins/ms';
 @use '../../../mixins/table';
 @use '../../../mixins/theme';
+@use '../../../objects/container/container';
```
tylersticka, line 12: Won't this result in the container CSS being output twice (once in its original position, and again here)?

It might be a better idea to use design tokens to share properties.

src/objects/deck/deck.scss
tylersticka: I realize this line isn't new, but is this resulting in the container CSS being output an additional time? If so, can we use design tokens to share whatever values we're harvesting from this file instead of including the whole file?

tylersticka (commented):
We're setting the negative margin in here, which is great, but we also need to set the padding, because the margin is fluid. Without that, the code samples don't share their alignment with adjacent text.

Screen Shot 2022-05-06 at 10 49 18 AM

src/vendor/wordpress/styles/_core-blocks.scss
spaceninja: The Sass documentation says it'll only be rendered once: https://sass-lang.com/documentation/at-rules/use

And if you review the compiled file in the `/dist` folder, you'll only see it once.

That said, I think you're right. What I'm doing is violating the intent of our design tokens. I'll see if I can refactor those into tokens.

spaceninja:
@tylersticka this has been update to fix the alignment issue and switch to using mixins rather than importing the container styles directly.

tylersticka (commented):
The left text alignment works now!

src/objects/deck/deck.scss
```diff
@@ -2,7 +2,8 @@
 @use '../../compiled/tokens/scss/size';
 @use '../../mixins/fluid';
 @use '../../mixins/media-query';
-@use '../container/container.scss';
+@use '../../mixins/spacing';
+@use '../container/container';
```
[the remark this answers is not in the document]
spaceninja: Nope! Good catch.

tylersticka (approved):

PR #1802, opened 2022-05-24
Fix double scrollbar issue

### LINKED ISSUE #1754: There should not be two vertical scrollbars
In Chrome (there may be other browsers), there are two scrollbars.

If I disable either `block-size` or `overflow-x` in the [following CSS block](https://github.com/cloudfour/cloudfour.com-patterns/blob/6b55a724e9a1071a2b61e35d220d2157174ae914/src/base/_defaults.scss#L59), the double scrollbar issue goes away:

```css
html,
body {
  block-size: 100%; // 1
  overflow-x: hidden; // 2
}
```

The following pages have two scrollbars (there may be more):
- Articles (`/thinks`)
- Speaking (`/presents`)
- Our Team (`/is`)
- Individual bio pages (`/is/scott`)

Two vertical scrollbars can be seen in the following screenshot:
Screen Shot 2022-04-27 at 5 08 38 PM

calebeby:
## Overview

Closes #1754

- Fixes the double-scrollbar issue
- Preserves the "pin to bottom" behavior of the footer for short pages

In order to reproduce this issue you must set the visible scrollbars macos option:

## Testing

- Get your local wp running
- Link your `@cloudfour/patterns` in the wp site to your local copy:
  - Go to your local repo copy of `@cloudfour/patterns`
  - `git checkout fix-double-scrollbar`
  - `npm run build`
  - `npm link` - this sets up a global symlink to this copy of the repo
  - Go to your local wp folder
  - `cd app/public/wp-content/themes/cloudfour2021`
  - `npm link @cloudfour/patterns` - this sets up that folder in node_modules to be symlinked to your local copy of the repo.
- Hard-refresh pages in the browser to make sure the new CSS is read through the symlinks
- Test a short page (that the footer stays pinned to the bottom): https://cloudfour.local/contributes/ (for comparison, [on c4dev](https://c4dev.wpengine.com/contributes/))
- Test a long page (that only one scrollbar shows up) https://cloudfour.local/thinks/faster-integration-with-web-components/ (for comparison, [on c4dev](https://c4dev.wpengine.com/thinks/faster-integration-with-web-components/))
- Test more if you want

### DIFF
--- a/src/base/_defaults.scss
+++ b/src/base/_defaults.scss
@@ -47,7 +47,6 @@ ul {
   margin: 0;
 }
 
-
 /**
  * 1. Forces the viewport to be filled. Necessary for the footer to
  *    "stick" to the bottom of short pages.
@@ -62,7 +61,8 @@ html {
  *    @see https://cloudfour.com/thinks/breaking-out-with-viewport-units-and-calc/
  */
 
-html, body {
+html,
+body {
   overflow-x: hidden; // 1
 }
 
--- a/src/base/_defaults.scss
+++ b/src/base/_defaults.scss
@@ -47,13 +47,22 @@ ul {
   margin: 0;
 }
 
+
+/**
+ * 1. Forces the viewport to be filled. Necessary for the footer to
+ *    "stick" to the bottom of short pages.
+ *    @see https://css-tricks.com/couple-takes-sticky-footer/#there-is-grid
+ */
+html {
+  block-size: 100%; // 1
+}
+
 /**
  * 1. Keep full-bleed items from triggering a horizontal scrollbar.
  *    @see https://cloudfour.com/thinks/breaking-out-with-viewport-units-and-calc/
  */
 
-html,
-body {
+html, body {
   overflow-x: hidden; // 1
 }
 
--- a/src/base/_defaults.scss
+++ b/src/base/_defaults.scss
@@ -48,17 +48,13 @@ ul {
 }
 
 /**
- * 1. Forces the viewport to be filled. Necessary for the footer to
- *    "stick" to the bottom of short pages.
- *    @see https://css-tricks.com/couple-takes-sticky-footer/#there-is-grid
- * 2. Keep full-bleed items from triggering a horizontal scrollbar.
+ * 1. Keep full-bleed items from triggering a horizontal scrollbar.
  *    @see https://cloudfour.com/thinks/breaking-out-with-viewport-units-and-calc/
  */
 
 html,
 body {
-  block-size: 100%; // 1
-  overflow-x: hidden; // 2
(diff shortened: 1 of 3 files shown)


src/base/_defaults.scss
```diff
@@ -48,17 +48,13 @@ ul {
 }
 
 /**
- * 1. Forces the viewport to be filled. Necessary for the footer to
- *    "stick" to the bottom of short pages.
- *    @see https://css-tricks.com/couple-takes-sticky-footer/#there-is-grid
```
calebeby, line 53: The `block-size: 100%` code already existed in the `.o-page` class and I changed it to `100vh` in the `.o-page` class so the 100% wouldn't need to be propagated all the way to `html` here.

This fixes the double-scrollbar issue, but I'm not completely sure why it had the issue as-is.

src/objects/page/page.scss
```diff
@@ -3,16 +3,16 @@
...
+ *    & keep the footer at the bottom of that area.
+ *    @see https://css-tricks.com/couple-takes-sticky-footer/#aa-there-is-grid
  */
 
 .o-page {
   display: grid;
   grid-template-columns: minmax(0, 1fr); // 1
   grid-template-rows: 1fr auto; // 2
-  min-block-size: 100%;
+  min-block-size: 100vh;
```
tylersticka, line 15: Will this result in short pages pushing the footer beneath the browser chrome in iOS Safari?

calebeby: Oh I should check that, good point

calebeby: Yes it does. Hmm.

Off the top of your head do you know any other way to fix the double scrollbar issue? I can fiddle with it more if not.

src/base/_defaults.scss
```diff
@@ -51,14 +51,19 @@ ul {
...
+}
+
+/**
+ * 1. Keep full-bleed items from triggering a horizontal scrollbar.
  *    @see https://cloudfour.com/thinks/breaking-out-with-viewport-units-and-calc/
  */
 
 html,
 body {
-  block-size: 100%; // 1
```
calebeby, line 60: This change was needed because `body` needed to not have `block-size: 100%` because it already had `min-block-size: 100%` in the page object.

I don't know why the combination of both `min-block-size` and `block-size` (set to the same value) made it have a second scrollbar.

calebeby:
Gerardo mentioned an issue with screenshots in firefox which I was not able to reproduce so I did not attempt to fix it. If anyone else can reproduce it, let me know!

src/objects/page/page.scss
calebeby: Found a solution

spaceninja (approved):

PR #1805, opened 2022-05-25
Add Core Element Comparison Stories

### LINKED ISSUE #1522: Ensure elements in Gutenberg markdown blocks match their block equivalents
We've added custom styles for a number of Gutenberg blocks: https://v-next--cloudfour-patterns.netlify.app/?path=/docs/vendor-wordpress-core-blocks--separator#wordpress-core-blocks

Some of the elements these blocks generate can also be generated using the markdown block (like `table`s, `hr`s, `blockquote`s, code blocks, etc.)

We should ensure that when these elements are created in a markdown block we're applying the same styles to them. The markdown block has a `.wp-block-jetpack-markdown` class we can hook into like so:

```css
.wp-block-jetpack-markdown blockquote{ }
```

spaceninja:
## Overview

This PR adds a new "Core Element Comparison" set of stories to the
WordPress section. These stories compare how we style core HTML
elements added via Markdown blocks, HTML blocks, and Gutenberg blocks.
We want them to have the same (or very comparable) styles.

Next steps: I will create a followup PR for the WordPress theme to tweak a few editor styles (notably, the Blockquote block has no styles applied to it, and is indistinguishable from plain text).

## Testing

On the preview deploy, visit the new "WordPress > Core Element Comparison" page, and verify that the styles for all the elements show matches.

Known issue: The image block is center aligned by default, while standard images are not. This is fine, since most images in our blocks use alignment utility classes as needed.

---

- Fixes #1522

### DIFF
--- a/.changeset/perfect-rats-raise.md
+++ b/.changeset/perfect-rats-raise.md
@@ -1,0 +1,5 @@
+---
+"@cloudfour/patterns": patch
+---
+
+Update styles for vanilla blockquotes and tables in Markdown blocks, HTML blocks, and legacy posts to match Gutenberg block styles.
--- a/.changeset/perfect-rats-raise.md
+++ b/.changeset/perfect-rats-raise.md
@@ -1,5 +1,5 @@
 ---
-"@cloudfour/patterns": patch
+"@cloudfour/patterns": minor
 ---
 
 Update styles for vanilla blockquotes and tables in Markdown blocks, HTML blocks, and legacy posts to match Gutenberg block styles.
--- a/src/vendor/wordpress/core-element-comparison.stories.mdx
+++ b/src/vendor/wordpress/core-element-comparison.stories.mdx
@@ -1,0 +1,34 @@
+import { Story, Canvas, Meta } from '@storybook/addon-docs';
+import MarkdownDemo from './demo/compare-markdown.twig';
+import HTMLDemo from './demo/compare-html.twig';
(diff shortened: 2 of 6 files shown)


tylersticka (commented):
Did you consider handling this with one story that accepts an argument corresponding to which version of the elements to view, and making adjustments in the template based on the value of that argument?

I ask because then you could have a story argument you could switch to see the differences between patterns instead of manually comparing them by pogo-sticking in between.

spaceninja:
> Did you consider handling this with one story that accepts an argument

No, because the markup is different. The HTML and Markdown stories are virtually identical (just the class on the code block), but the Gutenberg one has many additional classes.

I could standardize into one file and just add/remove those classes based on the rendering method chosen, but that felt like overkill.

Paul-Hebert (dismissed):
This LGTM! 

I noticed one visual discrepancy. I wasn't sure if this was a known issue or not. For the Gutenberg and Markdown examples, the code block has negative inline margins, but in the HTML example it doesn't have those margins and sits flush with the surrounding content. Is that a known issue?

(I figure I don't need to worry about only some examples having syntax highlighting.)

spaceninja:
@Paul-Hebert great catch! I just pushed a change to apply the outdenting to legacy blog posts as well.

And yeah, the syntax highlighting is a known issue.

Paul-Hebert (approved):
:shipit:

PR #1806, opened 2022-05-25
Responsively grow comment reply buttons

### LINKED ISSUE #1312: Update button group pattern for comment form functionality
Currently, the button group pattern does not support the responsive behavior shown in [the comment form designs](https://github.com/cloudfour/cloudfour.com-patterns/issues/1216). It should be updated to support:

- Equal-width buttons (even if text width differs)
- On small screens (only), stretch buttons to full width

https://github.com/cloudfour/cloudfour.com-patterns/pull/1256#discussion_r650221413

calebeby:
## Overview

Closes #1312

## Screenshots

Smaller screens:
Screen Shot 2022-05-25 at 4 06 33 PM

Larger screens:
Screen Shot 2022-05-25 at 4 07 10 PM

### DIFF
--- a/src/components/comment-form/comment-form.scss
+++ b/src/components/comment-form/comment-form.scss
@@ -18,3 +18,9 @@
     grid-column: span 1;
   }
 }
+
+@media (min-width: size.$breakpoint-m) {
+  .c-comment-form__reply-buttons {
+    justify-self: start;
+  }
+}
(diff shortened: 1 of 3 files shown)


src/components/comment-form/comment-form.scss
```diff
@@ -18,3 +18,9 @@
     grid-column: span 1;
   }
 }
+
+@media (min-width: size.$breakpoint-m) {
+  .c-comment-form__reply-buttons {
```
tylersticka, line 23: This is great, but could we name it `reply-actions` so it doesn't suggest a specific element? I know, super nitpicky, it just makes it a _tad_ more general.

PR #1808, opened 2022-05-27
Improve support in IE11 and other lacking browsers

### LINKED ISSUE #1096: Optimize baseline experience (before progressive enhancement)
As of this writing, IE usage has dipped to 1.13% globally ([source](https://gs.statcounter.com/)). It no longer makes sense to use IE as the "lowest common denominator" of our site's experience, preventing us from embracing what features like CSS Grid Layout, Custom Properties, Service Workers, etc. can offer us.

That said, we also believe fundamentally in accessibility and progressive enhancement. There's no reason for our content to be inaccessible to older browsers.

So our goal should _not_ be to closely approximate the appearance or behavior of our experiences in <= IE11, but to make sure we're fencing the right CSS in `@supports` blocks and the right JavaScript behind feature checks. The experience for IE11 and earlier users should ideally be that of a modestly styled but well-formed HTML document.

(More detail and to-dos will be added once we're in a good place to start this effort and can appraise what's already acceptable versus areas of improvement.)

calebeby:
## Overview

Closes #1096 

This improves the experience of using the cloud four site in IE11 and other lacking browsers.

@tylersticka a big thing that wasn't supported was logical properties.

Ideally code like this:
```css
.something {
  inline-size: 200px;
}
```

would be transpiled like this:

```css
.something {
  width: 200px;
}
@supports (inline-size: 200px) {
  inline-size: 200px;
  width: initial;
}
```

Unfortunately, the existing [postcss plugin](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-logical) instead transforms it like this:

```css
.something {
  width: 200px;
}
```

So it can sort of downgrade the code but that's all. Even though this is less-than-ideal, I figured that this is an ok solution because it drastically improves the experience for IE11 users, fixing a whole bunch of issues with very little effort.
If in the future we decide to have some sort of translations/other language content, we could either drop this plugin or send a PR to the plugin to automatically generate `@supports` blocks.

Testing:

- Get your local wp running
- Link your `@cloudfour/patterns` in the wp site to your local copy:
  - Go to your local repo copy of `@cloudfour/patterns`
  - `git checkout browser-support-improvements`
  - `npm run build`
  - `npm link` - this sets up a global symlink to this copy of the repo
  - Go to your local wp folder
  - `cd app/public/wp-content/themes/cloudfour2021`
  - `npm link @cloudfour/patterns` - this sets up that folder in node_modules to be symlinked to your local copy of the repo.
- Hard-refresh pages in the browser to make sure the new CSS is read through the symlinks
- Test in IE11 that at least the problems listed in the issue were fixed
- Use the "live link" feature in Local to test in Opera Mini with the data savings setting set to "extreme"


tylersticka (commented):
The color fallbacks are easy to approve, that's totally fine.

I'm more torn on the dropping of logical properties. When I use the Storybook "text flow" toolbar now, nothing works the way it should. And I don't know that I'm as concerned about layout issues in IE11 as long as content is readable and not covered up by things.

Would love some additional opinions on this.

calebeby:
> I'm more torn on the dropping of logical properties. When I use the Storybook "text flow" toolbar now, nothing works the way it should. And I don't know that I'm as concerned about layout issues in IE11 as long as content is readable and not covered up by things.

I could be wrong about this, but it feels like a CSS feature that we won't use, at least for now. If it is worth the time, I can send a PR to https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-logical to generate `@supports` blocks like I mentioned in the description.

calebeby:
We could also introduce a postcss build step for cloudfour.com-wp and move the postcss-logical step there instead of here, so the output from this pattern library is still logical properties. The reason I did it directly in this PR was because (even though they have broad browser support) logical properties are still a working draft, and it was a lot easier to put the transform here than to add a css build step to the WP site.

tylersticka (commented):
Honestly, this is already a lot better just with the custom property fallbacks and the image display type.

I had a question about the additional postcss plugin, other than that we may be able to call this good enough for now.

gulpfile.js/tasks/build-sass.js
```diff
@@ -18,6 +18,10 @@ const buildSass = () =>
     .pipe(rename({ basename: 'standalone' }))
     .pipe(dest(outDir))
     .pipe(postcss([cssnano()]))
+    // Cssnano minimizes alpha colors to 8-digit hex codes, which is still stage 2 and is not supported in some browsers
+    // There didn't seem to be an easy way to disable this minification within cssnano,
+    // and it doesn't seem to work to add the postcss-color-hex-alpha plugin in the same postcss pass as cssnano
+    .pipe(postcss([require('postcss-color-hex-alpha')()]))
```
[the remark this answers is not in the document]
calebeby: > Could we disable the colormin option in cssnano? [cssnano.co/docs/optimisations/colormin](https://cssnano.co/docs/optimisations/colormin/)
I think I tried that and it didn't seem to be working.

> Or maybe we polyfill this down the line instead?

Do you mean transpile it in the WP site? Not completely sure what you mean.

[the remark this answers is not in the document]
tylersticka: > Do you mean transpile it in the WP site? Not completely sure what you mean.

Yep! Kinda like what we were discussing for logical property support.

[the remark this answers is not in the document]
tylersticka: (Also open to cssnano alternatives if it isn't working for us!)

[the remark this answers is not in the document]
calebeby: > I tried that and it didn't seem to be working.

Oops I just looked again at the configuration and realized that I was passing `colormin: false` directly to cssnano instead of to the default preset in cssnano. I can change it to do that and it works.

> Yep! Kinda like what we were discussing for logical property support.
Like I said in slack, I prefer to publish the latest fully standardized syntax, which rgba hex is still only stage 2. Since it is easy to transpile it without loss of functionality, I think it makes sense to do that transpilation here instead of in consuming repos.

.changeset/slimy-readers-try.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Improve support in IE11 and other lacking browsers
```
tylersticka, line 5:
```suggestion
Add color and other miscellaneous fixes to improve accessibility of content in older browsers
```

postcss.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
     require('postcss-inline-svg')({
       paths: ['./src/assets'],
     }),
+    require('postcss-color-hex-alpha')(),
```
tylersticka, line 6: Is this still necessary now that `colormin` is disabled?

calebeby: This makes it so that if we author CSS using 4 or 8 digit hex codes, it'll get transpiled to rgba(). I don't know for sure whether we are or aren't using that, but I think there is not much harm in having this plugin there in case we do in the future.

tylersticka: I'd rather drop the dependency, as I don't see us using hex alpha colors manually very much.

calebeby: ✅

tylersticka (approved):

PR #1823, opened 2022-06-07
Bug/cloud cover content

AriannaChau:
## Overview

This PR creates a prototype that makes the cloud cover and subscribe component play better together.

The subscribe component has the heading "Never miss an article!" hardcoded into the component itself. It isn't really necessary in this situation and there isn't a way to remove it. I went ahead and removed it in my prototype by a hacky 
```
h2 {
display: none;
}
```
but, it would probably best to remove it and/or make it optional. 

## Screenshots

If shifts down when we get to larger screens, the intent is to keep it higher (like in this screenshot) for a longer period of time. It's a bit too low otherwise. Other than that, the prototype has the kind of behavior we want. There's also another breakpoint I've made `45em` that we don't have a variable. We should have one in this instance since this was the sweet spot I found.
Screen Shot 2022-06-07 at 12 06 54 PM

## Testing

1. [Deploy link](https://deploy-preview-1823--cloudfour-patterns.netlify.app/?path=/story/prototypes-cloud-cover-subscription--example)


tylersticka:
Instead of merging this prototype into `v-next`, I think it makes sense to keep it in draft form as a point of reference for a developer to try implementation. I say that because there are some issues with the prototype that we weren't able to resolve (such as the shift on large screens that you describe.)

Does that make sense, @AriannaChau?

Does this give us enough for implementation, @gerardo-rodriguez?

AriannaChau:
So this PR is being left as a draft to serve as a point of reference as Tyler mentioned already. I've changed a couple things like we have some new breakpoints but the content also needs to stay in place for a longer period of time, and as of right now it shifts down slightly too early.

gerardo-rodriguez:
> Does this give us enough for implementation, @gerardo-rodriguez?

@tylersticka I haven't looked at this in detail yet, but I can follow up with questions as needed. I'm sure what is here is enough to get rolling. 😉 

Thanks for the PR, @AriannaChau! 😄

spaceninja:
Closed via #1855

Issue #1822, opened 2022-06-07
Cloud Cover content feels a bit small at wider sizes


tylersticka:
Text content displayed below the main heading in the Cloud Cover can feel a little small on very wide screens:

<img width="1396" alt="Screen Shot 2022-06-07 at 8 04 31 AM" src="https://user-images.githubusercontent.com/69633/172414740-fe8829dd-54f8-4081-8a12-45f5a11a2d45.png">

I wonder if it's possible to scale this up a bit more at the largest breakpoint(s)?

<img width="1398" alt="Screen Shot 2022-06-07 at 8 04 11 AM" src="https://user-images.githubusercontent.com/69633/172414808-9a08b9ad-d20d-46ff-bd27-44e19c5573db.png">

PR #1828, opened 2022-06-09
Increase font size on larger screens

AriannaChau:
## Overview

This PR increases the font size of the cloud cover content on wider screen sizes.

## Screenshots

Screen Shot 2022-06-09 at 10 27 52 AM

## Testing

1. [Deploy link](https://deploy-preview-1828--cloudfour-patterns.netlify.app/?path=/story/components-cloud-cover--content)

Fixes https://github.com/cloudfour/cloudfour.com-patterns/issues/1822
---

### DIFF
--- a/.changeset/flat-emus-remain.md
+++ b/.changeset/flat-emus-remain.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Increase font size of cloud cover content on wider screen sizes
--- a/src/components/cloud-cover/cloud-cover.scss
+++ b/src/components/cloud-cover/cloud-cover.scss
@@ -199,14 +199,12 @@ $_gap: ms.step(1);
   grid-area: content;
 
   .cloud-cover__description {
-    @media (width > breakpoint.$l) {
-      font-size: fluid.fluid-clamp(
-        1em,
-        size.$font-big,
-        breakpoint.$xs,
-        breakpoint.$l
-      );
-    }
+    font-size: fluid.fluid-clamp(
+      1em,
+      size.$font-big,
+      breakpoint.$xs,
+      breakpoint.$l
+    );
   }
 }
 
--- a/src/components/cloud-cover/cloud-cover.scss
+++ b/src/components/cloud-cover/cloud-cover.scss
@@ -1,6 +1,7 @@
 @use 'sass:math';
 @use '../../compiled/tokens/scss/breakpoint';
 @use '../../compiled/tokens/scss/color';
+@use '../../compiled/tokens/scss/size';
 @use '../../mixins/ms';
 
 /// We can't use `grid-gap` exclusively due to some containers only being present
@@ -195,6 +196,10 @@ $_gap: ms.step(1);
 
 .c-cloud-cover__content {
   grid-area: content;
+
+  @media (width > breakpoint.$l) {
+    font-size: size.$font-big;
+  }
 }
 
 /**
--- a/src/components/cloud-cover/cloud-cover.scss
+++ b/src/components/cloud-cover/cloud-cover.scss
@@ -3,6 +3,7 @@
 @use '../../compiled/tokens/scss/color';
 @use '../../compiled/tokens/scss/size';
 @use '../../mixins/ms';
+@use '../../mixins/fluid';
 
 /// We can't use `grid-gap` exclusively due to some containers only being present
 /// some of the time, so we re-use this value for `grid-gap` and for `margin`
@@ -197,8 +198,15 @@ $_gap: ms.step(1);
 .c-cloud-cover__content {
   grid-area: content;
 
-  @media (width > breakpoint.$l) {
-    font-size: size.$font-big;
+  .cloud-cover__description {
+    @media (width > breakpoint.$l) {
+      font-size: fluid.fluid-clamp(
+        1em,
+        size.$font-big,
+        breakpoint.$xs,
+        breakpoint.$l
+      );
+    }
   }
 }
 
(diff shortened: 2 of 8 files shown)


src/components/cloud-cover/cloud-cover.scss
```diff
@@ -195,6 +196,10 @@ $_gap: ms.step(1);
 
 .c-cloud-cover__content {
   grid-area: content;
+
+  @media (width > breakpoint.$l) {
+    font-size: size.$font-big;
+  }
```
tylersticka, line 202: We're trying to avoid "hard" jumps like this in the new design in favor of gradual, fluid sizing increases.

So instead of a media query like this, we would add this to the start of the `.scss` file:

```scss
@use '../../mixins/fluid';
```

And then in `.c-cloud-cover__content`, we would use:

```scss
.c-cloud-cover__content {
   // ...
    font-size: fluid.fluid-clamp(
      1em,
      size.$font-big,
      breakpoint.$xs,
      breakpoint.$l
    );
}
```

This will cause the font size to gradually increase from `1em` to the value of `size.$font-big`.

src/components/cloud-cover/demo/content.twig
```diff
@@ -5,7 +5,7 @@
     } only %}
       {% block content %}
         {% include '@cloudfour/components/heading/heading.twig' with {
-          level: -2,
+          level: -1,
```
tylersticka, line 8: This change makes me wonder if there aren't other unexpected changes that might arise from this. For example, would buttons grow unexpectedly large as well?

Maybe the font size deserves its own class to better confine it. Or even some sort of `u-text-grow` utility?

Curious if anyone in @cloudfour/dev has an opinion (especially @spaceninja since he's usually my "utility or element class" conscience).

spaceninja: In this case, if the goal is to only affect the text size of the "description" content of the Cloud Cover header, then I'd say target the change to just that element (either via utility class or BEM class).

If the goal is to increase the text size of the header _overall_, including the heading and any other elements that might appear, such as a signup form, then this approach seems reasonable to me.

As for the "utility vs element class" argument, you caught the wrong week to ask me — have just done a LOT of manual search-and-replace of utility text colors in a client project, I'm feeling a bit down on utilities at the moment. I think they're great for prototyping, but as a broad rule of thumb, I'm finding myself swinging back in favor of modular CSS approaches like BEM — so that pattern library updates only need to happen in the CSS layer, and design changes are less likely to trigger a breaking change in the markup layer. But that's probably a bigger discussion than this PR, and since we're using utilities elsewhere, I wouldn't be mad to see them used to solve this problem.

AriannaChau:
I added a class to the description so I wouldn't have to use a utility so maybe I can make everyone happy lol.

src/components/cloud-cover/cloud-cover.scss
```diff
@@ -195,6 +197,15 @@ $_gap: ms.step(1);
 
 .c-cloud-cover__content {
   grid-area: content;
+
+  .cloud-cover__description {
```
tylersticka, line 201: This class name appears to be missing its prefix.

It also appears to be nested, which should not be necessary.

src/components/cloud-cover/demo/content.twig
```diff
@@ -8,7 +8,7 @@
           level: -2,
           content: 'Our Team'
         } only %}
-        <p>
+        <p class="cloud-cover__description">
```
tylersticka, line 11: This should be a containing element in order to support multiple classes. We also may need to think about altering the Twig template so this is its own block.

I also am not sure if `description` is too specific. Should it be `__copy`? I'm sorta disappointed that we're already using `content` now... 🤔

.changeset/flat-emus-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
```
tylersticka, line 2: I believe that with @spaceninja's addition this is now a `major` change (as it will break spacing if templates don't transition to using `heading`).

src/components/cloud-cover/cloud-cover.twig
```diff
@@ -1,12 +1,26 @@
...
+          _heading_block: _heading_block,
+          _content_block: _content_block
+        } only %}
+          {% block content %}
+            <div class="c-cloud-cover__heading">
+             {{ _heading_block }}
+            </div>
+            <div class="c-cloud-cover__copy">
+             {{ _content_block }}
+            </div>
```
tylersticka, line 21: This element should probably be a rhythm object so that multiple paragraphs will still have spacing between them.

(Note that it shouldn't be necessary to set the rhythm to `condensed`, as now the rhythm object will inherit the last rhythm value automatically.)

tylersticka (commented):
A couple of comments left inilne.

Also, I'm still curious how this looks when buttons or other elements are inside of the `c-cloud-cover__content` element (instead of just paragraphs). Are they too big now or do they look nice?

src/components/cloud-cover/cloud-cover.twig
```diff
@@ -1,12 +1,30 @@
...
+          class: 'o-rhythm--condensed',
+          _heading_block: _heading_block,
+          _content_block: _content_block
+        } only %}
+          {% block content %}
+            <div class="c-cloud-cover__heading">
+             {{ _heading_block }}
+            </div>
+            <div class="c-cloud-cover__copy">
+              {% embed '@cloudfour/objects/rhythm/rhythm.twig' with {} %}
```
tylersticka, line 20: You are probably going to want this to be `with { _content_block: _content_block } only` in order to consistently have access to `_content_block` without inheriting other values.

tylersticka (approved):

PR #1834, opened 2022-06-10
Update Markup Structure for Cloud Cover

spaceninja:
Here's the changes we were trying to make:

- Cloud Cover now has both a `content` and a `heading` block
- ~Moved the rhythm block to the component, and used HTML rather than an embed because the nested blocks were causing problems (maybe @tylersticka knows how to fix that?)~
- Updated the source code in the MDX file

You'll want to [hide whitespace changes](https://github.com/cloudfour/cloudfour.com-patterns/pull/1834/files?diff=unified&w=1) when reviewing this…

### DIFF
--- a/src/components/cloud-cover/cloud-cover.twig
+++ b/src/components/cloud-cover/cloud-cover.twig
@@ -5,8 +5,14 @@
   <div class="o-container o-container--pad-inline">
     <div class="o-container__content c-cloud-cover__inner">
       <div class="c-cloud-cover__content">
-        {% block content %}
-        {% endblock %}
+        <div class="o-rhythm o-rhythm--condensed">
+          <div class="c-cloud-cover__heading">
+            {% block heading %}{% endblock %}
+          </div>
+          <div class="c-cloud-cover__copy">
+            {% block content %}{% endblock %}
+          </div>
+        </div>
       </div>
       {% if _extra_block is not empty %}
         <div class="c-cloud-cover__extra">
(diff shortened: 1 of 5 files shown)


src/components/cloud-cover/cloud-cover.twig
```diff
@@ -5,8 +5,14 @@
...
-        {% block content %}
-        {% endblock %}
+        <div class="o-rhythm o-rhythm--condensed">
+          <div class="c-cloud-cover__heading">
+            {% block heading %}{% endblock %}
+          </div>
+          <div class="c-cloud-cover__copy">
+            {% block content %}{% endblock %}
+          </div>
+        </div>
```
tylersticka, line 15: Here is a method for doing this that shouldn't require the hard-coded HTML. Also uses the function form of `block` to cut down on tag noise.

First, you want to save the block output to a variable _before_ the include. How you do this depends on whether or not you want any default content. If so, you'd probably want to do something like this:

```twig
{% set _heading_block %}
  {% block heading %}
    (Default content goes here)
  {% endblock %}
{% endset %}
```

If not, you may be able to simplify this down to a single line:

```twig
{% set _heading_block = block('heading') %}
```

Now the contents of the `heading` block exist in the `_heading_block` variable. You can use this to pass those contents between includes. For example:

```twig
{% embed '@cloudfour/objects/rhythm/rhythm.twig' with {
  _heading_block: _heading_block
} only %}
  {% block content %}
    <div class="c-cloud-cover__heading">
      {{_heading_block}}
    </div>
  {% endblock %}
{% endembed %}
```

The Cloud Cover component template is already doing this, so you should have plenty of examples. If you want others, check out Icon, Event Log, Media, Card and Feature Group.

spaceninja: That makes sense! I'll update.

spaceninja: Done!

tylersticka (approved):

PR #1847, opened 2022-06-13
Outdent Group Blocks with Backgrounds

### LINKED ISSUE #1824: Gutenberg groups with backgrounds should have consistent alignment with surrounding elements
WordPress has a rather generic block called a "group." It's useful when you want to wrap more than one element in a background color.

As an example, here is a piece of content [from our live site](https://cloudfour.com/made/champion-power-equipment/) that am attempting to recreate in Gutenberg:

Screen Shot 2022-06-07 at 1 57 10 PM

In Gutenberg, I can create a group, choose its background color, and I can add a right-aligned image block that precedes a quote block:

Screen Shot 2022-06-07 at 1 57 58 PM

In practice, though, this doesn't work very well:

Screen Shot 2022-06-07 at 1 58 54 PM

The group gets rounded corners, which is nice and consistent with our code blocks, author bios, etc. But it doesn't have any padding, and it doesn't stretch into the margins the way those elements do:

Screen Shot 2022-06-07 at 2 00 01 PM.

spaceninja:
## Overview

This PR updates the the styles for Gutenberg "group" blocks with a
background color set to apply our standard padding, plus outdent the
block a bit to keep the group content aligned with the page content.

## Screenshots

Screen Shot 2022-06-13 at 1 45 39 PM

## Testing

1. On the preview deploy, review the "Group block" story
1. Use the arguments panel to apply a background
2. Confirm that the group's content is properly aligned with the page content.

---

- Fixes #1824


src/vendor/wordpress/styles/_core-blocks.scss
```diff
@@ -95,6 +95,20 @@ $wp-button-gap: size.$spacing-gap-button-group-default;
...
+ * Styles for Gutenberg group blocks
+ *
+ * 1. For groups with a background color set, add standard block padding,
+ *    then set inline margins and padding to outdent the block out a bit,
+ *    while keeping the group content aligned with the page content.
+ */
+.wp-block-group.has-background {
+  @include spacing.fluid-margin-inline-negative; // 1
+  @include spacing.fluid-padding-inline; // 1
+  padding-block: ms.step(1); // 1
```
tylersticka, line 109: Can we use the `size.$rhythm-default` token?

spaceninja: Done!

PR #1849, opened 2022-06-13
Misc. layout adjustments

### LINKED ISSUE #1829: Misc. layout adjustments
Now that we have full pages, I would like to explore some possible sizing adjustments:

- [ ] More generous horizontal container padding at sizes before our container max width
- [ ] Slightly reduced height of Cloud Cover

I'm bundling these together because even though they apply to separate elements, their visual impact is somewhat intertwined.

tylersticka:
## Overview

Makes a handful of sizing and spacing tweaks based on actual site content. Specifically:

- Lowered the "shirt size" widths starting with `xl`. Previously, content was widening to a size that made it difficult to compose imagery without it filling most of the viewport.
- Increased the Sky Nav max width and adjusted the Sky Nav logo padding so that it won't appear indented too far right.
- Reduced the size of the "cloud space" for the Cloud Cover component so these items won't be _quite_ so tall.
- Increased the minimum size of Deck object columns. In combination with the shirt size change, this means Deck objects will naturally limit themselves to three columns, which may eliminate the need for classes within our templates.
- Increased the amount of horizontal fluid padding (both minimum and maximum). This was feeling a little tight on some pages.

I have marked this as a `major` change, as changes to things like container widths can result in regressions.

## Screenshots

Although these are visual changes, it's kind of tough to show any one screenshot within the pattern library that will be noticeable on its own.

## Testing

The best way to test this would probably be to use `npm link` somehow to try this out locally in our WordPress theme. I've been having a lot of trouble doing so successfully... when I attempt to, the fonts stop working locally. Not sure what I'm doing wrong... or maybe now that this branch exists I could somehow install this version? Open to suggestions!

---

- Fixes #1829

### DIFF
--- a/.changeset/fifty-terms-beg.md
+++ b/.changeset/fifty-terms-beg.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': major
+---
+
+Adjusted breakpoints, max-widths and padding sizes of all full-width patterns, most notably the Container object and Sky Nav component, to better compose actual site content.
--- a/.changeset/fifty-terms-beg.md
+++ b/.changeset/fifty-terms-beg.md
@@ -2,4 +2,4 @@
 '@cloudfour/patterns': major
 ---
 
-Adjusted breakpoints, max-widths and padding sizes of all full-width patterns, most notably the Container object and Sky Nav component, to better compose actual site content.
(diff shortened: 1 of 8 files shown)


spaceninja (approved):
I have not tested this against the actual wordpress site, but I reviewed the pattern library and the code changes and have a high degree of confidence that these changes are solid.

PR #1852, opened 2022-06-14
Add active styles for Card and Event Summary

### LINKED ISSUE #1831: Hover scale effects should be negated by `:active` state
For mouse users, if you hover over a button, it scales up a bit. But because that transform is negated by `:active (`:hover:not(:active)`), you get a nice zoom back down when the user clicks the mouse down:

button-fx

This is not true of other elements with similar hover behavior, specifically:

- [Event Summary](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/components-event-summary--default-story)
- [Card](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/components-card--cover-image), specifically the cover image when linked

Ideally these elements would feel consistent in terms of the `:active` state negating the `:hover` state.

tylersticka:
## Overview

Previously, Button components had distinct hover and active states. But other interactive components with scale effects did not have `:active` states.

This PR adds the same Button `:active` effects to Event Summary calendars. Because card covers are a bit larger and would be jarring, the `:hover` state is simply negated on `:active`. This should make these interactive elements feel a bit more consistent across the site.

## Screenshots

Here is how the Button looks already for `:hover`, then `:active`:

button-fx

And here's how the Event Summary and Card components look in this PR:

event-summary-fx

card-cover-fx

## Testing

Test `:hover` and `:active` states in:

1. [Event summary](https://deploy-preview-1852--cloudfour-patterns.netlify.app/?path=/story/components-event-summary--default-story)
2. [Card with cover](https://deploy-preview-1852--cloudfour-patterns.netlify.app/?path=/story/components-card--cover-image)

---

- Fixes #1831

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -262,14 +262,17 @@ $_focus-overflow: (size.$edge-large * -1);
 
 /**
  * Hover effect for when the card contains a link. The image appears to zoom in
- * a very slight amount.
+ * a very slight amount. We negate that zoom for the `:active` state for some
+ * subtle consistency with other scale interactions, such as buttons and event
+ * summary calendars, but we don't go quite a dramatic since these are larger
+ * elements that would feel a bit more jarring.
  */
 
 .c-card__cover > *,
 .c-card__cover > picture > img {
   transition: transform transition.$prompt ease.$out;
 
-  .c-card--with-link:hover & {
+  .c-card--with-link:hover:not(:active) & {
     transform: scale(1.025);
   }
 }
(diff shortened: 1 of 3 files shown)


src/components/card/card.scss
```diff
@@ -262,14 +262,17 @@ $_focus-overflow: (size.$edge-large * -1);
 
 /**
  * Hover effect for when the card contains a link. The image appears to zoom in
- * a very slight amount.
+ * a very slight amount. We negate that zoom for the `:active` state for some
+ * subtle consistency with other scale interactions, such as buttons and event
+ * summary calendars, but we don't go quite a dramatic since these are larger
```
gerardo-rodriguez, line 267:
```suggestion
 * summary calendars, but we don't go quite as dramatic since these are larger
```

spaceninja (approved):

PR #1855, opened 2022-06-14
Improve Cloud Cover with Subscribe Combo

### LINKED ISSUE #1769: Cloud Cover extra content and Subscribe component should play nicely together
The Cloud Cover component contains an "extra content" element that, according to our own documentation, is meant for subscription interface elements.

As-is, it doesn't seem to accommodate the Subscribe component very well. See this comment for screenshots: https://github.com/cloudfour/cloudfour.com-wp/pull/680#issuecomment-1116464083

Screen Shot 2022-05-03 at 12 08 53 PM

Screen Shot 2022-05-03 at 12 09 02 PM

This issue would involve altering the extra content element and the Subscribe component in some way to better polish this layout, making it feel less claustrophobic. Perhaps the design should stay single column for a bit longer, maybe more space should exist between the main content and the extra content, or maybe there's some combination of solutions that would help.

Specific tasks:

- [ ] Replace the placeholder "newsletter signup would go here" content in the [Cloud Cover extra content story](https://v-next--cloudfour-patterns.netlify.app/?path=/story/components-cloud-cover--extra-content) with the actual [Subscribe component](https://v-next--cloudfour-patterns.netlify.app/?path=/docs/components-subscribe--default-story)
- [ ] Adjust the design of the Cloud Cover extra content to better accommodate the subscribe component, particularly at "medium" viewport widths (see above screenshot). It's also acceptable to introduce new modifiers or options for the Subscribe component if that would be helpful.
- [ ] Double check that the design also works well at other sizes (narrow, wide, etc.)

spaceninja:
## Overview

This PR modifies the Subscribe and Cloud Cover components to play nicer together.

It updates the Subscribe component to make the headings optional, and tweaks the layout to switch to horizontal once there's enough space for two buttons, rather than at a fixed breakpoint.

It updates the Cloud Cover component to switch layout at a custom breakpoint (`45em`) rather than `m`, and adjusts the proportion of space between the `content` and `extra` cells to give the `extra` content a bit more room to breathe at mid-size viewports.

#### X-Small Breakpoint
xs

#### Small Breakpoint
s

#### Medium Breakpoint
m

#### Custom Cloud Cover Breakpoint (45em)
45em

#### Large Breakpoint
l

## Testing

1. Review the "Cloud Cover > Extra Content" story on the preview environment at all breakpoints.
1. Review the other Cloud Cover stories for visual regressions.
2. Review the Subscribe component story for visual regressions.

---

- Fixes #1769

### DIFF
--- a/src/components/cloud-cover/cloud-cover.scss
+++ b/src/components/cloud-cover/cloud-cover.scss
@@ -162,9 +162,7 @@ $_gap: ms.step(1);
   z-index: 1; /* 2 */
 
   /**
-   * 1. Give the `extra` content a bit more space relative to the content to
-   *    begin with, and switch back to an even split at `xl` and up.
-   * 2. At larger sizes when the scene imagery occupies the rightmost column,
+   * 1. At larger sizes when the scene imagery occupies the rightmost column,
    *    we establish rows that are half the height of the clouds to allow more
    *    overlap.
    */
@@ -176,15 +174,23 @@ $_gap: ms.step(1);
       'content extra'
       '. .'
       '. .';
-    grid-template-columns: 3fr 4fr; /* 1 */
+    grid-template-columns: repeat(2, 1fr);
     grid-template-rows:
       repeat(2, _cloud-space(0.5))
       auto
-      repeat(2, _cloud-space(0.5)); /* 2 */
+      repeat(2, _cloud-space(0.5)); /* 1 */
   }
 
-  @media (width >= breakpoint.$xl) {
-    grid-template-columns: repeat(2, 1fr); /* 1 */
+  /**
+   * Modified styles for Cloud Cover with Extra content
+   *
+   * 1. Give the `extra` content a bit more space relative to the content
+   *    until the viewport is wide enough for the default evenly split layout.
+   */
+  .c-cloud-cover--has-extra & {
+    @media (width >= $_cloud-cover-breakpoint) and (width < breakpoint.$xl) {
+      grid-template-columns: 3fr 4fr; /* 1 */
+    }
   }
 }
 
--- a/src/components/cloud-cover/cloud-cover.scss
+++ b/src/components/cloud-cover/cloud-cover.scss
@@ -5,6 +5,14 @@
 @use '../../mixins/ms';
 @use '../../mixins/fluid';
 
+/// Setting the breakpoint to `m` (40em) results in an awkward layout
+/// when using the Subscribe form in the `extra` slot. However, setting
+/// the breakpoint to `l` (60em) results in the transition happening too
+/// late. As a result, we're using a custom breakpoint optimized for the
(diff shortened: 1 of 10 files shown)


src/components/subscribe/subscribe.scss
```diff
@@ -12,14 +12,20 @@
   position: relative;
 }
 
+.c-subscribe__heading {
+  margin-block-end: size.$rhythm-default-rem;
```
spaceninja, line 16: I removed the margins from the controls and the form because they were messing up the layout when there's no heading.

I moved the margin here to the heading instead, and changed it to rem units so the layout wouldn't change due to the heading's larger font size.

spaceninja:
Tests are failing, but I think that's not this PR's fault — they're also failing on `v-next`. I'll be looking into that separately.

tylersticka (commented):
Looks good overall, some comments left inline. Also looks like some markup-related tests need updating.

src/components/cloud-cover/cloud-cover.scss
```diff
@@ -172,6 +180,18 @@ $_gap: ms.step(1);
...
       repeat(2, _cloud-space(0.5)); /* 1 */
   }
+
+  /**
+   * Modified styles for Cloud Cover with Extra content
+   *
+   * 1. Give the `extra` content a bit more space relative to the content
+   *    until the viewport is wide enough for the default evenly split layout.
+   */
+  .c-cloud-cover--has-extra & {
```
tylersticka, line 190: While I think I _prefer_ `has-extra` in light of the `:has` selector, this goes against the convention of using `--with-` currently used by the Card and Heading components. For now we should probably make this `--with-extra` for consistency.

src/components/cloud-cover/cloud-cover.twig
```diff
@@ -3,7 +3,7 @@
 {% set _extra_block %}{% block extra %}{% endblock %}{% endset %}
 {% set _scene_block %}{% block scene %}{% endblock %}{% endset %}
 
-<div class="c-cloud-cover t-dark{% if class %} {{ class }}{% endif %}">
+<div class="c-cloud-cover t-dark{% if class %} {{ class }}{% endif %}{% if _extra_block is not empty %} c-cloud-cover--has-extra{% endif %}">
```
tylersticka, line 6: See earlier comment about `has` versus `with`

tylersticka (commented):
Looks great! Small class name suggestion inline.

tylersticka (approved):

gerardo-rodriguez:
> ~Tests are failing, but I think that's not this PR's fault — they're also failing on `v-next`. I'll be looking into that separately.~ (resolved)

@spaceninja What was the resolution?

gerardo-rodriguez (approved):
Code changes LGTM! 👍

PR #1858, opened 2022-06-16
Icon title feature + related bug fixes

### LINKED ISSUE #1856: Default fallback icon does not exist
On line 1 of our Icon component's Twig template, we define the fallback icon as `close`.

That icon key no longer exists. It should be `x` instead.

If you include the icon template without a `name`, an error will occur instead of the intended behavior of displaying the `x` icon.

### LINKED ISSUE #1857: Icon component should have method of specifying alt text
While attempting to create a `c4_icon` shortcode for our new WordPress theme, I realized I couldn't actually pass anything to the SVG template's `before` block because it would have to bust through the icon component first. Ideally the component could have a simplified way of specifying `alt` text so this wouldn't be necessary.

tylersticka:
## Overview

Adds an optional `title` property to the Icon component.

If omitted, icons will now have their `aria_hidden` property set to `"true"`.

If provided, the contents of `title` will be prepended to the SVG contents as `<title>{{title}}</title>` and the `role` will be set to `"img"`. (I started to also add support for specifying a title ID, but that apparently was addressing a Chrome-specific bug that was fixed [more than six years ago](https://bugs.chromium.org/p/chromium/issues/detail?id=566252).)

Based on this, I was able to update a lot of our existing icon usage to no longer specify `aria_hidden` manually. Along the way, I found and fixed a few small bugs:

- The Alert component's icons were setting `aria-hidden` to `1` (easy mistake to make with booleans instead of strings, a good example of why this PR's feature will be helpful)
- Some older iconographic buttons were not using the `content_start_icon` or `content_end_icon` properties, resulting in icons that looked a bit off compared to other buttons
- The default icon fallback used an outdated name

## Screenshots

Screen Shot 2022-06-16 at 1 25 16 PM

## Testing

On [the deploy preview](https://deploy-preview-1858--cloudfour-patterns.netlify.app/)...

1. In the canvas view of the icon component, confirm that accessibility tests are passing. Also confirm using the HTML tab that decorative icons display with `aria-hidden="true"` while non-decorative icons use `role="img"` and include a `title` element.
2. In the canvas view of the Icon ▸ Basic story, delete the value of `name` in the "Controls" tab. The icon should change from its previous value to `x`.
3. Confirm using the HTML tab of the canvas view of each Alert story that any icons have `aria-hidden="true"` (not `aria-hidden="1"`).
4. Confirm using the HTML tab of the canvas view of each Ground Nav story that the button icon is now within a `c-button__extra` element (rather than being included alongside label content).
5. Open [the accessibility example story](https://deploy-preview-1858--cloudfour-patterns.netlify.app/iframe.html?id=components-icon--accessibility&args=&viewMode=story) in Safari and toggle VoiceOver on. Confirm that the text is read aloud...
    - The first line's icon should read aloud as "love" or "love image."
    - The second line should completely ignore the icon.

---

- Fixes #1857
- Fixes #1856

### DIFF
--- a/src/components/ground-nav/ground-nav.twig
+++ b/src/components/ground-nav/ground-nav.twig
@@ -18,23 +18,11 @@
         {{ lead_in }}
       </h2>
       <p class="c-ground-nav__action-title">
-        {% embed '@cloudfour/components/button/button.twig' with {
+        {% include '@cloudfour/components/button/button.twig' with {
           href: link,
           label: title,
-          _ground_nav_icon: icon
+          content_start_icon: icon ?? 'link'
         } only %}
-          {% block content %}
-            {% if _ground_nav_icon %}
-              {% include '@cloudfour/components/icon/icon.twig' with {
-                name: _ground_nav_icon,
-                fallback: 'link',
-                inline: true,
-                focusable: 'false'
-              } only %}
-            {% endif %}
-            {{ label }}
-          {% endblock %}
-        {% endembed %}
       </p>
       {% include '@cloudfour/assets/illustrations/portland.svg.twig' with {
         class: 'c-ground-nav__action-illustration'
--- a/.changeset/cuddly-poems-hang.md
+++ b/.changeset/cuddly-poems-hang.md
@@ -2,4 +2,4 @@
 '@cloudfour/patterns': minor
 ---
 
-Add new `title` property to Icon component to make fallback text for semantic icons easier to specify.
+Add new `title` property to Icon component to make semantic fallback text much more straightforward
--- a/.changeset/cuddly-poems-hang.md
+++ b/.changeset/cuddly-poems-hang.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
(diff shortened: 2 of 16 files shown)


src/components/ground-nav/ground-nav.twig
```diff
@@ -18,23 +18,11 @@
         {{ lead_in }}
       </h2>
       <p class="c-ground-nav__action-title">
-        {% embed '@cloudfour/components/button/button.twig' with {
+        {% include '@cloudfour/components/button/button.twig' with {
           href: link,
           label: title,
-          _ground_nav_icon: icon
+          content_start_icon: icon ?? 'link'
```
tylersticka, line 24: I recently learned about Twig's "null-coalescing" operator, which seems preferable to `default` in many cases. If this were written as `icon|default('link')`, you would not be able to override it later with `content_start_icon: false`, because `default` will trigger. But using `icon ?? 'link'` says to use `icon` unless it is undefined or null.

gerardo-rodriguez (approved):

PR #1866, opened 2022-06-21
Fix rhythm heading cascade issue

tylersticka:
## Overview

Our rhythm object supports setting different rhythm for headings than other elements. Previously, this was done by setting a default `--rhythm-headings` property that was set to `var(--rhythm)`. But this actually only cascades as deep as the initial `--rhythm` declaration, which means if you were to set `--rhythm` later on (like in a modifier), the headings won't update in lockstep.

This PR changes the approach. Instead of declaring a default `--rhythm-headings` value, we use `--rhythm` as the fallback in the heading styles (`var(--rhythm-headings, var(--rhythm))`). This works as intended, where headings always default to the value of `--rhythm` _unless_ `--rhythm-headings` is explicitly declared.

## Screenshots

Note how the "more sections below" heading does not have consistent rhythm with the sections before the change:

Before | After
--- | ---
Screen Shot 2022-06-21 at 9 41 50 AM | Screen Shot 2022-06-21 at 9 42 08 AM

## Testing

On [the deploy preview](https://deploy-preview-1866--cloudfour-patterns.netlify.app/)...

1. View [the generous headings story](https://deploy-preview-1866--cloudfour-patterns.netlify.app/?path=/story/objects-rhythm--generous-headings). Confirm that the headings have more top margin than other elements.
2. View [the lavish story](https://deploy-preview-1866--cloudfour-patterns.netlify.app/?path=/story/objects-rhythm--lavish). Confirm that the margin above the "more sections" heading is the same as the space between sections.

### DIFF
--- a/.changeset/cuddly-dryers-build.md
+++ b/.changeset/cuddly-dryers-build.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Fix headings in Rhythm objects not respecting cascade of `--rhythm` custom property
--- a/src/mixins/_spacing.scss
+++ b/src/mixins/_spacing.scss
@@ -71,7 +71,7 @@ $fluid-spacing-inline-negative: fluid.fluid-clamp(
     > * + h5,
     > * + h6,
     > * + [class*='heading'] {
-      margin-block-start: var(--rhythm-headings);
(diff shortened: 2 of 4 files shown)


spaceninja (approved):
Code looks good, and works as described.

PR #1868, opened 2022-06-22
Improve rhythm object and its docs

tylersticka:
## Overview

I thought #1866 would un-block a WordPress task I was working on, but I realized once I started actually implementing it that what I _really_ needed was a `--lavish-headings` modifier. While attempting to add that, I got a bit tangled up in this component's existing documentation, so I decided to simplify the examples down to two demos. Along the way I also added `amount` and `heading_amount` properties so we don't have to use string interpolation to format class names.

## Screenshots

Screenshot 2022-06-22 at 10-22-52 objects-rhythm--example

## Testing

On [the deploy preview](https://deploy-preview-1868--cloudfour-patterns.netlify.app/)…

1. Proof-read [the revised rhythm docs](https://deploy-preview-1868--cloudfour-patterns.netlify.app/?path=/docs/objects-rhythm--example).
1. View [the default example story](https://deploy-preview-1868--cloudfour-patterns.netlify.app/?path=/story/objects-rhythm--example), mess with the controls and observe that the outcome works as described.
1. View [the nesting story](https://deploy-preview-1868--cloudfour-patterns.netlify.app/?path=/story/objects-rhythm--nesting), mess with the controls and observe that the outcome works as described.


src/objects/rhythm/rhythm.stories.mdx
```diff
@@ -8,132 +9,125 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs';
...
+The vertical spacing applied by the Rhythm object only applies to direct children. In some cases, you may want elements below to inherit rhythm. This can be accomplished in two main ways.
 
-The `o-rhythm--generous-headings` modifier helps differentiate sections of articles (or similarly long prose content) by adding more whitespace before headings (or elements with a class name containing `heading`, such as [our heading component](/docs/components-heading--levels#heading)).
+### Nesting Rhythm Objects
 
-<!--
-This example isn't inlined so any media queries affecting the container will
-behave as expected.
--->
+Nested rhythm objects will inherit any spacing set via modifiers. In this example, the first list inherits the parent element's rhythm while the second resets it using the `default` amount:
```
gerardo-rodriguez, line 117: > In this example, the first list inherits the parent element's rhythm while the second resets it using the default amount:

I only see one list. Am I misunderstanding? 🤔

src/objects/rhythm/rhythm.stories.mdx
```diff
@@ -8,132 +9,125 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs';
...
-    {sectionsDemo}
+    {(args) => demo(args)}
   </Story>
 </Canvas>
 
-To revert to the default spacing in nested rhythm objects, use the `o-rhythm--default` modifier.
+## Inheriting Rhythm
 
-## Heading Modifiers
+The vertical spacing applied by the Rhythm object only applies to direct children. In some cases, you may want elements below to inherit rhythm. This can be accomplished in two main ways.
```
gerardo-rodriguez, line 113:
```suggestion
The vertical spacing applied by the Rhythm object only applies to direct children. In some cases, you may want elements below to inherit rhythm. This can be accomplished in two main ways, by nesting Rhythm objects or via Sass.
```
And then below, you can change the heading "Via Sass" to only "Sass".

Up to you on this one. I wanted the transition to connect a bit more so I made this suggestion. 🙂

gerardo-rodriguez (approved):
Left minor bits of feedback. LGTM otherwise!

src/objects/rhythm/rhythm.stories.mdx
tylersticka: Ah, good catch. At the last minute I realized that the second list in the example made it seem more complicated than it was, so I removed it, but forgot to update the preceding copy.

```suggestion
Nested rhythm objects will inherit any spacing set via modifiers. In this example, the unordered list is also a Rhythm object. It inherits the parent object's spacing even though no modifier is set:
```

gerardo-rodriguez (approved):
🚢

PR #1879, opened 2022-06-27
Add is-loading and aria-disabled to Button

### LINKED ISSUE #1873: Add loading state support to buttons
There is ongoing discussion in our WordPress project of the need for an intermediate/indeterminate state for a button that may need a moment to take effect.

I provided a simple mockup of what that loading state could be like: https://codepen.io/tylersticka/pen/NWyQPVq

It seems like we should make this a supported state for our buttons?

tylersticka:
## Overview

- Adds a `aria_disabled` property to our Button component so the `aria-disabled` attribute may be set
- Extends the existing `:disabled` selector to also include `[aria-disabled="true"]`
- Adds styles for an `is-loading` state class to our button mixin

**Note:** This took longer than I thought because I fell down a rabbit hole. I initially wanted to add a `loading` boolean property to the Twig template, but then I noticed that `is-slashed` didn't work that way, then as I was untangling that I noticed some issues with the existing Storybook args, and eventually I found myself halfway into a complete refactor of the stories and template. This PR is an attempt to harvest _just_ the user-facing changes from that work instead.

## Screenshots

Animated loading state:

c-button-is-loading

Loading state when the user prefers reduced motion:

Screen Shot 2022-06-27 at 11 41 34 AM

Docs:

Screen Shot 2022-06-27 at 11 45 27 AM

## Testing

[On the deploy preview...](https://deploy-preview-1879--cloudfour-patterns.netlify.app/?path=/docs/components-button--disabled)

- Observe that the `disabled` and `aria-disabled` stories are visually the same.
- Observe that the `is-loading` example animates as expected.
- Opt into preferred reduce motion (located in the Accessibility settings of your Mac's System Preferences), observe that the `is-loading` example is no longer animated and appears dashed instead (see screenshots).

---

- Fixes #1873

### DIFF
--- a/.changeset/cuddly-birds-march.md
+++ b/.changeset/cuddly-birds-march.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add `is-loading` state class to Buttons
--- a/.changeset/rare-wasps-fly.md
+++ b/.changeset/rare-wasps-fly.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add support for `aria-disabled` attribute to Buttons
--- a/src/base/_animation.scss
+++ b/src/base/_animation.scss
@@ -1,0 +1,5 @@
+@keyframes rotate-clockwise {
+  to {
+    transform: rotate(360deg);
+  }
+}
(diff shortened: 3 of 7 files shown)


gerardo-rodriguez (approved):
Nice, love it! 🎉

PR #1882, opened 2022-06-27
Subscribe component: Better `destroy()`, add `init()`

### LINKED ISSUE #1885: Add a `init()` method, refactor `destroy()` method for Subscribe component
To better integrate with OneSignal push notifications, the Subscribe component needs a new `init()` method as well as a refactored `destroy()` method.

You should be able to destroy/reinitialize programmatically. This will allow OneSignal UI logic to update the UI state of the Subscribe component depending on OneSignal's availability or the user's notifications permission state.

gerardo-rodriguez:
## Overview

This PR updates the Subscribe component:
- adds a ~`reinit()`~ `init()` method
- expands the `destroy()` method to better fit its use case

These updates will allow this component to better integrate with the OneSignal push notifications feature providing a better UX.

### Initializing

~Works the same as before (either through the initial `initSubscribe()` or the `reinit()` method):~
Per [PR feedback below](https://github.com/cloudfour/cloudfour.com-patterns/pull/1882#discussion_r910158353), you now initialize explicitly:
```js
// Initialize
const subscribe = createSubscribe(element);
subscribe.init();
```
Functionality should be the same:
- You can tab through to reveal the form, tabbing backwards will reveal the CTA buttons
- If you `blur` away from the form, it will reveal the CTA buttons after a short delay
- Hitting <kbd>Escape</kbd> will visually hide the form and reveal the CTA buttons

initialized

### Destroy

When calling the `destroy()` method:
- The CTA buttons are removed (you cannot tab back into the CTA buttons UI)
- The form is revealed
- No <kbd>Escape</kbd> key event listener
- No `blur` event listener

destroyed

## Testing

Preview: https://deploy-preview-1882--cloudfour-patterns.netlify.app/iframe.html?args=&id=components-subscribe--default-story&viewMode=story

### Code review

Hide whitespace for an easier diff review.

### Destroyed

1. Click the "**Destroy Subscribe component**" button
2. Tab through the UI and confirm:
    - [ ] You cannot access the "Get notifications"/"Get Weekly Digests" CTA buttons
    - [ ] The form is not hidden after you remove focus from it
    - [ ] The form does not hide when you hit the <kbd>Escape</kbd> key
 
### Initialized

1. Click the "**Initialize Subscribe component**" button
2. Tab through the UI and confirm:
    - [ ] You can tab back and forth between the CTA buttons UI and the form
    - [ ] The form will visually hide after a delay when you move `:focus` away from it
    - [ ] The form will visually hide immediately if you hit the <kbd>Escape</kbd> key, your `:focus` should go back to the "Get Weekly Digests" button
    - [ ] Clicking "Get Weekly Digest" reveals the form and puts the focus on the input field

### Toggle `destroy()`/`init()`

- [ ] Try to break it by toggling quickly, tabbing and toggling quickly, clicking and toggling quickly, etc.

---

- Closes #1885


src/components/subscribe/subscribe.scss
```diff
@@ -43,7 +45,10 @@
   inline-size: 100%;
   inset-block-start: 0;
   inset-inline-start: 0;
-  position: absolute;
+
+  .c-subscribe:not(.is-destroyed) & {
+    position: absolute; // 2
+  }
```
tylersticka, line 51: Could this be simplified to `&:not(:only-child)` so it doesn't require the state class?

src/components/subscribe/subscribe.stories.mdx
```diff
@@ -224,14 +242,21 @@ The Subscribe component `.js-subscribe` root element.
 
 ### Return value
 
-An object with a `destroy` function that removes all event listeners added by this component.
+An object with two functions to programmatically control the Subscribe component's
+state:
 
 ```js
 {
+  reinit: () => void
```
tylersticka, line 250: This might just be me, but I find the concept of `reinit` really confusing.

I think in the past I've seen patterns like this...

```js
// Instantiate the component
const subscribe = SubscribeComponent(element);
// Destroy later
subscribe.destroy();
// Re-initialize
subscribe.init();
```

Sometimes they require an _initial_ `init` call, or sometimes they make the `init` call return the main object so you can use a shorthand like this:

```js
const subscribe = SubscribeComponent(element).init();
subscribe.destroy();
```

I believe the benefit of that approach is that you can optionally set up the thing in one step and then get it kicked off in another. But I couldn't say for sure.

The main reason I find it more intuitive is that `init` and `destroy` basically mean `on` and `off`. Wanna activate the component? That's `init`. Wanna deactivate it? That's `destroy`.

I don't know where `reinit` fits in there. It's a weird term, and there's no `init` method for me to compare it to.

I dunno. Maybe I've just got 4pm brain or something.

src/components/subscribe/subscribe.ts
```diff
@@ -23,10 +23,15 @@ export const initSubscribe = (containerEl: HTMLElement) => {
   );
   const formEl = containerEl.querySelector('form');
   const formFocusableEls = containerEl.querySelectorAll('label, input, button');
-  const controlEls = containerEl.querySelectorAll('.js-subscribe__control');
+  const controlEls = containerEl.querySelectorAll<HTMLElement>(
+    '.js-subscribe__control'
+  );
+  const controlsUiWrapper = containerEl.querySelector<HTMLElement>(
+    '.js-c-subscribe__controls-ui'
```
tylersticka, line 30: Why does one of these class names include a `-c-` in it?

gerardo-rodriguez: Oops, I will remove the `c`. 👍🏽

src/components/subscribe/subscribe.stories.mdx
gerardo-rodriguez: Yeah, I actually had `init`/`destroy` but then I thought it might be odd that it initializes automatically when you run:
```js
const subscribe = initSubscribe(element);
```
So I thought, since it's already initialized, your options from there are to destroy it or reinitialize it. But, I'm fine switching it back from `reinit` to `init` ([`reinit` calls `init` under the hood anyways](https://github.com/cloudfour/cloudfour.com-patterns/pull/1882/files#diff-2cf35c939f52cd7cb90f679aef3a3bdb9d78236a9588414e3b49f41a006f218dR144)). 👍🏽 

Are you also suggesting the following change from:
```js
export const initSubscribe = (containerEl: HTMLElement) => { ... }
```
to
```js
export const SubscribeComponent = (containerEl: HTMLElement) => { ... }
```

gerardo-rodriguez: Additional info...

[Here is where `init()` is automatically called](https://github.com/cloudfour/cloudfour.com-patterns/blob/c1df9aff13616bfc0720ff01829a100645085b4e/src/components/subscribe/subscribe.ts#L101) when you call `initSubscribe`.

And regarding only exporting the single `init*()` function, I was following [existing patterns](https://github.com/cloudfour/cloudfour.com-patterns/blob/c1df9aff13616bfc0720ff01829a100645085b4e/src/components/sky-nav/sky-nav.ts#L14) already set up in the repo. 

I don't mind switching to this pattern:
```js
// Instantiate the component
const subscribe = SubscribeComponent(element);
// Destroy later
subscribe.destroy();
// Re-initialize
subscribe.init();
```
But we should probably do a sweep of all JS components so that there is consistency across all of their APIs. Is it safe to assume `SubscribeComponent()` does not run `init()`?

tylersticka: Ultimately the most important thing is that you and your @cloudfour/dev collaborators are all on board. I just wanted to poke at this because sometimes in a team it can be tough for others to know what is or isn't up for discussion if there isn't someone swooping and pooping on certain details from time to time. 😅

calebeby: I agree that the implicit initialization being a separate thing than `reinit` might be a little confusing.

Can we call the first function `createSubscribe`?

```
// Instantiate the component
const subscribe = createSubscribe(element);
subscribe.init();

// Destroy later
subscribe.destroy();
// Re-initialize
subscribe.init();
```

tylersticka: One more thought...

> But we should probably do a sweep of all JS components so that there is consistency across all of their APIs.

In our client work, we often have to think about the best structure and consistency for the customer. Often that involves accommodating some work they've already done to that point. In this case, _we're_ the client. Any conventions we're following are our own, and we should feel empowered to change them, or to break consistency for one component and make an issue to refactor the others later, or whatever makes sense for our values and workflow.

I have a feeling that in this case, @calebeby's suggestion makes a lot of sense. I'm not suggesting that we need to balloon this task into a full-on refactor, just making sure folks remember the power we have in this particular space.

calebeby: I think using the existing API pattern in other components, if we wanted to be able destroy them and initialize them again would be like this:

```js
let something = initializeWhatever(element);
// ...later
something.destroy();
// ...later re-initialize
something = initializeWhatever(element);
```

But, I don't think for other components there is really the use-case for wanting to destroy and reinitialize them multiple times.

For the subscribe component, uninitializing and reinitializing is more of a common workflow. I would be fine with either way, calling the top-level method multiple times to reinitialize, or calling a returned `init` method each time we want to initialize.

Because the usage pattern of this component is pretty different from other components, I don't think there is a strong case that the API needs to be exactly consistent. I don't think seeing the `init` function on just this one component would be too out of place.

spaceninja: > But we should probably do a sweep of all JS components so that there is consistency across all of their APIs.

> I'm not suggesting that we need to balloon this task into a full-on refactor, just making sure folks remember the power we have in this particular space.

If you'll pardon me jumping in, I think this is a point where I have some useful experience working on product teams for longer-term products.

When a team discovers a better way to do things they've been doing, they need to avoid falling into either of these traps:

- Sunk-cost fallacy: We can't change this now, we've already done it this way in too many places.
- Burn it all down: If we change it here, we have to change it everywhere.

No one likes it when we change a standard or best practice and then the code base has a split of files using two techniques. That said, I've participated in several large-scale refactoring efforts of long-term projects. In these teams, there simply isn't a feasable way to refactor all existing code at once, because the effort required would block all ability to do any other work. But you can't be unwilling to change best practices, either. In those teams, it's really common to communicate changes out to the team, and make it clear "this is how we do this moving forward, and if you touch old code, you should take the time to update it."

Ironically, because this codebase is our baby, it muddies the waters. When it's a client project, we usually have a firm deadline or budget driving us, so it's clear when a refactoring effort would blow our scope. But since this is our project on our deadline, it's harder to make a decision.

Thankfully, in this case, I would argue we _do_ have a deadline. We have a mandate from Megan to burn towards releasing the new site as soon as possible. For me, that says clearly that we should not attempt to go back and refactor old code, but we should note and share the new standard with the team, apply that standard to all PRs moving forward (including this one), and file tickets (which can be prioritized after launch) for refactoring certain pain points that we may already be aware of.

Apologies if everyone was already on the same page, but this was buzzing around my head last night and I wanted to share it with the team.

gerardo-rodriguez: > Can we call the first function `createSubscribe`?

This works for me!

> Because the usage pattern of this component is pretty different from other components, I don't think there is a strong case that the API needs to be exactly consistent. I don't think seeing the init function on just this one component would be too out of place.

Yep, I agree. This is how I started feeling about this overnight as well. 

I'm going to:
- rename `initSubscribe` to `createSubscribe`
- remove the implicit initilization
- change `reinit` to `init`

gerardo-rodriguez:
Flipping this PR back into `draft` status so I can [address feedback](https://github.com/cloudfour/cloudfour.com-patterns/pull/1882#discussion_r910158353).

src/components/subscribe/subscribe.ts
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1882/commits/ca09368787159d05ef6b5f45f6890b324be8da9a

gerardo-rodriguez:
@cloudfour/dev @tylersticka Ready for another review, thank you! 🎉

src/components/subscribe/subscribe.scss
```diff
@@ -43,7 +45,10 @@
   inline-size: 100%;
   inset-block-start: 0;
   inset-inline-start: 0;
-  position: absolute;
+
+  .c-subscribe:not(.is-destroyed) & {
```
tylersticka, line 49: So I'm guessing `&:not(:only-child)` did not work?

src/components/subscribe/subscribe.stories.mdx
```diff
@@ -172,6 +183,14 @@ The Subscribe component provides the UI to subscribe to notifications and/or ema
 
 This component embeds the [Button Swap component](/?path=/docs/components-button-swap--default-story) for the "Get Notifications" button.
 
+## Destroy/Initialize
+
+The Subscribe component allows you to programmatically disable/enable the JavaScript functionality via its `destroy()` and `init()` methods. When "destroyed", the following will occur:
```
tylersticka, line 188:
```suggestion
The Subscribe component allows you to programmatically disable/enable the JavaScript functionality via its `destroy()` and `init()` methods. When "destroyed," the following will occur:
```

src/components/subscribe/subscribe.stories.mdx
```diff
@@ -172,6 +183,14 @@ The Subscribe component provides the UI to subscribe to notifications and/or ema
...
+
+- The "Get notifcations"/"Get Weekly Digests" UI will be removed
```
tylersticka, line 190:
```suggestion
- The "Get notifications"/"Get Weekly Digests" UI will be removed
```

src/components/subscribe/subscribe.ts
```diff
@@ -8,8 +8,9 @@
  *   Subscribe component elements
  * - The form can be hidden by pressing the `Escape` key
  */
-export const initSubscribe = (containerEl: HTMLElement) => {
+export const createSubscribe = (containerEl: HTMLElement) => {
   const SHOW_FORM_CLASS = 'activate-form';
+  const DESTROYED_STATE_CLASS = 'is-destroyed';
```
tylersticka, line 13: I think part of the reason I was wondering if `:only-child` works or not is that I don't really like this class name.

The class name relates to the JavaScript... the component was destroyed. But the HTML elements and their styles remain, so `is-destroyed` feels strange there. Is it really supposed to be `is-inactive`? Or should it be `is-enhanced` when `is-destroyed` is _not_ true?

Something's sticking in my craw about it right now. Of course, if `:only-child` worked for that selector it might be a completely moot point!

src/components/subscribe/subscribe.scss
gerardo-rodriguez: @tylersticka I _resolved_ this conversation prematurely. Sorry!

src/components/subscribe/subscribe.ts
gerardo-rodriguez: @tylersticka Fair points!

Since the `is-destroyed` CSS state class has a 1:1 relationship with the `destroy()` method, does that mean you are suggesting something other than "destroy" for the method name as well? 🔍 👀

src/components/subscribe/subscribe.scss
tylersticka: @gerardo-rodriguez I did not, thanks!

If the previous sibling is `hidden`, then you should be able to use some form of sibling selector?

```scss
.c-subscribe__controls-ui:not([hidden]) ~ & {
  position: absolute;
}
```

gerardo-rodriguez: Oh, good idea, thanks @tylersticka! Let me try it. 👍🏽

gerardo-rodriguez: @tylersticka Genius, this should work! 🎉

gerardo-rodriguez: Yay, thanks, @tylersticka! 

https://github.com/cloudfour/cloudfour.com-patterns/pull/1882/commits/864b7de01a8e5ced1d29158f4b08dca97e6ee027

src/components/subscribe/subscribe.ts
gerardo-rodriguez: Update: [It's moot](https://github.com/cloudfour/cloudfour.com-patterns/pull/1882#discussion_r910309872)! 🕺🏽

src/components/subscribe/subscribe.test.ts
```diff
@@ -50,13 +53,15 @@ const expectElementNotToBeVisuallyHidden = async (
...
   utils.runJS(`
-    import { initSubscribe } from './subscribe'
-    export default () => initSubscribe(
+    import { createSubscribe } from './subscribe';
+    const subscribe = createSubscribe(
       document.querySelector('.js-subscribe')
-    )
+    );
+    subscribe.init();
+    export default () => subscribe;
```
calebeby, line 61: I realize this was exported before, but it is not clear to me why. I think this line can be removed?

.changeset/gold-shirts-yell.md
```diff
@@ -0,0 +1,6 @@
+---
+'@cloudfour/patterns': minor
```
calebeby, line 2: This is a major change now, the name of the exported function changed, and now it requires calling `init()`. Could you copy-paste one of the code examples from the docs into the changeset briefly showing the new usage?

src/components/subscribe/subscribe.test.ts
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1882/commits/4aca16a04d345f8024851c764dd2d22b9fa63105

.changeset/gold-shirts-yell.md
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/1882/commits/7eeac0072f15509ec5b509a340d3019cfc80157c and https://github.com/cloudfour/cloudfour.com-patterns/pull/1882/commits/6921adc1fe6861aed2b6c075a0ebef469194ef51

gerardo-rodriguez:
@cloudfour/dev @tylersticka Ready for another review, thanks!

calebeby (approved):

gerardo-rodriguez:
@tylersticka You provided some great PR feedback, would you like to review it again, or is @calebeby's approval sufficient? Thanks!

PR #1917, opened 2022-07-07
Remove button swap JS

### LINKED ISSUE #1899: Remove Button Swap JS
The Button Swap [JS](https://github.com/cloudfour/cloudfour.com-patterns/blob/v-next/src/components/button-swap/button-swap.ts) can be deleted as it's [not being used in the implementation](https://github.com/cloudfour/cloudfour.com-wp/pull/760#discussion_r912202335).

The Button Swap [tests](https://github.com/cloudfour/cloudfour.com-patterns/blob/v-next/src/components/button-swap/button-swap.test.ts ) will likely need review as well. Some of them will no longer apply (e.g. `'Callback functions are called'`).

Also, make sure to remove/update the Button Swap documentation.

calebeby:
Closes #1899


calebeby:
@gerardo-rodriguez In making this change I realized that when we reimplemented this on the WP side, we did not reimplement the [visually hidden status message functionality](https://github.com/cloudfour/cloudfour.com-patterns/blob/v-next/src/components/button-swap/button-swap.ts#L30-L32) that was in the button-swap JS. Is that something we should reimplement on the WP side?

gerardo-rodriguez:
> > @gerardo-rodriguez In making this change I realized that when we reimplemented this on the WP side, we did not reimplement the [visually hidden status message functionality](https://github.com/cloudfour/cloudfour.com-patterns/blob/v-next/src/components/button-swap/button-swap.ts#L30-L32) that was in the button-swap JS. Is that something we should reimplement on the WP side?
> 
> @calebeby Good catch! Maybe? I'm trying to get the push notifications to work in Safari on my local and I can't I wanted to see what the UX was like. 🤔

@calebeby Can you add a GH issue to follow up on your question? I can't get Safari to work locally or in production. I wonder if something is up with OneSignal + Safari? 🤔

gerardo-rodriguez (approved):

PR #1936, opened 2022-07-12
Add `is_sectioning` check to Card and Overview

### LINKED ISSUE #1934: Do not use `header` or `footer` in Cards using `div`
By default the Card component renders as an `article` element. The Card's header and footer sections are rendered using `header` and `footer` elements.

However, if you change the Card's element to a `div` it causes the VO Rotor to display a list of unlabelled banner landmarks. That's because the banner role is removed from `header` elements when they're a descendent of an `article` element.

To fix this, we just need to add a bit of conditional logic — if the `tag_name` is `article` or `section`, then `header_tag_name` should be `header` and `footer_tag_name` should be `footer`. If `tag_name` is anything else (such as `div`) then `header_tag_name` and `footer_tag_name` should both be `div`.

Note the same change should be made to the Overview object template.

spaceninja:
## Overview

This PR updates the template for the Card component and Overview object
to check if `tag_name` is a sectioning element like `article` or
`section`, and then change the default for `header_tag_element` and
`footer_tag_element` to either `header`/`footer` or `div`.

This avoids an issue when `header` elements are rendered in a `div`,
it causes useless "banner" landmarks to display in the VO Rotor.

## Testing

Confirm tests pass on this PR.

Review the Card and Overview stories on the preview deploy. There should be no visual changes.

---

- Fixes #1934

### DIFF
--- a/src/components/card/card.test.ts
+++ b/src/components/card/card.test.ts
@@ -1,0 +1,57 @@
+import path from 'path';
+
+import type { ElementHandle } from 'pleasantest';
+import { getAccessibilityTree, withBrowser } from 'pleasantest';
+
+import { loadTwigTemplate } from '../../../test-utils.js';
+
+/** Helper to load the Twig template file */
+const template = loadTwigTemplate(path.join(__dirname, './demo/single.twig'));
+const divTemplate = loadTwigTemplate(path.join(__dirname, './demo/div.twig'));
+
+describe('Card component', () => {
+  test(
+    'should use header/footer with article',
+    withBrowser(async ({ utils, page }) => {
+      await utils.injectHTML(
+        await template({
+          show_heading: true,
+          show_footer: true,
+        })
+      );
+
+      const body = await page.evaluateHandle<ElementHandle>(
+        () => document.body
+      );
+      expect(await getAccessibilityTree(body)).toMatchInlineSnapshot(`
+        article
+          banner
+            heading "Lorem ipsum dolor sit amet" (level=2)
+              text "Lorem ipsum dolor sit amet"
+          contentinfo
+            text "Jul 12, 2022"
+      `);
+    })
+  );
+
+  test(
+    'should not use header/footer with div',
+    withBrowser(async ({ utils, page }) => {
+      await utils.injectHTML(
+        await divTemplate({
+          show_heading: true,
+          show_footer: true,
+        })
+      );
+
+      const body = await page.evaluateHandle<ElementHandle>(
+        () => document.body
+      );
+      expect(await getAccessibilityTree(body)).toMatchInlineSnapshot(`
+        heading "Lorem ipsum dolor sit amet" (level=2)
+          text "Lorem ipsum dolor sit amet"
+        text "Jul 12, 2022"
+      `);
+    })
+  );
+});
--- a/src/components/card/card.test.ts
+++ b/src/components/card/card.test.ts
@@ -45,10 +45,8 @@ describe('Card component', () => {
       const body = await page.evaluateHandle<ElementHandle>(
         () => document.body
       );
-      expect(await getAccessibilityTree(body)).toMatchInlineSnapshot(`
+      expect(await getAccessibilityTree(body, {includeText: false})).toMatchInlineSnapshot(`
         heading "Lorem ipsum dolor sit amet" (level=2)
-          text "Lorem ipsum dolor sit amet"
-        text "Jul 12, 2022"
       `);
     })
   );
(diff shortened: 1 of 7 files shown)


src/components/card/card.twig
```diff
@@ -1,32 +1,35 @@
-{% set tag_name = tag_name|default('article') %}
-{% set header_tag_name = header_tag_name|default('header') %}
-{% set footer_tag_name = footer_tag_name|default('footer') %}
+{% set _tag_name = tag_name|default('article') %}
+{% set is_sectioning = _tag_name in ['article', 'section'] %}
+{% set default_header_tag = is_sectioning ? 'header' : 'div' %}
+{% set default_footer_tag = is_sectioning ? 'footer' : 'div' %}
```
tylersticka, line 4: This is a total nit, but in general, if I don't expect developers to use a property via `include` or `embed`, I lead its name with an underscore. Just helps me sort out which properties are "private" to the template versus which ones I expect users to actually manipulate.

src/components/card/card.test.ts
```diff
@@ -0,0 +1,57 @@
...
+      const body = await page.evaluateHandle<ElementHandle>(
+        () => document.body
+      );
+      expect(await getAccessibilityTree(body)).toMatchInlineSnapshot(`
+        article
+          banner
+            heading "Lorem ipsum dolor sit amet" (level=2)
+              text "Lorem ipsum dolor sit amet"
+          contentinfo
+            text "Jul 12, 2022"
```
Paul-Hebert, line 32: Total nit, but we could simplify this test output by using `{includeText: false}`
```suggestion
      expect(await getAccessibilityTree(body, {includeText: false})).toMatchInlineSnapshot(`
        article
          banner
            heading "Lorem ipsum dolor sit amet" (level=2)
          contentinfo
```

src/components/card/card.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        })
+      );
+
+      const body = await page.evaluateHandle<ElementHandle>(
+        () => document.body
+      );
+      expect(await getAccessibilityTree(body)).toMatchInlineSnapshot(`
+        heading "Lorem ipsum dolor sit amet" (level=2)
+          text "Lorem ipsum dolor sit amet"
+        text "Jul 12, 2022"
```
Paul-Hebert, line 53: Same nit as above

```suggestion
      expect(await getAccessibilityTree(body, {includeText: false})).toMatchInlineSnapshot(`
        heading "Lorem ipsum dolor sit amet" (level=2)
```

src/components/card/card.twig
```diff
@@ -1,32 +1,35 @@
-{% set tag_name = tag_name|default('article') %}
-{% set header_tag_name = header_tag_name|default('header') %}
-{% set footer_tag_name = footer_tag_name|default('footer') %}
+{% set _tag_name = tag_name|default('article') %}
+{% set _is_sectioning = _tag_name in ['article', 'section'] %}
+{% set _default_header_tag = _is_sectioning ? 'header' : 'div' %}
+{% set _default_footer_tag = _is_sectioning ? 'footer' : 'div' %}
```
Paul-Hebert, line 4: Can we add some explanatory comments here about what we're doing?

src/objects/overview/demo/div.twig
```diff
@@ -0,0 +1,11 @@
+{% embed '@cloudfour/objects/overview/overview.twig' with { overview_tag: 'div' } %}
```
gerardo-rodriguez, line 1: @spaceninja Glad this worked! Can you add a comment that this is used in tests? 👍🏽

tylersticka (approved):

gerardo-rodriguez (approved):
Nice work, @spaceninja! 🎉

Paul-Hebert (approved):
Overall this looks good. Left a few nits inline. 

Out of curiosity are there use cases where we're using Overview as something other than a `section`? Or are we just making this change to be safe?

spaceninja:
> are there use cases where we're using Overview as something other than a section?

Yeah, it was part of the changes in https://github.com/cloudfour/cloudfour.com-wp/pull/778

Paul-Hebert:
>>are there use cases where we're using Overview as something other than a section?
>
>Yeah, it was part of the changes in https://github.com/cloudfour/cloudfour.com-wp/pull/778

Cool! Carry on then :+1:

Paul-Hebert (approved):
🚢

gerardo-rodriguez (approved):

PR #1962, opened 2022-07-18
Remove deprecated components and outdated prototypes + assets

tylersticka:
## Overview

Hey look, it's a cleanup PR!

- Removes the Event Summary component, which was deprecated by the arrival of Media Summary and removed from our WordPress theme earlier today.
- Corrects some errors in our Button Swap docs: The CSS pattern is _not_ deprecated, and the references to JavaScript function calls were invalid since it no longer includes any JavaScript!
- Updates the Event Log docs to link to Media Summary.
- Removing Event Summary broke a few prototypes, but those prototypes had fallen into disrepair for one reason or another. Since I was cleaning up stuff anyway, I went ahead and removed most of them and any assets that were orphaned by their removal.
- In one case I replaced one remaining team member avatar with a default avatar just because open sourcing likenesses kinda creeps me out and it felt weird to single one person out. Duplicated the default avatar in the static directory isn't very DRY, but I didn't want this to snowball into an asset import task… I figure we can all live with the addition since this PR otherwise makes 4,000+ deletions.
- I noticed a small formatting issue in our changelog and fixed it.

## Testing

Make sure that tests pass and my changes to documentation make sense


calebeby (approved):
🎉 🚮

PR #1976, opened 2022-07-25
Replace `u-inline-block` with `u-rescue-orphan`, include `text-decoration` inheritance

tylersticka:
## Overview

The `u-inline-block` utility was added for a "rescue orphan" function, but it failed to accomplish this very well within links because it would disrupt `text-decoration`. ~This adds a soft default using the `:where` selector so if _anything_ else is setting `text-decoration` on the element, it will take precedence, but otherwise it will inherit from the parent.~ This renames the utility to `u-rescue-orphan` and adds `text-decoration: inherit`.

## Screenshots

Before | After
--- | ---
Screen Shot 2022-07-25 at 2 21 59 PM | Screen Shot 2022-07-25 at 2 22 11 PM

## Testing

[Deploy preview](https://deploy-preview-1976--cloudfour-patterns.netlify.app/?path=/story/utilities-display--rescue-orphan)

### DIFF
--- a/.changeset/giant-otters-grab.md
+++ b/.changeset/giant-otters-grab.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Prevent `u-inline-block` from unintentionally disrupting `text-decoration` styles
--- a/.changeset/giant-otters-grab.md
+++ b/.changeset/giant-otters-grab.md
@@ -1,5 +1,0 @@
----
-'@cloudfour/patterns': patch
----
-
-Prevent `u-inline-block` from unintentionally disrupting `text-decoration` styles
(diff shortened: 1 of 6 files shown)


calebeby:
@tylersticka Can we set text-decoration:inherit (and maybe rename u-inline-block to something more specific to this use case)?

tylersticka:
@calebeby Any specific suggestions?

calebeby:
Maybe `u-rescue-orphans`?

tylersticka:
@calebeby I went with singular `u-rescue-orphan` as I wanted it to be clear that this doesn't apply to a parent element or magically fix all orphans.

calebeby (approved):

PR #1982, opened 2022-07-26
Add `role="text"` to Author `span` elements

### LINKED ISSUE #1981: Author component: Use `role="text"` for `span` elements
Without `role="text"`, the `span` elements currently used for the visually hidden prefix and the individual author names may be read aloud as a separate element, which may impact the readability of this pattern.

---

Observed by @gerardo-rodriguez

tylersticka:
## Overview

> Add `role="text"` to `span` elements inside of the Author component, increasing the likelihood that its contents will be read aloud seamlessly

## Testing

[On deploy preview](https://deploy-preview-1982--cloudfour-patterns.netlify.app/?path=/docs/components-author--basic-usage), test VO experience

---

- Fixes #1981

### DIFF
--- a/src/components/author/author.twig
+++ b/src/components/author/author.twig
@@ -68,8 +68,8 @@
       </span>
       {% block authors %}
         {% for author in authors %}
-          {% if loop.last and loop.length > 1 %}and {% endif %}
-          {%- if author.link and not unlink -%}
+          {% if loop.last and loop.length > 1 %}and{% endif %}
+          {% if author.link and not unlink -%}
             <a class="c-author__author" href="{{ author.link }}">{{ author.name }}</a>
           {%- else -%}
             <span class="c-author__author" role="text">{{ author.name }}</span>
--- a/src/components/author/author.twig
+++ b/src/components/author/author.twig
@@ -12,7 +12,7 @@
     {% elseif date %}
       <p>
         {# This creates a more accessible, and less confusing, user experience #}
-        <span class="u-hidden-visually">
+        <span class="u-hidden-visually" role="text">
           {{ date_prefix|default('Published on') }} {{ date|date('F jS, Y') }}
         </span>
         {#
@@ -63,14 +63,16 @@
   <div class="c-author__content-container o-media__content">
     {# Author links #}
     <p>
-      <span class="u-hidden-visually">{{ author_prefix|default("By") }}</span>
+      <span class="u-hidden-visually" role="text">
+        {{ author_prefix|default("By") }}
+      </span>
       {% block authors %}
         {% for author in authors %}
           {% if loop.last and loop.length > 1 %}and {% endif %}
           {%- if author.link and not unlink -%}
             <a class="c-author__author" href="{{ author.link }}">{{ author.name }}</a>
           {%- else -%}
-            <span class="c-author__author">{{ author.name }}</span>
+            <span class="c-author__author" role="text">{{ author.name }}</span>
           {%- endif -%}
           {%- if not loop.last and loop.length > 2 %},{% endif %}
         {% endfor %}
--- a/src/components/author/author.twig
+++ b/src/components/author/author.twig
@@ -12,7 +12,7 @@
     {% elseif date %}
       <p>
         {# This creates a more accessible, and less confusing, user experience #}
-        <span class="u-hidden-visually" role="text">
+        <span class="u-hidden-visually">
           {{ date_prefix|default('Published on') }} {{ date|date('F jS, Y') }}
         </span>
         {#
@@ -63,6 +63,18 @@
   <div class="c-author__content-container o-media__content">
     {# Author links #}
     <p>
+      {#
+        We're adding `role="text"` to these elements to avoid a Safari issue
+        where the text won't be read together as intended. This may be an
+        overreach: `role="text"` is non-standard, and there's an argument to be
+        made that this contradicts the expected VoiceOver user experience. In
+        this case, it feels reasonable because the spans exist to enhance the
+        visual experience and do not add any valuable meaningful delineation
+        between elements. But maybe later we'll change our mind!
+
+        @see https://www.tpgi.com/using-the-text-role/
+        @see https://github.com/dequelabs/axe-core/issues/1597
+      #}
       <span class="u-hidden-visually" role="text">
         {{ author_prefix|default("By") }}
       </span>
--- a/.changeset/tall-boxes-invent.md
+++ b/.changeset/tall-boxes-invent.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Add `role="text"` to `span` elements inside of the Author component, increasing the likelihood that its contents will be read aloud seamlessly
--- a/src/components/author/author.test.ts
+++ b/src/components/author/author.test.ts
@@ -33,8 +33,7 @@ test(
         text "By"
       link "Shakira Isabel Mebarak Ripoll"
         text "Shakira Isabel Mebarak Ripoll"
-      text
-        text "Presented on March 31st, 2021"
+      text "Presented on March 31st, 2021"
     `);
   })
 );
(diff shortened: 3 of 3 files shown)


tylersticka:
@gerardo-rodriguez @calebeby I know tests are failing, but I've requested a review because the accessibility tree comparison has me a little... I dunno... worried this is the wrong move? The tree _before_ this change makes more sense to me than the tree after, which seems to have nested `text` elements?

So I guess I'd just love another pair of 👀 making sure that I'm not missing something critical here. If I just need to update the test, that's fine, I'm just skittish now.

calebeby:
Hmm. I agree that the accessibility tree looks a little funny in the test. I tested this PR in VO and it definitely feels like an improvement.

The pleasantest accessibility tree feature uses `text` to denote text that is not associated with an element's `name`. It'll also use `text` to denote the `text` role. I _think_ there is code to avoid printing text if it is already displayed as an element's name, but I'm not sure. I'll look into this more, but for now I'd just let Jest update the test even though it looks funny.

calebeby:
@tylersticka can we add a space between "by" and the author's name? In VO it was reading as ByAuthorfirstname Authorlastname for ones without links. (This is an issue on `main` too but I figure it is worth fixing here while we're at it).

tylersticka:
> can we add a space between "by" and the author's name?

Will spaces actually solve the problem? There are already newlines between the elements, which traditionally get collapsed into spaces in HTML.

calebeby:
@tylersticka I think there are not newlines in the HTML because the twig tags have the dashes which eat the whitespace

calebeby:
> @tylersticka I think there are not newlines in the HTML because the twig tags have the dashes which eat the whitespace

Nevermind, I see that there are some spaces not being eaten by twig. Hmm, I wonder why it is collapsing the words 🤔

tylersticka:
@calebeby I just tweaked the dashes and locally in Safari it's being read as I'd expect it to. Will push that along with a text update soon.

src/components/author/author.twig
```diff
@@ -63,14 +63,16 @@
...
+        {{ author_prefix|default("By") }}
+      </span>
       {% block authors %}
         {% for author in authors %}
           {% if loop.last and loop.length > 1 %}and {% endif %}
           {%- if author.link and not unlink -%}
             <a class="c-author__author" href="{{ author.link }}">{{ author.name }}</a>
           {%- else -%}
-            <span class="c-author__author">{{ author.name }}</span>
+            <span class="c-author__author" role="text">{{ author.name }}</span>
```
calebeby, line 75: Hmm, if we add a space here it reads "by" and the first name as two separate words:

```suggestion
            <span class="c-author__author" role="text"> {{ author.name }}</span>
```

But, it also messes with it visually, so I'm not sure this is what we want

src/components/author/author.twig
```diff
@@ -12,7 +12,7 @@
     {% elseif date %}
       <p>
         {# This creates a more accessible, and less confusing, user experience #}
-        <span class="u-hidden-visually">
+        <span class="u-hidden-visually" role="text">
```
gerardo-rodriguez, line 15: We shouldn't need the addition of `role="text"` here since it's one `span` that contains the full text. 😉 
```suggestion
        <span class="u-hidden-visually">
```

tylersticka: @gerardo-rodriguez Can you elaborate on what you mean by "contains the full text?"

calebeby:
It looks like role="text" is nonstandard. It provides a usability improvement for VO+safari users, so I'm not opposed to using it if we don't have a better option, but just thought I would mention that it was nonstandard in case you didn't know.

Draft spec: https://w3c.github.io/aria/
Published spec: https://www.w3.org/TR/wai-aria/
More details on role="text": https://www.tpgi.com/using-the-text-role/

src/components/author/author.twig
gerardo-rodriguez: @tylersticka Sure, sorry for not being clear. 😬 
```twig
{# 
  All of the text that is expected to be read together is within a single span element, so we 
  don't need to add the role="text" workaround to have it read as one whole unit of text 
#}
<span class="u-hidden-visually">
  {{ date_prefix|default('Published on') }} {{ date|date('F jS, Y') }}
</span>
```

gerardo-rodriguez: Am I being clear? I can try again if I still didn't explain myself well. 😅

tylersticka:
Hmm, this paragraph is really standing out to me:

> To a developer or user who is used to the way JAWS or NVDA announces content, the behavior of VO might seem fragmented and annoying, but to someone who regularly uses VO, this pattern of speech output is usual, and may be the expected or even preferred pattern.

@gerardo-rodriguez I fear we may be overstepping to bring VO usability in line with other tools. I'm going to shift this back to "draft" for now and consider this a bit further.

gerardo-rodriguez:
> It looks like role="text" is nonstandard. It provides a usability improvement for VO+safari users, so I'm not opposed to using it if we don't have a better option, but just thought I would mention that it was nonstandard in case you didn't know.

I was just reading this thread: https://github.com/dequelabs/axe-core/issues/1597

gerardo-rodriguez:
> Hmm, this paragraph is really standing out to me:
> 
> > To a developer or user who is used to the way JAWS or NVDA announces content, the behavior of VO might seem fragmented and annoying, but to someone who regularly uses VO, this pattern of speech output is usual, and may be the expected or even preferred pattern.
> 
> @gerardo-rodriguez I fear we may be overstepping to bring VO usability in line with other tools. I'm going to shift this back to "draft" for now and consider this a bit further.

@tylersticka Totally fair! We haven't had the opportunity to have actual user testing so we are making _a lot_ of assumptions to try and provide what we think is the best UX based on our best intentions.

src/components/author/author.twig
tylersticka: Okay, I think I get it. If the element is an "only child," there's no _gap_ between elements the user has to manually traverse. If there were an additional `span` after this one, then there would be.

Thank you for the explanation.

gerardo-rodriguez:
> I fear we may be overstepping to bring VO usability in line with other tools. 

Though, it _could_ be argued that if it leads to a better end-user experience, it's worth it. (Priority of constituencies?) Why not push the status quo?

But, I guess again, I'm making an assumption that a VO (macOS or iOS) end-user prefers not needing to have VO read/navigate the broken-up text. 😕

tylersticka:
@gerardo-rodriguez @calebeby After re-reading both threads and playing more with the VO experience, I think this is a valid use of `role="text"`.

It's not a "home run" sort of decision. Clearly, Safari has decided that these `span` elements are worthy of separate consideration. If a user has chosen VoiceOver as their assistive device, then surely they must be used to that sort of distinction.

I just think in this case it's very clearly not useful. The reason the `span` elements exist is entirely visual: Hiding text that's unnecessary in a larger visual context and preventing line breaks. They are not _emphasizing_ content, they are not grouping it in a meaningful way… it's purely presentational.

That said, from a development standpoint this seems pretty ridiculous. I think there's a limit to how many attributes we should add to every single adjacent `span` that we could consider reasonable. But in this case, we're talking about two `span` elements in a template that seem to make the experience better. I think we should continue to judge this sort of thing on a case by case basis.

Per @gerardo-rodriguez's earlier comment, I've removed the unnecessary `role` from the `time` fallback text (and updated the related test to match). I've also added a large comment with this context preceding the part of the template with `role="text"`.

If you disagree with this conclusion, let me know and we can chat more about it with voices. Thanks for bearing with me on this!

gerardo-rodriguez:
To add to the conversation. 🙃 

Screen Shot 2022-07-26 at 4 05 45 PM

Sorry to drag you along this journey. 
- I was navigating the site with VO
- I was annoyed when text was split up when it was read to me
- I wondered if I could do anything about that
- I found this: https://axesslab.com/text-splitting/
- It worked
- I shared the possibility of potentially enhancing the UX
- And here we are 

I appreciate ya'll digging more.

I still don't know what the "right" answer is. I guess, if I step back, there is no right answer because we are talking about UX. It's subjective and as @tylersticka mentioned, should be judged on a case by case basis.

tylersticka:
If Adrian says we shouldn't do this, then I'm going to go with what he says. 😅

PR #2022, opened 2022-08-24
Improved Comment Styles

### LINKED ISSUE #1997: Code blocks in comments have too much negative margin
Code blocks inside of comments seem to be extending too far. In this screenshot I'm comparing the edges of the short bio in [this article](https://cloudfour.com/thinks/css-circles/) with the edges of the comment:

Screen Shot 2022-08-03 at 11 53 49 AM

At small sizes this causes the code block to extend past the viewport edge:

Screen Shot 2022-08-03 at 11 56 26 AM

I'm open to ideas on how to adjust for this context.

(Thanks @spaceninja for spotting this one.)

### LINKED ISSUE #2021: Tables in comments are unstyled
We should apply our default table styles to tables in comments.

spaceninja:
## Overview

This PR updates some of the core block styles to better match base
element styles between posts and comments. Specifically:

1. Applying table styles to tables in comments.
2. Removing outdenting from code blocks in comments on legacy posts.

## Screenshots

Screen Shot 2022-08-24 at 4 39 47 PM

## Testing

1. On the preview deploy, go to the [Vendor > WordPress > Core Element Comparison > Comments](https://deploy-preview-2022--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-core-element-comparison--comments) page.
1. Confirm that the table has the base table styles applied (border under the `thead`)
1. Confirm that the code block is _not_ outdented.
1. View source, add the `legacy-post` class to the `#root` element.
1. Confirm that the code block is still not outdented. (previously, `legacy-post` would do this)

---

- Fixes #1997
- Fixes #2021

### DIFF
--- a/src/vendor/wordpress/demo/compare-comment.twig
+++ b/src/vendor/wordpress/demo/compare-comment.twig
@@ -1,0 +1,76 @@
+<div
+  class="demo-container c-comment"
+  style="max-width: 40rem; margin: 0 auto; padding: 0 1.25rem"
+>
+  <div class="o-rhythm c-comment__content o-rhythm--condensed">
+  <h1>Heading Level 1</h1>
+  <h2>Heading Level 2</h2>
+  <h3>Heading Level 3</h3>
+  <h4>Heading Level 4</h4>
+  <h5>Heading Level 5</h5>
+  <h6>Heading Level 6</h6>
+  <p>
+    This is a paragraph. It contains some <em>italic text</em>, some
+    <strong>bold text</strong>, some
+    <em><strong>bold and italic text</strong></em
+    >, and some inline <code>code</code>, and a
+    <a href="https://example.com">link</a>.
+  </p>
+  <pre
+    class="language-css"
+  ><code class="language-css">.code-block { color: hotpink; }</code></pre>
+  <blockquote>
+    <p>Here’s a blockquote</p>
+    <cite>with a citation</cite>
+  </blockquote>
+  <ul>
+    <li>Alpha</li>
+    <li>
+      Beta
+      <ul>
+        <li>Beta A</li>
+        <li>Beta B</li>
+      </ul>
+    </li>
+    <li>Gamma</li>
+  </ul>
+  <ol>
+    <li>Alpha</li>
+    <li>
+      Beta
+      <ol>
+        <li>Beta A</li>
+        <li>Beta B</li>
+      </ol>
+    </li>
+    <li>Gamma</li>
+  </ol>
+  <table>
+    <thead>
+      <tr>
+        <th>First Header</th>
+        <th>Second Header</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr>
+        <td>Content Cell</td>
+        <td>Content Cell</td>
+      </tr>
+      <tr>
+        <td>Content Cell</td>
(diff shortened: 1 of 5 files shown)


src/vendor/wordpress/demo/compare-comment.twig
```diff
@@ -0,0 +1,76 @@
...
+  <p>
+    <img
+      src="https://api.lorem.space/image/movie?w=150&amp;h=220&amp;t=1"
+      alt="Movie Poster"
+    />
+  </p>
+  <p>And to wrap it all up, here’s a horizontal rule:</p>
+  <hr />
+</div>
+</div>
```
gerardo-rodriguez, line 76: Nit: Can you fix the indentation of this file? ➡️

gerardo-rodriguez (approved):
Left [one nit comment](https://github.com/cloudfour/cloudfour.com-patterns/pull/2022/files#r954963488), LGTM otherwise! 🎉

src/vendor/wordpress/styles/_core-blocks.scss
```diff
@@ -115,7 +115,7 @@ $wp-button-gap: size.$spacing-gap-button-group-default;
  */
 .wp-block-code,
 .wp-block-jetpack-markdown pre,
-.legacy-post pre {
+.legacy-post :not(.c-comment) > :not(.c-comment__content) > pre {
```
Paul-Hebert, line 118: I'm trying to make sure I understand this:

1. Target `pre` elements
2. That are the direct child of an element that's _not_ `.c-comment__content`
3. As long as that parent element is _not_ the direct child of `.c-comment`

A couple of questions:

1. Is that third rule necessary? Could this line be switched to `.legacy-post :not(.c-comment__content) > pre`?
2. Will our legacy posts always have `pre` elements nested deeply enough for this to work? (a.k.a will there always be two layers of elements in between `.legacy-post` and the `pre` we want to style?

spaceninja: 1. Good point — I can drop the `:not(.c-comment) >` part of the selector.

2. Yeah, should be — `legacy-post` is applied to the outermost `o-container__content`, and `pre` elements will either be part of a blog post or a comment, and in both cases nested deeper in the DOM from that point.

Paul-Hebert: Cool, I just had one other thought about a potential edge case. Would this break if the `pre` was nested deeper within a comment. For example, imagine the following comment:

```md
> You wrote this code:
>
> ```js
> console.log('boop');
> ```
>
> I think you meant `beep`
```

In this scenario the `pre` would not be a direct child of `.c-comment__content` so wouldn't get the correct styles.

This is a total edge case but it makes me wonder if it would be safer to just un-set the relevant styles when a `pre` is in a comment?

```css
.c-comment pre {
  margin-inline: 0;
  padding-inline: 0;
}
```

spaceninja: ☹️ Yeah, good call. I'll update.

Paul-Hebert (approved):
LGTM!

Paul-Hebert (approved):

Paul-Hebert (approved):

PR #2039, opened 2022-09-09
Add CSS for Outlined style for WP Image block

spaceninja:
## Overview

This PR adds the styles for the new "Outlined" block style we added to the Image block in the WordPress theme. Ideally, we would just add the `u-border-small` class directly, but WordPress adds the class to the outermost element — in this case, the `figure` — and we only want the border to apply to the nested `img` element.

## Screenshots

Screen Shot 2022-09-09 at 1 37 35 PM

_It's subtle, but there's a 1px border around the image._

## Testing

1. On the preview deploy, go to [Vendor > WordPress > Core Blocks > Image](https://deploy-preview-2039--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-core-blocks--image).
2. In the controls for the story, you can set the style to "Outlined" which will match the markup rendered by WordPress.
3. Verify that an outline is added when you set the Outlined style, and not otherwise.
4. Test in combination with assorted Alignment options as well — the outline should always be applied to the image directly.

---

- Fixes #862

### DIFF
--- a/src/vendor/wordpress/core-blocks.stories.mdx
+++ b/src/vendor/wordpress/core-blocks.stories.mdx
@@ -2,14 +2,6 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
 import blockImageDemo from './demo/image.twig';
 import blockCodeDemo from './demo/code.twig';
 import blockGroupDemo from './demo/group.twig';
-const alignmentClasses = {
-  None: '',
-  Left: 'alignleft',
-  Center: 'aligncenter',
-  Right: 'alignright',
-  Full: 'alignfull',
-  Wide: 'alignwide',
-};
 
 <Meta title="Vendor/WordPress/Core Blocks" />
 
@@ -397,12 +389,28 @@ closing wrapper tag.
     name="Image"
     argTypes={{
       alignment: {
-        options: alignmentClasses,
+        options: {
+          None: '',
+          Left: 'alignleft',
+          Center: 'aligncenter',
+          Right: 'alignright',
+          Full: 'alignfull',
+          Wide: 'alignwide',
+        },
+        defaultValue: '',
+        control: { type: 'select' },
+      },
+      style: {
+        options: {
+          None: 'is-style-default',
+          Outlined: 'is-style-outlined',
+        },
+        defaultValue: 'is-style-default',
         control: { type: 'select' },
       },
     }}
   >
-    {(args) => blockImageDemo({ class: args.alignment })}
+    {(args) => blockImageDemo({ alignment: args.alignment, style: args.style })}
   </Story>
 </Canvas>
 
--- a/src/vendor/wordpress/demo/image.twig
+++ b/src/vendor/wordpress/demo/image.twig
@@ -3,12 +3,13 @@
   style="max-width: 40rem; margin: 0 auto; padding: 0 1.25rem"
 >
   <p>Integer posuere erat a ante venenatis dapibus posuere velit aliquet. Nullam quis risus eget urna mollis ornare vel eu leo. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec id elit non mi porta gravida at eget metus. Vestibulum id ligula porta felis euismod semper.</p>
-  {%
-    if class == 'alignleft' or class == 'aligncenter' or class == 'alignright'
-  %}
-    <div class="wp-block-image">
+  {% if alignment in ['alignleft', 'aligncenter', 'alignright'] %}
+    <div class="wp-block-image {{style}}">
   {% endif %}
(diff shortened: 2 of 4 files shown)


src/vendor/wordpress/core-blocks.stories.mdx
```diff
@@ -2,14 +2,6 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
...
 import blockCodeDemo from './demo/code.twig';
 import blockGroupDemo from './demo/group.twig';
-const alignmentClasses = {
-  None: '',
-  Left: 'alignleft',
-  Center: 'aligncenter',
-  Right: 'alignright',
-  Full: 'alignfull',
-  Wide: 'alignwide',
-};
```
spaceninja, line 12: These are only used in the Image story, so I moved them below to set directly.

src/vendor/wordpress/core-blocks.stories.mdx
```diff
@@ -397,12 +389,28 @@ closing wrapper tag.
...
+          None: 'is-style-default',
+          Outlined: 'is-style-outlined',
+        },
+        defaultValue: 'is-style-default',
         control: { type: 'select' },
       },
     }}
   >
-    {(args) => blockImageDemo({ class: args.alignment })}
+    {(args) => blockImageDemo({ alignment: args.alignment, style: args.style })}
```
spaceninja, line 413: I changed `class` to `alignment` because we're setting multiple classes now, and it made the logic in the twig file easier to understand.

src/vendor/wordpress/demo/image.twig
```diff
@@ -3,12 +3,13 @@
   style="max-width: 40rem; margin: 0 auto; padding: 0 1.25rem"
 >
   <p>Integer posuere erat a ante venenatis dapibus posuere velit aliquet. Nullam quis risus eget urna mollis ornare vel eu leo. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec id elit non mi porta gravida at eget metus. Vestibulum id ligula porta felis euismod semper.</p>
-  {%
-    if class == 'alignleft' or class == 'aligncenter' or class == 'alignright'
-  %}
-    <div class="wp-block-image">
+  {% if alignment in ['alignleft', 'aligncenter', 'alignright'] %}
+    <div class="wp-block-image {{style}}">
```
spaceninja, line 7: This is explained in the CSS, but this awkward thing where we're adding a wrapper element for certain alignments is mirroring the behavior of WordPress. I simplified the `if` check and added the block style, since it's always set on the element that gets `wp-block-image`.

src/vendor/wordpress/styles/_core-blocks.scss
```diff
@@ -157,6 +157,17 @@ $wp-button-gap: size.$spacing-gap-button-group-default;
     margin-block-start: 0; // 1
     max-inline-size: calc(50% - 0.5em); // 2
   }
+
+  /// Apply border only to image
+  ///
+  /// 1. We're using 1px borders here rather than size.$edge-small because it
+  ///    was causing subpixel rendering errors leaving a slight gap between the
+  ///    image and the border.
+  &.is-style-outlined img {
```
gerardo-rodriguez, line 166: @spaceninja Can you help me understand why the code uses a "state" `is-` prefix here? I was expecting a modifier class (e.g., `.wp-block-image--outlined`). 

I might be missing something obvious. I haven't had a chance to create new blocks or CSS for blocks yet. 🙂

src/vendor/wordpress/demo/image.twig
gerardo-rodriguez: Nice! I like the `if blah in array` pattern. 👍🏽

src/vendor/wordpress/demo/image.twig
```diff
@@ -21,9 +22,7 @@
       />
       <figcaption>Mont Blanc appears—still, snowy, and serene.</figcaption>
     </figure>
-  {%
-    if class == 'alignleft' or class == 'aligncenter' or class == 'alignright'
-  %}
+  {% if alignment in ['alignleft', 'aligncenter', 'alignright'] %}
```
gerardo-rodriguez, line 25: Does it make sense to DRY the `['alignleft', 'aligncenter', 'alignright']` array?

src/vendor/wordpress/styles/_core-blocks.scss
spaceninja: Yep — we don't have control over it. We're using WordPress's native "block styles" which let you name your style (in this case "outlined" and it automatically adds a class, in this case `is-style-outlined`.

Since we're dealing with extending a built-in block, our options are somewhat limited.

src/vendor/wordpress/demo/image.twig
spaceninja: I think it's fine — it's also worth remembering that this is just the storybook example, so it's not going to affect the real site if someone edited this and broke it.

gerardo-rodriguez (approved):
:shipit:

PR #2040, opened 2022-09-09
Enhance Calendar Date screen reader UX

### LINKED ISSUE #1724: Enhance Calendar Date screen reader experience
# Summary 

The Calendar Date is used by the Event Summary component. When reviewing https://github.com/cloudfour/cloudfour.com-wp/pull/644, @Paul-Hebert said:
> I also found it a little weird having the talk duration inside of the larger date <time> element. I wonder if it would be better if the date and duration were separate time elements? I'm not sure. Curious what other folks think.

I also agree we can enhance the experience a bit more.

# Ideas

I'm not convinced we need two `<time>` elements, but it's worth exploring to understand what the screen reader experience is like.

I'd also be curious to explore using `aria-hidden` on the month abbreviation and day number, but keeping it off of the note. In theory, a screen reader like VoiceOver might read something like "3-day event starting on April 12, 2022".

We might also consider looking at the Author component which also uses the `<time>` element but [makes some modifications to enhance the UX](https://github.com/cloudfour/cloudfour.com-patterns/blob/5af1d70ca06a8b6c8cf37e5f03c1cafc14d70513/src/components/author/author.twig#L19).

gerardo-rodriguez:
## Overview

This PR modifies the Calendar Date component in an attempt to create a better screen reader user experience.

Changes include:
- added a "single-day event" visually hidden note fallback
- refactor component markup to lessen the VoiceOver duplicate date announcment
- enhance NVDA screen reader UX

## Screenshots

No visual changes.

### Before

Screenshot | NVDA + Firefox | VoiceOver + Safari
--- | --- | ---
Screen Shot 2022-09-09 at 2 10 49 PM | "SEP" <br> "9" | "SEP, September 9, 2022" <br> "9, September 9, 2022"
Screen Shot 2022-09-09 at 2 10 56 PM | "SEP" <br> "9" <br> "3-day event" | "SEP, September 9, 2022" <br> "9, September 9, 2022" <br> "3-day event, September 9, 2022"

### After

Screenshot | NVDA + Firefox | VoiceOver + Safari
--- | --- | ---
Screen Shot 2022-09-09 at 2 10 49 PM | "September 9, 2022" <br> "Single-day event" | "September 9, 2022, September 9, 2022" <br> "Single-day event"
Screen Shot 2022-09-09 at 2 10 56 PM | "September 9, 2022" <br> "3-day event" | "September 9, 2022, September 9, 2022" <br> "3-day event"

### Why is VoiceOver duplicating the date?
Because VoiceOver is silly. It's the only screen reader that exposes the `datetime` attribute. It probably shouldn't. The NVDA UX is much improved, while the VoiceOver UX is also better, even though it's slightly annoying that the `datetime` value is read, causing the date to be announced twice. Nothing we can do about that, and in my opinion, this solution finds a nice middle-ground where both NVDA and VoiceOver UX are better and acceptable, even though the VoiceOver UX is not perfect.

### Previews

- [Calendar Date component](https://deploy-preview-2040--cloudfour-patterns.netlify.app/?path=/docs/components-calendar-date--basic)
- [Media Summary component with Calendar Date component](https://deploy-preview-2040--cloudfour-patterns.netlify.app/?path=/story/components-media-summary--event)

1. Use VoiceOver and NVDA (if you have access to it) to test the screen reader UX

---

- Closes #1724

### DIFF
--- a/.changeset/dry-kangaroos-try.md
+++ b/.changeset/dry-kangaroos-try.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Enhance Calendar Date screen reader UX
--- a/src/components/calendar-date/calendar-date.twig
+++ b/src/components/calendar-date/calendar-date.twig
@@ -1,21 +1,40 @@
-{% set datetime = datetime|default('now') %}
-{% set days = days|default(1) %}
+{% set _datetime = datetime|default('now') %}
 {% set _note %}{% block note %}{{ note }}{% endblock %}{% endset %}
 
-<time class="c-calendar-date c-calendar-date--{{ datetime|date('F')|lower }}{% if class %} {{ class }}{% endif %}"
-  datetime="{{ datetime|date('Y-m-d') }}"
+<div
+  class="c-calendar-date c-calendar-date--{{ _datetime|date('F')|lower }}{% if class %} {{ class }}{% endif %}"
 >
-  <span class="c-calendar-date__header">
-    {{ datetime|date('M') }}
+  {# The abbreviated month is aria-hidden as it's only presentational #}
+  <span class="c-calendar-date__header" aria-hidden="true">
+    {{ _datetime|date('M') }}
   </span>
+
   <span class="c-calendar-date__main">
-    <span class="c-calendar-date__day">
-      {{ datetime|date('j') }}
-    </span>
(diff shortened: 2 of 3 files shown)


spaceninja (approved):
Code looks good, and changes work as described. Nice job! :shipit:

PR #2049, opened 2022-09-15
Allow WordPress Group blocks to inherit rhythm

### LINKED ISSUE #2033: Allow Group block to use o-rhythm class
I grouped multiple paragraphs, a quote, and a heading into a Group block so I could add a background color. I can add the `o-rhythym` class via the "Additional CSS classes" input, but it doesn't actually work because the Group block adds an inner `.wp-block-group__inner-container` container.

Screen Shot 2022-09-02 at 2 04 21 PM

Screen Shot 2022-09-02 at 2 06 53 PM

tylersticka:
## Overview

Addresses an issue with `.wp-block-group` elements where they wouldn't inherit vertical rhythm from the `.o-rhythm` object. This could not be resolved via WordPress admin settings because the `class` attribute will not impact the `.wp-block-group__inner-container` element.

A common use case for this block is to group floating images and adjacent content. To prevent extra space for showing up, a little bit of a selector hack was used to negate the top margin on items following a left- or right-aligned image block.

Alternatively I could have replaced this fix with `.wp-block-group.o-rhythm > * { @include spacing.vertical-rhythm; }`, but then you'd need to add the class every time, and it wouldn't include a fix for the image alignment feature above. 

### Before

Note the lack of space between the blockquote and paragraph…

Screen Shot 2022-09-15 at 2 28 35 PM

### After

Screen Shot 2022-09-15 at 2 28 01 PM

## Testing

[Try out this story on the deploy preview](https://deploy-preview-2049--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-core-blocks--group&args=background:has-gray-lighter-background-color+has-background)

---

- Fixes #2033

### DIFF
--- a/.changeset/empty-forks-type.md
+++ b/.changeset/empty-forks-type.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Allow WordPress Group blocks to inherit vertical rhythm
--- a/src/vendor/wordpress/demo/group.twig
+++ b/src/vendor/wordpress/demo/group.twig
@@ -26,6 +26,7 @@
         </p>
         <cite>Scott Henning, Marketing Director</cite>
       </blockquote>
+      <p>One more element in this group.</p>
     </div>
   </div>
   <p>
--- a/src/vendor/wordpress/styles/_core-blocks.scss
+++ b/src/vendor/wordpress/styles/_core-blocks.scss
@@ -205,6 +205,21 @@ figure.wp-block-image {
   }
 }
 
+/// Gutenberg block: Group
+///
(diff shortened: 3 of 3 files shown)


gerardo-rodriguez (approved):
OMG, I ran into this in my [most recent blog post](https://cloudfour.com/thinks/removing-layout-shift-from-a-progressively-enhanced-burger-menu/#why-can%E2%80%99t-i-just-add-a-%3Ccode%3Erole%3D%22button%22%3C%2Fcode%3E-to-the-%3Ccode%3E%26lt%3Ba%26gt%3B%3C%2Fcode%3E-element%3F). Thank you for fixing this!

Yes! Thank you!

PR #2051, opened 2022-09-15
Fix nested rhythm object issue

### LINKED ISSUE #1933: Better Handling for Nested Rhythm Objects
Currently, if you have nested Rhythm objects, it can have unexpected behaviors. Example:

```html
<div class="o-rhythm o-rhythm--generous">
  <h1>blog title</h1>
  <div class="o-rhythm o-rhythm--default o-rhythm--generous-headings">
     <p>blog content</p>
     <p>blog content</p>
  </div>
</div>
```

I think a reasonable reading of the intent here is that there should be generous spacing between the blog title and the div wrapping the content, and then default spacing between the content paragraphs.

What actually happens is default spacing between the blog title and the div wrapping the content.

That's because the spacing on the div is set to `--rhythm` and that custom property is reset by `o-rhythm--default` to default spacing. The _intent_ is that custom property resets spacing _inside_ the element. The _effect_ is that the custom property resets spacing on the element itself as well.

In the WordPress theme, you can see an example of us working around this shortcoming in `talk.twig`, where several nested Rhythm objects are wrapped in a `div` just to keep them from affecting the overall page spacing.

tylersticka:
## Overview

By default, a nested rhythm object inherits the `--rhythm` of its parent. In cases where you did _not_ want to inherit the rhythm, however, the containing element would lose its parent `--rhythm`, which necessitated multiple containing elements to overcome.

This PR changes the modifiers so the `--rhythm` is applied to immediate children, which prevents the element itself from losing its own rhythm.

### Before

Notice how the list with `compact` spacing runs up against its sibling, even though the containing element has `generous` spacing:

Screen Shot 2022-09-15 at 4 11 15 PM

### After

Now the `compact` spacing only applies to the element's children, so the parent spacing is intact:

Screen Shot 2022-09-15 at 4 12 11 PM

## Testing

Open [this story in the deploy preview](https://deploy-preview-2051--cloudfour-patterns.netlify.app/?path=/story/objects-rhythm--nesting&args=amount:generous;nested_amount:compact) and mess with the `amount` and `nested_amount` controls. The `amount` spacing should always apply to the `ul` itself, while the `nested_amount` should only apply to the list elements therein.

---

- Fixes #1933

### DIFF
--- a/.changeset/hip-numbers-learn.md
+++ b/.changeset/hip-numbers-learn.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Nested rhythm objects with modifiers no longer lose their own vertical rhythm
--- a/src/objects/rhythm/demo/nesting.twig
+++ b/src/objects/rhythm/demo/nesting.twig
@@ -1,9 +1,26 @@
 {% embed '@cloudfour/objects/rhythm/rhythm.twig' %}
   {% block content %}
-    <p>This paragraph is in a rhythm object with {% if amount %}<code>{{amount}}</code>{% else %}no{% endif %} spacing set.</p>
-    <p>This list is also a rhythm object, inheriting that spacing:</p>
+    <p>
+      This paragraph is in a rhythm object with
+      {% if amount %}
+        <code>{{amount}}</code>
+      {% else %}
+        no
+      {% endif %}
+      spacing set.
+    </p>
+    <p>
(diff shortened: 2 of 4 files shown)


spaceninja (approved):
🎉  Thank you! Great change, code looks good, and works as described.

gerardo-rodriguez (approved):
🎉

PR #2053, opened 2022-09-16
Add aspect ratio CSS utility classes, remove Embed object

### LINKED ISSUE #1999: Simplify aspect ratio boxes where possible
Our [Embed object](https://cloudfour-patterns.netlify.app/?path=/docs/objects-embed--image) and component elements like [Card covers](https://cloudfour-patterns.netlify.app/?path=/docs/components-card--cover-image) rely on [a padding hack](https://css-tricks.com/aspect-ratio-boxes/) as of this writing. Could the [recent widespread support of the CSS `aspect-ratio` property](https://caniuse.com/mdn-css_properties_aspect-ratio) allow us to simplify?

gerardo-rodriguez:
## Overview

This PR does the following:
- Introduces `u-aspect-*` CSS utility classes
- Removes the Embed object
- Removes the Rounded example per [PR feedback](https://github.com/cloudfour/cloudfour.com-patterns/pull/2053/commits/0ce231bd5ef3fd5b5fde50a24574c2a9343d3f0c)

Reference I used: https://web.dev/aspect-ratio/

## Screenshots

Left: Before
Right: After

Screen Shot 2022-09-22 at 1 12 45 PM

## Testing

Before: https://cloudfour-patterns.netlify.app/?path=/docs/objects-embed--image
After: https://deploy-preview-2053--cloudfour-patterns.netlify.app/?path=/docs/utilities-aspect-ratio--image

**Check all browsers!** Especially Safari. 😅 

- [ ] Confirm all four aspect ratio utility classes work (`square`, `full`, `wide`, `cinema`)
- [ ] Ensure the `--aspect-ratio` CSS custom property works
- [ ] Compare before/after, there should be no visual differences

---

- Closes #1999


tylersticka:
I haven't tested thoroughly, but this PR seems to do a good job of swapping the properties.

I think the bigger question I have is whether or not this property invalidates the need for this component entirely? Like, is there value in having the containing element at all?

Here's an example of where I replace the component with a utility on the embedded element directly: https://codepen.io/tylersticka/pen/zYjZdvB/491a9324047e4651d25bcf6a7bdefbb4

gerardo-rodriguez:
> I think the bigger question I have is whether or not this property invalidates the need for this component entirely? Like, is there value in having the containing element at all?

> Here's an example of where I replace the component with a utility on the embedded element directly: https://codepen.io/tylersticka/pen/zYjZdvB/491a9324047e4651d25bcf6a7bdefbb4

@tylersticka Interesting! I can look into making this a CSS utility class instead. Thanks!

tylersticka:
@gerardo-rodriguez I think that's kind of what I was asking when I asked this:

> Like, is there value in having the containing element at all?

For example, I think the Avatar component does not use the Embed object directly, so in that case it feels like the CSS could be self-contained. I wasn't sure if there are components like Card that rely on it and, if so, what that looks like.

We could also do a phased approach where we "deprecate" the Embed component and move over to the aspect ratio utilities slowly.

Or you could chat with others in @cloudfour/dev and determine that there's no harm in switching Embed over and I'm over-complicating things. Mainly I just want to make sure we've considered the usefulness of the component in a world without the padding hack.

tylersticka:
> Or you could chat with others in @cloudfour/dev and determine that there's no harm in switching Embed over and I'm over-complicating things.

Just to elaborate on this point: If you all decide that maybe in the future it might be useful to add _other things_ to an embed, like [an icon or overlay](https://cpe-patterns.netlify.app/patterns/layout/flex-embed.html#transparent-overlay-with-icon) or [some other visual element](https://baptist-health-patterns.netlify.app/bh/patterns/components/flex-embed/#with-shadow), then maybe it's fine to keep a component where the class is on the containing element.

I just want to be clear that I'm asking for confirmation that we've considered the component's continued usefulness and a recommendation for what to do next: My intent is not to provide specific direction right now.

Happy to chat with voices if that helps!

gerardo-rodriguez:
@tylersticka @cloudfour/dev This PR has been reopened and is ready for review. Thank you! 😄

spaceninja (dismissed):
Code looks good and works as described. I'll defer final approval to @tylersticka

src/utilities/aspect-ratio/aspect-ratio.scss
```diff
@@ -0,0 +1,21 @@
...
+$aspect-ratio-map: meta.module-variables('aspect-ratio');
+
+.u-aspect-ratio {
+  @each $name, $ratio in $aspect-ratio-map {
+    &-#{$name} {
+      @include media-query.breakpoint-classes {
+        --aspect-ratio: #{$ratio};
+        aspect-ratio: var(--aspect-ratio);
+        block-size: 100%;
+        inline-size: 100%;
```
tylersticka, line 16: Are these properties always necessary?

If so, then this maybe needs to be a component?

src/utilities/aspect-ratio/aspect-ratio.scss
```diff
@@ -0,0 +1,21 @@
...
+        object-fit: cover;
```
tylersticka, line 17: Wasn't there a bug with Safari video embeds and this property?

src/utilities/aspect-ratio/aspect-ratio.scss
```diff
@@ -0,0 +1,21 @@
+@use 'sass:meta';
+@use '../../compiled/tokens/scss/aspect-ratio';
+@use '../../mixins/media-query';
+
+/// Saving a map of aspect ratio values makes it easier to loop through.
+/// @type Map
+$aspect-ratio-map: meta.module-variables('aspect-ratio');
+
+.u-aspect-ratio {
```
tylersticka, line 9: Could we simplify these class names to `u-aspect-*`?

src/utilities/aspect-ratio/aspect-ratio.stories.mdx
```diff
@@ -92,9 +79,11 @@ Embeds may be combined with [rounded corner utilities](/docs/utilities-rounded-c
   <Story
     name="Rounded"
```
tylersticka, line 80: I'm not sure this example makes sense anymore since it's not really relevant to this type of utility. We can probably remove it.

src/utilities/aspect-ratio/aspect-ratio.scss
gerardo-rodriguez: > Are these properties always necessary?

@tylersticka They seem to both be necessary when using them for a video embed.

> If so, then this maybe needs to be a component? (Edit: Or object?)

Can you elaborate on your thinking process? It's unclear to me why it should be a component or object over a utility class. This will help me in the future when I need to make this decision again. 🙂

src/utilities/aspect-ratio/aspect-ratio.scss
gerardo-rodriguez: Oh, good catch. I didn't _play_ the video. 🤦🏽

gerardo-rodriguez: Hmm...the `video` element works fine. It's the YouTube `<iframe>` embed that has the issue. 

If I add `object-position: left`, it seems to fix the YouTube `<iframe>` embedded video but then all of the rest of the examples (e.g. image or picture) are also get repositioned `left` which is not what we want.

gerardo-rodriguez: > If I add `object-position: left`, it seems to fix the YouTube `<iframe>` embedded video but then all of the rest of the examples (e.g. image or picture) also get repositioned `left` which is not what we want.

Or, if I _remove_ `object-fit: cover`, then the Safari issue goes away as well. But then the image assets get squished to fit within their aspect ratio space. 🤔

gerardo-rodriguez: I suppose, if [I did switch to an object or component](https://github.com/cloudfour/cloudfour.com-patterns/pull/2053#discussion_r978114186), I could have a "YouTube" modifier that unsets `object-fit: cover` maybe?

src/utilities/aspect-ratio/aspect-ratio.scss
tylersticka: So according to [Harry Roberts](https://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/):

> Utilities are complete [single responsibility](https://csswizardry.com/2012/04/the-single-responsibility-principle-applied-to-css/) rules which have a very specific and targeted task.

[SUIT](https://github.com/suitcss/suit/tree/master/packages/utils#templating) adds a bit more detail to that:

> Each utility class modifies a single trait (a small collection of similar styles).

So if the purpose of these utilities are to apply a specific aspect ratio, I guess the question becomes: Will these additional sizing styles be expected or will they be burdensome?

One possibility would be to use `:is` or `:where` to apply the opinionated convenience styles, so they will be easy to override in case someone wants an element to be `50%` width with a specific aspect ratio?

I guess as I'm writing this, I think this _can_ work as a utility. I just want to make sure that adding these `size` properties won't make the elements harder to resize in contexts where we only want them to fill _some_ of the space?

src/utilities/aspect-ratio/aspect-ratio.scss
tylersticka: Or you could move that to a separate utility.

Or you could qualify that property to only `img` elements?

gerardo-rodriguez: I found a fix for the Safari bug. Thanks for the idea, @tylersticka! https://github.com/cloudfour/cloudfour.com-patterns/pull/2053/commits/21cac28d0886d7e371b13f81b2547e324800b698

src/utilities/aspect-ratio/aspect-ratio.scss
gerardo-rodriguez: > One possibility would be to use `:is` or `:where` to apply the opinionated convenience styles, so they will be easy to override in case someone wants an element to be `50%` width with a specific aspect ratio?

Ah, thanks for helping me understand the concern! Let me see if I can get it to work. 🔍 👀

gerardo-rodriguez:
@tylersticka @cloudfour/dev Thanks for your patience! I got myself all tied up in a knot, but I think I got it now!

src/utilities/aspect-ratio/aspect-ratio.stories.mdx
```diff
@@ -2,54 +2,41 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs';
...
+Specifying aspect ratios for imagery and embedded media is important to prevent jarring layout shifts as content loads. But specifying width and height may not work for all types of embedded media. The `u-aspect-*` utility classes use the [`aspect-ratio` CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio) to maintain intended aspect ratio.
 
-Embed objects support two methods for specifying the desired aspect ratio:
+There are two methods for specifying the desired aspect ratio:
 
-1. Via a modifier class. All of our [aspect ratio tokens](/docs/design-tokens-aspect-ratio--page) are supported. Example: `o-embed--wide`
+1. Via the `u-aspect-*` utility class. All of our [aspect ratio tokens](/docs/design-tokens-aspect-ratio--page) are supported. Example: `u-aspect-wide`
 1. Via the `--aspect-ratio` custom property. This can be helpful for one-off images in articles or other content that do not conform to a typical size. Example: `--aspect-ratio: 320/480`
```
tylersticka, line 40: I don't know if this use case is relevant anymore. It would be confusing to specify one aspect ratio via the utility class (`u-aspect-wide`) only to override it with `--aspect-ratio`. And looking at the CSS, the use of the custom property is no longer gaining us any efficiencies because we're setting the same styles on every selector.

src/utilities/aspect-ratio/demo/image.twig
```diff
@@ -0,0 +1,17 @@
+{% if aspectRatio %}
+  {% set style = (style|default('')) ~ ';--aspect-ratio:' ~ aspectRatio %}
+{% endif %}
+
+<img
+  class="{{ aspectRatioClass }}"
+  src="{{ src }}"
+  alt=""
+  width="900"
```
tylersticka, line 9: What happens now if `width` isn't set?

src/utilities/aspect-ratio/demo/youtube-video.twig
```diff
@@ -0,0 +1,16 @@
...
+  {% set style = (style|default('')) ~ ';--aspect-ratio:' ~ aspectRatio %}
+{% endif %}
+
+<iframe
+  class="{{ aspectRatioClass }} u-border-medium"
+  width="560"
+  height="315"
+  src="https://www.youtube.com/embed/Ty5jtMZXbmk"
+  allowfullscreen
+  style="width: 100%;
```
tylersticka, line 11: What happens if this `width` `style` is omitted?

gerardo-rodriguez:
Moving back to "draft" until feedback is addressed.

src/utilities/aspect-ratio/aspect-ratio.stories.mdx
gerardo-rodriguez: @tylersticka Oh, interesting. I hadn't thought about it from that perspective.

What about the "[Responsive](https://deploy-preview-2053--cloudfour-patterns.netlify.app/?path=/docs/utilities-aspect-ratio--responsive)" use case?

Is that a use case still needing the `u-aspect-*` utility classes?

gerardo-rodriguez: Your question makes me wonder if utility classes like `u-aspect-wide` should not be configurable. But maybe a generic `u-aspect-ratio` utility can be configurable via a CSS custom prop?

src/utilities/aspect-ratio/demo/image.twig
gerardo-rodriguez: I adjusted the CSS and removed this `width` attribute. The intent is that it should work without it.

src/utilities/aspect-ratio/demo/youtube-video.twig
gerardo-rodriguez: I adjusted the CSS and removed this `style` `width` attribute. The intent is that it should work without it.

src/utilities/aspect-ratio/aspect-ratio.stories.mdx
gerardo-rodriguez: Or maybe the generic customizable utility could be `u-aspect-custom`?

gerardo-rodriguez: Curious is @cloudfour/dev has any thoughts.

gerardo-rodriguez: To be clear, what I am proposing is:
- `u-aspect-wide`, `u-aspect-full`, `u-aspect-square`, `u-aspect-cinema` + responsive versions
    - Aspect ratio _cannot_ be modified, it is hard-coded)
- `u-aspect-custom` 
    - The aspect ratio can be customized via the `--aspect-ratio` CSS variable

tylersticka: It's been a few months since that comment, so I can't be 100% sure what my concern was then. But looking at it now with fresh eyes, two things stick out to me.

First, these utilities are doing a lot more than we'd typically want utility classes to do. We're setting `aspect-ratio`, we're setting `object-size`, we're doing special things for iframes. All of that is useful, but I don't know that I'd look at a utility class like `u-aspect-*` and think "that's doing more than setting `aspect-ratio`." This may actually be a component class.

Second, it feels weird that we have the utility class name specifying the aspect ratio, which uses an `--aspect-ratio` custom property, which sets the value of the `aspect-ratio` rule. Setting aside all the extra stuff the utilities do, on their surface these all seem equivalent:

```html
<div class="u-aspect-square"></div>

<div class="u-aspect-square" style="--aspect-ratio: 1;"></div>

<div class="u-aspect-square" style="aspect-ratio: 1;"></div>
```

Which begs the question: What's the point of the custom property?

I wonder if what we really want to do is make this a _breaking change_ to `c-embed`, where the class is applied directly to the aspect ratio element instead of to its children?

tylersticka:
This PR is now several months old, and our embed styles are now shared between the embed component and multiple blocks. I'm going to close this PR in favor of re-approaching this with fresh eyes at a future date.

PR #2057, opened 2022-09-20
Use CSS aspect-ratio for Card component

gerardo-rodriguez:
## Overview

This PR proposes a solution for using the CSS `aspect-ratio` property and removing the `ratio-box` dependency.

Other changes include:
- Remove `ratio-box` import in Avatar CSS (missed this in my last PR)
- Fix a bug where the circle cover variant wasn't a circle

## Screenshots

Left: Before
Right: After

Screen Shot 2022-09-20 at 2 30 48 PM
Screen Shot 2022-09-20 at 2 31 12 PM
Screen Shot 2022-09-20 at 2 35 42 PM
Screen Shot 2022-09-20 at 2 36 49 PM

I also fixed a bug where the circle wasn't a circle at ~640px wide viewports:

Screen Shot 2022-09-20 at 2 38 59 PM

## Testing

Before: https://cloudfour-patterns.netlify.app/?path=/docs/components-card--cover-image
After: https://deploy-preview-2057--cloudfour-patterns.netlify.app/?path=/docs/components-card--cover-image

- [ ] Review at various viewport sizes, confirm layout is the same as before
- [ ] Confirm [the circle is a circle](https://deploy-preview-2057--cloudfour-patterns.netlify.app/iframe.html?args=&id=components-card--circular-cover-image&viewMode=story) at a viewport width of ~640px

---

- #1999

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -121,7 +121,7 @@ $_focus-overflow: (size.$edge-large * -1);
      * When the cover image is circular (and thus taller), give it a smaller
      * amount of space and keep the text left-aligned.
      *
-     * 1. Fixes bug where the cirlce wasn't a circle anymore at smallest viewports
+     * 1. Fixes bug where the circle cover wasn't a circle at smallest viewports
      */
 
     &.c-card--circle-cover {
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -235,15 +235,8 @@ $_focus-overflow: (size.$edge-large * -1);
   background-color: var(--theme-color-background-secondary); /* 1 */
   border-radius: size.$border-radius-medium;
   grid-area: cover;
-  transform: translate(0, 0); /* 2 */
   overflow: hidden;
-
-  > *,
-  > picture > img {
-    block-size: 100%;
-    inline-size: 100%;
-    object-fit: cover;
-  }
+  transform: translate(0, 0); /* 2 */
 
   /**
    * Subtly brighten covers within our alternate theme containers to help offset
@@ -295,6 +288,9 @@ $_focus-overflow: (size.$edge-large * -1);
 
 .c-card__cover > *,
 .c-card__cover > picture > img {
+  block-size: 100%;
+  inline-size: 100%;
+  object-fit: cover;
   transition: transform transition.$prompt ease.$out;
 
   .c-card--with-link:hover:not(:active) & {
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -244,11 +244,8 @@ $_focus-overflow: (size.$edge-large * -1);
    */
 
   .c-card--circle-cover & {
+    aspect-ratio: 1;
     border-radius: size.$border-radius-full;
-
-    &::before {
-      padding-block-end: 100%;
-    }
   }
 
   /**
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -11,7 +11,6 @@
 @use '../../mixins/headings';
 @use '../../mixins/media-query';
 @use '../../mixins/ms';
-@use '../../mixins/ratio-box';
 @use '../../mixins/theme';
 
 /// We can't use `grid-gap` exclusively due to some containers only being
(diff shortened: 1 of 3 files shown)


src/components/card/card.scss
```diff
@@ -125,6 +132,11 @@ $_focus-overflow: (size.$edge-large * -1);
         'cover footer footer'
         'cover . .';
       text-align: start;
+
+      .c-card__cover {
+        block-size: auto; /* 1 */
+        inline-size: auto; /* 1 */
```
tylersticka, line 138: If I remove `inline-size: auto`, it seems to still work fine? Any reason we need both?

src/components/card/card.scss
```diff
@@ -112,9 +112,16 @@ $_focus-overflow: (size.$edge-large * -1);
...
+    .c-card__cover {
+      block-size: 100%; /* 4 */
+      inline-size: 100%; /* 4 */
+    }
+
     /**
      * When the cover image is circular (and thus taller), give it a smaller
      * amount of space and keep the text left-aligned.
+     *
+     * 1. Fixes bug where the circle cover wasn't a circle at smallest viewports
```
tylersticka, line 124: Nit: Could we make this comment easier to follow by moving it closer to the block it pertains to?

src/components/card/card.scss
tylersticka: Also, can we move this style down to the `.c-card__cover` section instead of being a child of the modifier class? That way all of the cover styles are in one place.

gerardo-rodriguez: > If I remove `inline-size: auto`, it seems to still work fine? Any reason we need both?

Good catch, both are likely not needed. 👍🏽

src/components/card/card.scss
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/2057/commits/12912dbd0fc35c8dfefab39d2bc83e063b19331c

src/components/card/card.scss
gerardo-rodriguez: > Also, can we move this style down to the `.c-card__cover` section instead of being a child of the modifier class? That way all of the cover styles are in one place.

@tylersticka I'm not sure if I've painted myself into a corner, but the `block-size: auto` needs to be applied as a child of the `.c-card--circle-cover` modifier:

- `.c-card__cover`: Doesn't need explicit `block-size`
- `.c-card--horizontal .c-card__cover`: Needs `block-size` and `inline-size` of `100%`
- `.c-card--horizontal.c-card--circle-cover .c-card__cover`: Needs `block-size: auto` (fixes the bug I noticed)

Have I made it too complex? 🔍 👀

gerardo-rodriguez: Ah, got it. Thanks, @tylersticka, appreciate it!

src/components/card/card.scss
```diff
@@ -112,6 +112,11 @@ $_focus-overflow: (size.$edge-large * -1);
     grid-template-columns: repeat(3, 1fr); /* 2 */
     grid-template-rows: 1fr repeat(3, minmax(0, auto)) 1fr; /* 3 */
 
+    .c-card__cover {
+      block-size: 100%; /* 4 */
+      inline-size: 100%; /* 4 */
+    }
```
tylersticka, line 118: These styles should be closer to the element they are modifying, not nested within a modifier of the parent element (see previous thread).

gerardo-rodriguez: @tylersticka Got it!

So, in this case, it'd be something like the following?
```scss
.c-card__cover {
  /* card cover styles */

  .c-card--horizontal & {
    @include media-query.breakpoint-classes(
      $from: m,
      $to: xl,
      $include-default: false
    ) {
      block-size: 100%; 
      inline-size: 100%; 
    }
  }
}
```

tylersticka: @gerardo-rodriguez I recommend using the `breakpoint-parent-classes` mixin instead (see the comments in `_media-query.scss` for explanation and an example), it was meant for use cases like this.

gerardo-rodriguez: Thanks, @tylersticka. I didn't know about the `breakpoint-parent-classes` mixin, it works great! 😁 🎉 

https://github.com/cloudfour/cloudfour.com-patterns/pull/2057/commits/9f1705c64b1e95c732f8c46c019c3da7a881c957

spaceninja (approved):
Code looks good and behaves as described. I'll defer final approval to @tylersticka, though.

Issue #1403, opened 2021-07-16
Consider switching from Webpack to Vite


tylersticka:
Storybook 6.3 introduces support for bundlers other than Webpack: https://storybook.js.org/blog/storybook-for-vite/

@calebeby pointed out while reviewing a PR by @gerardo-rodriguez that using Vite might simplify the process we've been using for exposing template source code to stories: https://github.com/cloudfour/cloudfour.com-patterns/pull/1402/files#r670882613

> Storybook has pluggable bundlers now, so it's not very tied to webpack anymore. But we could make a vite plugin for this syntax if we switched to vite

Unfortunately we haven't upgraded to 6.3 yet because of unresolved regressions related, perhaps ironically, to Webpack. See: #1349

While I don't _love_ the idea of refactoring the project's configuration (which is currently pretty Webpack-centric), if the 6.3 upgrade will require _some_ reconfiguration _anyway_, we might as well consider switching to a bundler we like better and that might gain us some efficiencies.

Paul-Hebert:
I've heard good things about Vite! I also don't love the idea of refactoring the config, but I do love the idea of getting experience with Vite!

calebeby:
I think it would make a lot of sense to switch, here's some reasons:

- Fast startup time (<500ms)
- Faster HMR/updates on save (<200ms)
- Smaller production bundle (because rollup)
- Easier and more readable configuration/better customizability
- Smaller dependency graph (no more zillions of dependencies from the webpack ecosystem)

I'm taking a look to try to size this issue

calebeby:
Here are things I noticed while testing out the swap:

- Vite isn't able to statically determine the module graph so the pre-bundling dependencies step runs 3-4 times as it builds some stuff, tries to load it, runs into some dependencies, runs the pre-bundling again, and reloads. This is annoying but not too much of a problem since it should cache them.
- Something is preventing the pre-bundling result from being cached. This is a problem.
- Any time we use `import { ... } from 'path'` in the browser, that does not work because `path` doesn't exist in the browser.
- Any time we do `require()` or `require.context` in the browser, that does not work because `require` does not exist in the browser and `require.context` is a webpack-ism. We can get something relatively similar using glob import (a vite-ism): https://vitejs.dev/guide/features.html#glob-import
- The webpack loader syntax for files does not work (`!!raw-loader!`, etc) does not work either, but we can make it work easily by adding a rollup plugin.
[...]

PR #2080, opened 2022-11-08
Set specific Node version

tylersticka:
## Overview

A few days ago, Node v18 became the latest LTS release. But our Storybook build commands fail in Node v17 or higher.

The reason for this is that Node changed its cryptographic security requirements in version 17. This breaks the way Webpack generates hashes.

After consulting [this StackOverflow thread](https://stackoverflow.com/q/69692842), I had a few options to address this:

1. Run `npm audit fix --force`. This introduces _many_ breaking changes.
2. Set the `NODE_OPTIONS` environment variable to `--openssl-legacy-provider` everywhere. This felt clunky to me.
3. Set [some Webpack options](https://stackoverflow.com/a/73465262). I attempted this, but it turns out those options are only available in Webpack 5, and this project is currently using version 4.
4. Specify the previous LTS release in our `.nvmrc`.

I went with the last option.

We can update to the latest LTS release once Webpack is updated to version 5 or later, which I believe depends on Storybook. Or we can switch to Vite (see #1403).

## Testing

1. Locally, make sure your nvm aliases are up to date by running `nvm ls-remote`
2. `git checkout main`
3. Recreate the error: `nvm use && npm ci && npm start`
4. Now check out this branch.
5. Try the previous command again: `nvm use && npm ci && npm start`
6. Confirm there is no longer a build error.

Alternatively, you can compare the GitHub checks on this PR to that of #2079, which was stalled by this issue.

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/*
+lts/gallium


spaceninja (approved):
Perfectly reasonable solution, thanks Tyler!

PR #2081, opened 2022-11-22
Moar social media exodus icons

tylersticka:
## Overview

As the big social media exodus continues, I want some freedom in what social media we choose to promote. I also wanted to make fill a few gaps in our A List Apart-related content.

## Screenshots

Screen Shot 2022-11-21 at 4 42 19 PM


Paul-Hebert (approved):
Looks good!

The Feed icon doesn't display properly as a raw SVG:

image

I think this is fine, since we plan to use it with an icon component that sets stroke colors, but I wanted to flag this in case it was unexpected.

tylersticka:
> I think this is fine, since we plan to use it with an icon component that sets stroke colors, but I wanted to flag this in case it was unexpected.

Nope, that's expected. The same is true of a lot of the arrow icons and stuff like that. Since our CSS sets the stroke color, it's kind of redundant to include `stroke="currentColor"` in our SVGs.

At least, that was the logic at the time. I thought it better to be consistent for now and re-think later if we run into hurdles with that.

Paul-Hebert:
Cool, that makes sense :+1:

PR #2089, opened 2022-11-30
Allow captions for centered WordPress Image Blocks to fill their container

### LINKED ISSUE #2019: Image block captions should be able to extend past the width of the image
Right now, the layout is a bit funky as the width of the caption is restrained to the width of the image.

Screen Shot 2022-08-23 at 4 41 01 PM

tylersticka:
## Overview

The default styles for WordPress image blocks with captions that are right-, left- or center-aligned use `table` display styles to constrain captions to the width of the image.

That makes less sense for our content, where we often need to include narrow-width screenshots with accompanying explanations that can feel a bit _confined_ in their space. It also makes them less consistent with regular ol' `figure` styles in our larger project.

## Screenshots

Before | After
--- | ---
Screenshot 2022-11-30 at 11 41 36 AM | Screenshot 2022-11-30 at 11 41 22 AM

## Testing

1. View [the Vendor ▸ WordPress ▸ Utilities ▸ Alignment story](https://deploy-preview-2089--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-utilities--alignment&args=alignment:aligncenter) with `aligncenter` selected.
2. Inspect the image and set its width to be smaller than the caption _or_ inspect the caption and make it longer.
3. Observe that the caption is able to extend past the image boundaries.

---

- Fixes #2019 

/CC @gerardo-rodriguez

### DIFF
--- a/src/vendor/wordpress/styles/_core-blocks.scss
+++ b/src/vendor/wordpress/styles/_core-blocks.scss
@@ -158,6 +158,17 @@ $wp-button-gap: size.$spacing-gap-button-group-default;
     max-inline-size: calc(50% - 0.5em); // 2
   }
 
+  /// By default, WordPress uses table styles to constrain captions to the size
+  /// of the image. This makes sense for left- and right-aligned image blocks,
+  /// but for consistency with our non-block `figure` elements, we allow
+  /// captions to go as wide as they need for centered elements.
+  .aligncenter {
+    &,
+    > figcaption {
+      display: block;
+    }
+  }
+
   /// Apply border only to image
   ///
   /// 1. We're using 1px borders here rather than size.$edge-small because it
--- a/.changeset/honest-spoons-reflect.md
+++ b/.changeset/honest-spoons-reflect.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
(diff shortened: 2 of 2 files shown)


src/vendor/wordpress/styles/_core-blocks.scss
```diff
@@ -158,6 +158,17 @@ $wp-button-gap: size.$spacing-gap-button-group-default;
...
+  /// By default, WordPress uses table styles to constrain captions to the size
+  /// of the image. This makes sense for left- and right-aligned image blocks,
+  /// but for consistency with our non-block `figure` elements, we allow
+  /// captions to go as wide as they need for centered elements.
+  .aligncenter {
+    &,
+    > figcaption {
+      display: block;
+    }
+  }
```
tylersticka, line 170: I considered using `display: initial`, but that actually doesn't get us what we want since images aren't block display by default.

I also considered using `display: block` for `.wp-block-image .aligncenter` but `display: inherit` for the `figcaption`. But I couldn't think of a scenario where we'd actually _want_ that value to inherit... if I set a class on `.aligncenter` that changed its display to `flex`, I wouldn't want that to also impact the `figcaption`.

So I landed on just setting both to `display: block` explicitly.

gerardo-rodriguez (approved):
👍🏽

PR #2098, opened 2022-12-14
Disable resizing for Elastic Textareas

gerardo-rodriguez:
## Overview

This PR progressively enhances the Elastic Textarea experience by disabling [`resize`](https://developer.mozilla.org/en-US/docs/Web/CSS/resize).

## Screenshots

Screen Shot 2022-12-15 at 2 15 41 PM

## Testing

Preview: https://deploy-preview-2098--cloudfour-patterns.netlify.app/?path=/docs/components-input--elastic-textarea

Confirm:
- [ ] When JS is disabled, `resize` is not disabled
- [ ] When JS is enabled, `resize` is disabled

---

- https://github.com/cloudfour/cloudfour.com-wp/issues/945

### DIFF
--- a/.changeset/strange-steaks-sneeze.md
+++ b/.changeset/strange-steaks-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Add option to disable elastic textarea resize
--- a/.changeset/strange-steaks-sneeze.md
+++ b/.changeset/strange-steaks-sneeze.md
@@ -2,4 +2,4 @@
 '@cloudfour/patterns': minor
 ---
 
-Disable resizing for `c-input--elastic` modifier
+Disable resizing for Elastic Textareas
--- a/.changeset/strange-steaks-sneeze.md
+++ b/.changeset/strange-steaks-sneeze.md
@@ -2,4 +2,4 @@
 '@cloudfour/patterns': minor
 ---
 
-Add `c-input--noResize` modifier
+Disable resizing for `c-input--elastic` modifier
--- a/.changeset/strange-steaks-sneeze.md
+++ b/.changeset/strange-steaks-sneeze.md
@@ -2,4 +2,4 @@
 '@cloudfour/patterns': minor
 ---
 
-Add option to disable Elastic TextArea resize
+Add `c-input--noResize` modifier
--- a/.changeset/strange-steaks-sneeze.md
+++ b/.changeset/strange-steaks-sneeze.md
@@ -2,4 +2,4 @@
 '@cloudfour/patterns': minor
 ---
 
-Add option to disable elastic textarea resize
+Add option to disable Elastic TextArea resize
--- a/src/components/input/elastic-textarea.ts
+++ b/src/components/input/elastic-textarea.ts
@@ -1,3 +1,7 @@
+export interface ElasticTextAreaOpts {
+  disableResize?: boolean;
+}
+
 /**
  * Create Elastic TextArea
  *
@@ -7,19 +11,19 @@
  * method to remove the event listener.
  *
  * @param textarea - the target `textarea` element
- * @param {boolean} [disableResize=false] - Disables the textarea resize functionality if set to `true`
+ * @param {ElasticTextAreaOpts} [elasticTextAreaOpts] - Disables the textarea resize functionality if set to `true`
  */
 export const createElasticTextArea = (
   textarea: HTMLTextAreaElement,
-  disableResize = false
+  elasticTextAreaOpts: ElasticTextAreaOpts = {}
 ) => {
   const minRows = Number(textarea.getAttribute('rows')) || 2;
   let rows = Number(textarea.getAttribute('rows')) || minRows;
   textarea.setAttribute('rows', String(rows));
 
   // Disables the native textarea resize functionality via inline CSS
-  if (disableResize) {
-    textarea.style.resize = 'none';
+  if (elasticTextAreaOpts.disableResize === true) {
+    textArea.style.resize = 'none';
   }
 
   /** Check if the textarea is currently scrolling */
--- a/src/components/input/elastic-textarea.ts
+++ b/src/components/input/elastic-textarea.ts
@@ -9,8 +9,9 @@
  * @param textarea - the target `textarea` element
  */
 export const createElasticTextArea = (textarea: HTMLTextAreaElement) => {
-  // Progressively enhances experience
-  textarea.classList.add('is-elastic');
+  // Provides a CSS hook for JS-only styles
+  const JS_ENABLED_HOOK = 'is-elastic';
+  textarea.classList.add(JS_ENABLED_HOOK);
 
   const minRows = Number(textarea.getAttribute('rows')) || 2;
   let rows = Number(textarea.getAttribute('rows')) || minRows;
@@ -63,7 +64,10 @@ export const createElasticTextArea = (textarea: HTMLTextAreaElement) => {
   };
 
   /** As part of the public API, allow users to remove the event listener */
-  const destroy = () => textarea.removeEventListener('input', update);
+  const destroy = () => {
+    textarea.classList.remove(JS_ENABLED_HOOK);
+    textarea.removeEventListener('input', update);
+  };
 
   // Initialize the textarea with elastic behavior
   textarea.addEventListener('input', update);
--- a/src/components/input/elastic-textarea.ts
+++ b/src/components/input/elastic-textarea.ts
@@ -1,7 +1,3 @@
-export interface ElasticTextAreaOpts {
-  disableResize?: boolean;
-}
-
 /**
  * Create Elastic TextArea
  *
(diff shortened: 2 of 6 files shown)


tylersticka (commented):
@gerardo-rodriguez Have you already considered using [the CSS `resize` property](http://developer.mozilla.org/en-US/docs/Web/CSS/resize) to accomplish this without JavaScript?

gerardo-rodriguez:
> @gerardo-rodriguez Have you already considered using [the CSS `resize` property](http://developer.mozilla.org/en-US/docs/Web/CSS/resize) to accomplish this without JavaScript?

Can you elaborate, @tylersticka? I'm not sure I'm following. 

I'm using the CSS `resize` property, but I'm setting it via JS. I figured since you need to instantiate the functionality via a JS function call, it'd be nice to also pass a boolean prop if you wanted to disable resizing when initializing the Elastic TextArea.

tylersticka:
> I figured since you need to instantiate the functionality via a JS function call, it'd be nice to also pass a boolean prop if you wanted to disable resizing when initializing the Elastic TextArea.

It might be simpler to provide a modifier class for restricting resize on `c-input`?

I think this might be a bit more complicated than what we need as-is.

gerardo-rodriguez:
> It might be simpler to provide a modifier class for restricting resize on `c-input`?

@tylersticka Oh, I see. So instead of allowing consumers to set a `disableResize` prop when calling the `createElasticTextArea()` function, you are suggesting I add a modifer class that can manually be added as needed in the HTML to the `<textarea>`. 

Sure, I can do that!

gerardo-rodriguez:
PR feedback has been addressed, thanks, @tylersticka!

src/components/input/input.stories.mdx
```diff
@@ -127,7 +123,45 @@ If the `rows` attribute is set, it will be used as the minimum height.
...
+      // @see https://github.com/storybookjs/storybook/issues/7786
+      useEffect(() => {
+        createElasticTextArea(document.querySelector('.js-elastic-textarea'));
+      }, []);
+      return input(args);
+    }}
+  </Story>
+</Canvas>
+
+<h3 id="javascript-instructions">JavaScript Instructions</h3>
```
Paul-Hebert, line 164: Nit: is there any reason this switched from Markdown to HTML?

Paul-Hebert (approved):
LGTM! One nitpick inline

src/components/input/input.scss
```diff
@@ -85,6 +85,13 @@
     block-size: auto;
   }
 
+  /**
+   * Removes ability to resize element
+   */
+  &.c-input--noResize {
```
tylersticka, line 91: I believe this should be `c-input--no-resize` to conform to our class naming conventions?

src/components/input/input.stories.mdx
```diff
@@ -127,7 +123,45 @@ If the `rows` attribute is set, it will be used as the minimum height.
...
+    name="Elastic Textarea with No Resize"
+    height="250px"
+    argTypes={{
+      rows: { type: { name: 'number' } },
+      class: { type: { name: 'string' } },
+    }}
+    args={{
+      ...elasticTextAreaConfig,
+      id: 'demo-elastic-no-resize',
+      class: 'c-input--elastic c-input--noResize js-elastic-textarea',
```
tylersticka, line 139: Would it be simpler to apply this style to the existing `--elastic` modifier?

tylersticka: Ooh, or maybe even restrict it to when a `rows` attribute is specified?

```css
.c-input--elastic[rows] {
  resize: none;
}
```

tylersticka: In fact, I wonder if the modifier is necessary at all?

```css
.c-input[rows] {
  height: auto;
  resize: none;
}
```

gerardo-rodriguez: I like where you’re going with this, @tylersticka. I’ll make that change, thank you!

gerardo-rodriguez: @tylersticka I ended up adding [`resize: none` to the `c-input--elastic` modifier](https://github.com/cloudfour/cloudfour.com-patterns/pull/2098/commits/e557648aea4693370bf08815745d6b9e560df37b) based on the following assumptions:
- All Elastic TextAreas should have resizing disabled
- Not all Elastic TextAreas will have a `rows` attribute

Paul-Hebert: That makes sense. I generally wouldn't want to disable resizing on a non-elastic textarea.

Paul-Hebert (approved):

src/components/input/input.stories.mdx
tylersticka: @gerardo-rodriguez I'd like to push back toward my original suggestion for a few reasons:

- As-is, our textareas _ignore_ the `rows` attribute. Try adding `rows="5"` to one of them, it won't work. I consider this a bug. Adding `height: auto` to `[rows]` would fix this.
- Our elastic textareas work by setting `rows`. Can you explain when an elastic textarea wouldn't have that attribute?
- If we disable resizing `c-input--elastic`, we're removing fallback behavior if the JavaScript is never applied. Whereas if we only apply this change when `[rows]` are set, it's basically progressive enhancement.

tylersticka: One thought: Maybe instead of a modifier this should be a _state_ class, like `is-elastic`? That way it can truly be progressively enhanced.

gerardo-rodriguez: > As-is, our textareas ignore the rows attribute. Try adding rows="5" to one of them, it won't work. I consider this a bug. Adding height: auto to [rows] would fix this.

@tylersticka Excellent point!

---

> Our elastic textareas work by setting rows. Can you explain when an elastic textarea wouldn't have that attribute?

My mistake, I was basing my understanding on the [documentation](https://cloudfour-patterns.netlify.app/?path=/docs/components-input--elastic-textarea), which made `rows` sound optional to me:
> If the `rows` attribute is set, it will be used as the minimum height.

Scanning the JS, if `rows` is not set, a default of `2` is used.

---

> One thought: Maybe instead of a modifier this should be a state class, like `is-elastic`? That way it can truly be progressively enhanced.

Great idea, thank you!

src/components/input/input.scss
```diff
@@ -71,18 +71,27 @@
...
+  &.is-elastic {
+    resize: none;
+  }
+
+  /**
+   * Multi-line types should be taller
+   */
+
+  @at-root textarea#{&} {
+    block-size: size.$height-control-multiline;
```
gerardo-rodriguez, line 94: The diff is funny, no changes here.

gerardo-rodriguez:
This PR is ready for another review, thank you!

.changeset/strange-steaks-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cloudfour/patterns': minor
+---
+
+Disable resizing for Elastic Textareas
```
tylersticka, line 5: Can we also add a changeset for fixing the bug where `c-input` did not respect `rows`?

src/components/input/elastic-textarea.ts
```diff
@@ -9,6 +9,9 @@
  * @param textarea - the target `textarea` element
  */
 export const createElasticTextArea = (textarea: HTMLTextAreaElement) => {
+  // Progressively enhances experience
+  textarea.classList.add('is-elastic');
```
tylersticka, line 13: Should the `destroy` method that's returned by this function remove this class?

src/components/input/input.stories.mdx
```diff
@@ -86,12 +86,14 @@ You can also add `c-input` to `<select>` elements, in which case you’ll get a
...
 
-## Elastic textarea
+## Elastic TextArea
 
-Add the `c-input--elastic` modifier class to a `<textarea>` element and follow the [JavaScript instructions](#javascript-instructions) to have it expand as the user types more than will fit.
+Follow the [JavaScript instructions](#javascript-instructions) below to have a `<textarea>` expand as the user types more than will fit.
 
 If the `rows` attribute is set, it will be used as the minimum height.
 
+The experience is progressively enhanced by disabling `resize` when JavaScript is enabled.
```
tylersticka, line 95: Can you elaborate on why this is an enhancement? It sounds a little like "we've enhanced this element by removing a feature," which is a bit confusing.

Or maybe this is an unnecessary detail at this point?

src/components/input/elastic-textarea.ts
gerardo-rodriguez: Yes, it should. Good catch!

src/components/input/input.stories.mdx
gerardo-rodriguez: I'll remove this line. 👍🏽

gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/2098/commits/a09fe661f395079a5ac56ba8df74d8fb63889c80

src/components/input/elastic-textarea.ts
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/2098/commits/a09fe661f395079a5ac56ba8df74d8fb63889c80

.changeset/strange-steaks-sneeze.md
gerardo-rodriguez: https://github.com/cloudfour/cloudfour.com-patterns/pull/2098/commits/e38c51c015fbfee76cb4b49888ae820051ae2430

gerardo-rodriguez:
Latest round of PR feedback has been addressed, thanks, @tylersticka!

tylersticka (approved):
LGTM!

PR #2109, opened 2022-12-20
Add a Draft Email prototype

gerardo-rodriguez:
## Overview

This PR adds a "Draft Email" prototype to save the work I've been doing in WordPress land. The intent is for all of the `<textarea>` + `<button>` HTML + CSS + JavaScript logic to live inside a Gutenberg Custom HTML Block in WordPress. The CSS + JavaScript will be inlined in the Gutenberg block.

### Review Expectations

Even though this is a prototype, I would like a review of the following:
- HTML markup (you can ignore the `o-container` and footer markup, which was only copied/pasted for demo purposes)
- The `runProposedInlineJS()` function JavaScript logic (this is the JS that will be inlined in the Gutenberg Custom HTML block alongside the `<textarea>` and `<button>` elements)
- The CSS (this will also be inlined in the Gutenberg Custom HTML block)

## Screenshots

Screen Shot 2023-01-04 at 10 33 23 AM

### In action

in-action

### Clipboard API copy fail use case

failing

### Clipboard API unavailable use case

The "Copy" button is hidden when the Clipboard API is unavailable:
Screen Shot 2023-01-04 at 4 36 58 PM

### Respecting users' motion preferences

The increased motion animation (vertical shift) only occurs when the user has no motion preference.

reduced-motion

## Testing

Preview: https://deploy-preview-2109--cloudfour-patterns.netlify.app/iframe.html?args=&id=prototypes-draft-email--demo&viewMode=story

- Make sure to review the CSS animation at small and large viewports
- Make sure to also review in Safari (I had to fix a CSS layout issue with the tooltip)

### Motion preferences

- [ ] Change your motion preferences to "Reduce motion" and confirm the intro animation does not move vertically

### "Success" use case

Confirm the following:
- [ ] After activating the "Copy" button:
    - you should see a "✔️ Copied to clipboard" tooltip animate in below the "Copy" button
    - the tooltip should animate out after a few seconds
- [ ] The "Copy" button copies all of the email body text to your clipboard
- [ ] Using VoiceOver, you hear "Copied to clipboard" after activating the "Copy" button
- [ ] The "Draft email" button opens a new email with subject and body content

### "Fail" use case

Paste the following JS into your browser console:
```js
navigator.clipboard.writeText = () =>
  new Promise((resolve, reject) => {
    reject(new Error('Oopsie'));
  });
```

Confirm the following:
- [ ] After activating the "Copy" button:
    - you should see a "Whoops, failed to copy" tooltip below the "Copy" button
    - the tooltip should animate out after a few seconds
- [ ] Using VoiceOver, you hear "Whoops, failed to copy" after activating the "Copy" button
- [ ] The "Draft email" button opens a new email with subject and body content (should still work)

---

- https://github.com/cloudfour/cloudfour.com-wp/issues/945


src/prototypes/draft-email-textarea/example/example.twig
```diff
@@ -0,0 +1,68 @@
...
+
+I’m reaching out to discuss [project, question, challenge]. [Additional details, if any.]
+
+Our budget for this effort is [budget, range, etc.] and we’re interested in wrapping up in [timeframe].
+
+I’m available to chat [potential meeting times], let me know if any of those work for you.
+
+Have a great day!
+
+[Firstname Lastname]</textarea>
```
gerardo-rodriguez, line 20: A quick review here, please.

src/prototypes/draft-email-textarea/example/example.twig
```diff
@@ -0,0 +1,68 @@
...
+			<a href="mailto:info@cloudfour.com?subject=Discuss potential project&body=Hi there,%0D%0A%0D%0AMy name is [Firstname Lastname], I’m a [role] at [organization].%0D%0A%0D%0AI’m reaching out to discuss [project, question, challenge]. [Additional details, if any.]%0D%0A%0D%0AOur budget for this effort is [budget, range, etc.] and we’re interested in wrapping up in [timeframe].%0D%0A%0D%0AI’m available to chat [potential meeting times], let me know if any of those work for you.%0D%0A%0D%0AHave a great day!%0D%0A%0D%0A[Firstname Lastname]" class="c-button c-button--secondary">
+				<span class="c-button__extra">
+					<svg width="24" height="24" viewBox="0 0 24 24" aria-hidden="true" class="c-icon">
+						<path d="M12 13.31l11-7V6a2 2 0 00-2-2H3a2 2 0 00-2 2v.31z"></path>
+						<path d="M13.07 15a2 2 0 01-2.14 0L1 8.69V18a2 2 0 002 2h18a2 2 0 002-2V8.69z"></path>
+					</svg>
+				</span>
+				<span class="c-button__content"> Draft email </span>
+			</a>
+		</div>
```
gerardo-rodriguez, line 42: A quick review here, please.

src/prototypes/draft-email-textarea/example/example.twig
```diff
@@ -0,0 +1,68 @@
...
+
+				// Select the text on click for a better UX
+				draftEl.select();
+				draftEl.setSelectionRange(0, 99999); // For mobile devices
+
+				copyToClipboard(draftEl.value);
+			};
+
+			document.getElementById("copy-btn").addEventListener("click", onCopyClick);
+		</script>
```
gerardo-rodriguez, line 66: A quick review here, please.

src/prototypes/draft-email-textarea/example/example.twig
```diff
@@ -0,0 +1,68 @@
...
+
+I’m reaching out to discuss [project, question, challenge]. [Additional details, if any.]
+
+Our budget for this effort is [budget, range, etc.] and we’re interested in wrapping up in [timeframe].
+
+I’m available to chat [potential meeting times], let me know if any of those work for you.
+
+Have a great day!
+
+[Firstname Lastname]</textarea>
```
gerardo-rodriguez, line 20: A quick review here, please.

tylersticka:
> I could not get the Elastic TextArea JS to load in the context of the prototype.

@gerardo-rodriguez Did you try using `useEffect` to initialize the textarea?

You can refer to the elastic textarea stories for an example: https://github.com/cloudfour/cloudfour.com-patterns/blob/c07e1d5053274dd0016a27ec1984f4366f1bc14e/src/components/input/input.stories.mdx?plain=1#L115-L121

src/prototypes/draft-email-textarea/example/example.twig
```diff
@@ -0,0 +1,64 @@
...
+
+		<div class="o-button-group o-button-group--grow u-space-block-start-n2">
```
tylersticka, line 18: Is there a reason we're using raw HTML here instead of Liquid templates?

src/prototypes/draft-email-textarea/example/example.twig
```diff
@@ -0,0 +1,64 @@
...
+				<span class="c-button__extra">
+					<svg width="24" height="24" viewBox="0 0 24 24" aria-hidden="true" class="c-icon">
+						<path d="M16 16v5c0 .55-.45 1-1 1H3c-.55 0-1-.45-1-1V9c0-.55.45-1 1-1h5" fill="none" stroke-linecap="round" stroke-width="2"></path>
+						<rect x="8" y="2" width="14" height="14" rx="1" ry="1" fill="none" stroke-linecap="round" stroke-width="2"></rect>
+					</svg>
+				</span>
+				<span class="c-button__content"> Copy </span>
+			</button>
+
+			<a href="mailto:info@cloudfour.com?subject=Discuss potential project&body=Hi there,%0D%0A%0D%0AMy name is [Firstname Lastname], I’m a [role] at [organization].%0D%0A%0D%0AI’m reaching out to discuss [project, question, challenge]. [Additional details, if any.]%0D%0A%0D%0AOur budget for this effort is [budget, range, etc.] and we’re interested in wrapping up in [timeframe].%0D%0A%0D%0AI’m available to chat [potential meeting times], let me know if any of those work for you.%0D%0A%0D%0AHave a great day!%0D%0A%0D%0A[Firstname Lastname]" class="c-button c-button--secondary">
```
tylersticka, line 29: If you wanted to keep things DRY, you could pass the textarea content to the template from the story definition as a string, then output it once in the textarea, and once here.

Twig has a `url_encode` filter that comes in handy for things like this: https://twig.symfony.com/doc/3.x/filters/url_encode.html

tylersticka (commented):
Overall this is looking good!

1. Do you think it might be possible for the "Copy" button to momentarily say "Copied" (maybe with a checkmark icon) for a few beats after click?
2. If the clipboard API isn't available, can we show the "draft email" button only?

gerardo-rodriguez:
> @gerardo-rodriguez Did you try using `useEffect` to initialize the textarea?

Thanks, @tylersticka! That's a good idea.

Using your idea, I'm now trying to figure out how to implement your suggestion within [`src/prototypes/draft-email-textarea/draft-email-textarea.stories.js`](https://github.com/cloudfour/cloudfour.com-patterns/pull/2109/files#diff-e3cdc90edc1d320622f53bd78d8488f8dd14d6816b27fa671d205381a69cff6c).

I copied the contents from other prototypes so not sure how to add a `useEffect` at the moment. Going to start searching to see if I find anything helpful.

src/prototypes/draft-email-textarea/example/example.twig
gerardo-rodriguez: Oh, how do I do that from the WP Gutenberg editor?

src/prototypes/draft-email-textarea/example/example.twig
gerardo-rodriguez: Thanks that sounds like an excellent idea! 

Though, I'm having a hard time wrapping my head around your suggestion within the context of the WP Gutenberg editor. 🤔

gerardo-rodriguez:
@tylersticka Might you have time for a quick chat today? Maybe around 10:30 or 11 am (after Santi's first nap)? 

I feel I'm getting confused about the implementation of this within the WP Gutenberg editor. I'd like to share with you what I've got in the Gutenberg editor to make sure I'm heading down a good path. 

Thank you!

src/prototypes/draft-email-textarea/example/example.twig
gerardo-rodriguez: Oh, now I follow! You asked why I didn't use Liquid templates _in the prototype_. My bad, I misunderstood. LOL

My brain was stuck in WP admin land so I did a straight copy + paste of the code I had over there. I never considered using Liquid templates in the prototype because it just didn't occur to me, but I should!

gerardo-rodriguez:
Placing back into "draft" status while I look to address feedback.

gerardo-rodriguez:
Quick update…
> 1. Do you think it might be possible for the "Copy" button to momentarily say "Copied" (maybe with a checkmark icon) for a few beats after click?
> 2. If the clipboard API isn't available, can we show the "draft email" button only?

1. In progress. I'm trying to make sure the UX is accessible. My first couple of attempts/tests with VoiceOver are less than ideal UX.
2. I haven't given this deep thought, but yes, I assume this is possible. 👍🏽

gerardo-rodriguez:
Update…
> 1. Do you think it might be possible for the "Copy" button to momentarily say "Copied" (maybe with a checkmark icon) for a few beats after click?
> 2. If the clipboard API isn't available, can we show the "draft email" button only?

1. Still in progress
2. I added logic to handle this use case 👍🏽

gerardo-rodriguez:
Thanks for the feedback, @tylersticka! This PR is ready for another review. 👍🏽

src/prototypes/draft-email-textarea/demo/demo.twig
```diff
@@ -0,0 +1,55 @@
...
+<div class="o-container o-container--pad o-container--prose">
+	<div class="o-container__content o-rhythm o-rhythm--default o-rhythm--generous-headings">
+		{# Draft email message #}
+		<textarea
+			id="email-draft"
+			class="c-input u-pad-1 js-elastic-textarea"
+			spellcheck="false"
+		>
+			{{ draftMessage }}
+		</textarea>
```
tylersticka, line 11: Can we omit the whitespace so the textarea has enough room all around?

tylersticka (commented):
Overall, this is looking good! Three nitpicks I see:

- The textarea has several leading spaces.
- After copying the text, the text remains selected, which looks a bit funky.
- When you edit the textarea content, it does not update the "draft email" link content accordingly.

gerardo-rodriguez:
Thanks, @tylersticka!

> * The textarea has several leading spaces.

Oops, silly me. 🙃 

> * After copying the text, the text remains selected, which looks a bit funky.

Interesting. I'm going to look around a bit to see if I find "best UX" advice. Different experiences I've seen for copy to clipboard:
- The copied text is selected and stays selected (the current implementation)
- The copied text is not selected ever
- On _hover_ of the "copy" btn, the text that will be copied is highlighted

Have you found any resources for the best copy-to-clipboard UX? I'd be curious to read.

> * When you edit the textarea content, it does not update the "draft email" link content accordingly.

LOL Good catch. Just a _slight_ oversight. 😅

src/prototypes/draft-email-textarea/demo/demo.twig
gerardo-rodriguez: Done! https://github.com/cloudfour/cloudfour.com-patterns/pull/2109/commits/7b529eef8ea148f1c0f488a5211a91cdc04d9904

tylersticka:
> Have you found any resources for the best copy-to-clipboard UX? I'd be curious to read.

I haven't. I just personally find the current experience a little jarring. It took me from a state where I wasn't focused on the textarea at all to one where I had made a selection of the text, which felt unexpected to me.

gerardo-rodriguez:
Update:

> * The textarea has several leading spaces.

Fixed!

> * After copying the text, the text remains selected, which looks a bit funky.

The text selection logic was removed, but the audible UX is a bit funky now. TBD how we proceed.

> * When you edit the textarea content, it does not update the "draft email" link content accordingly.

Fixed!

gerardo-rodriguez:
@tylersticka, I've updated the PR! Ready for another review, appreciate it. Thank you! 🙂

src/prototypes/draft-email-textarea/demo/demo.twig
```diff
@@ -0,0 +1,48 @@
...
+					content_start_icon: 'envelope',
+					href: "mailto:info@cloudfour.com?subject=#{subject}&body=#{message}",
+				} only %}
+			{% endblock %}
+		{% endembed %}
+
+		{# "Copied" message shown when the "copy" btn is activated #}
+		<p
+			id="copied-action-message"
+			role="status"
```
gerardo-rodriguez, line 38: Based on my understanding, a `role` of `"status"` should be used for this message.

See: https://www.w3.org/WAI/WCAG22/Understanding/status-messages.html

gerardo-rodriguez:
Putting back into "draft" mode. I just realized `navigator.clipboard.writeText()` returns a promise. I need to modify the JS logic a bit.

gerardo-rodriguez:
Okay, PR has been updated. Ready for another review, @tylersticka. Thank you! 😃

src/prototypes/draft-email-textarea/demo/demo.twig
```diff
@@ -0,0 +1,61 @@
...
+			Copied to clipboard
+		</p>
+
+		{# The copy action "fail" message #}
+		<p
+			id="copy-fail-message"
+			role="status"
+			hidden
+		>
+			Sorry, copy to clipboard failed
```
tylersticka, line 54: Can this be shortened a bit?

```suggestion
			Whoops, failed to copy
```

tylersticka (commented):
Asked some design questions in Slack, otherwise a small verbiage suggestion left inline.

src/prototypes/draft-email-textarea/demo/demo.scss
```diff
@@ -0,0 +1,131 @@
...
+  border: ms.step(-4) solid $tooltip-color-background;
+  border-radius: size.$border-radius-medium;
+  background: $tooltip-color-background;
+  color: var(--theme-color-text-emphasis);
+  box-shadow: 0 0 0 size.$edge-medium $tooltip-color-border,
+    size.$edge-large size.$edge-large 0 size.$edge-medium #{sass-color.change(
+        color.$brand-primary-darker,
+        $alpha: 0.2
+      )};
+  inline-size: max-content; // Is this the right way to handle this?
```
gerardo-rodriguez, line 30: I'm unsure if this is the best way to handle the following bug. Notice the tooltip text breaks into two lines (it should be one line):

Screen Shot 2023-01-09 at 5 17 53 AM


It seems to be working. 🤷🏽

src/prototypes/draft-email-textarea/demo/demo.twig
```diff
@@ -0,0 +1,246 @@
...
+		<p>The size of an engagement varies based on project requirements, typically in the range of $XX,000 to $XXX,000. We are always happy to discuss ongoing retainer agreements and unique partnerships.</p>
+		<p>If you aren’t sure how to start, you can modify this example email:</p>
+
+		{# Draft email message #}
+		<textarea
+      id="email-draft"
+      class="c-input u-pad-1 js-elastic-textarea"
+      spellcheck="false"
+      data-message-subject="{{ draftSubject }}"
+    >{{ draftMessage }}</textarea>
```
gerardo-rodriguez, line 21: This should be reviewed.

src/prototypes/draft-email-textarea/demo/demo.twig
```diff
@@ -0,0 +1,246 @@
...
+				{# The "Draft email" button #}
+				{% include '@cloudfour/components/button/button.twig' with {
+					label: 'Draft email',
+					id: 'draft-email-btn',
+					class: 'c-button--secondary',
+					content_start_icon: 'envelope',
+					href: "mailto:info@cloudfour.com?subject=#{subject}&body=#{message}",
+				} only %}
+			{% endblock %}
+		{% endembed %}
```
gerardo-rodriguez, line 74: This should be reviewed.

src/prototypes/draft-email-textarea/demo/demo.scss
```diff
@@ -0,0 +1,146 @@
+@use '../../../compiled/tokens/scss/size';
```
gerardo-rodriguez, line 1: This file should be reviewed.

gerardo-rodriguez, line 1: This file should be reviewed.

src/prototypes/draft-email-textarea/draft-email-textarea.stories.mdx
```diff
@@ -0,0 +1,48 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
```
gerardo-rodriguez, line 1: This file can be reviewed.

src/prototypes/draft-email-textarea/js/draft-email-textarea.js
```diff
@@ -0,0 +1,128 @@
+/**
+ * The contents of this function will be inlined in a Gutenberg Custom HTML block
+ */
+export const runProposedInlineJS = () => {
```
gerardo-rodriguez, line 4: All of the JS in the `runProposedInlineJS()` function should be reviewed.

src/tokens/ease/ease.js
```diff
@@ -17,5 +17,17 @@ module.exports = {
...
+      comment: 'https://easings.net/#easeInOutSine',
+    },
+    out_sine: {
+      value: 'cubic-bezier(0.61, 1, 0.88, 1)',
+      comment: 'https://easings.net/#easeOutSine',
+    },
+    in_back: {
+      value: 'cubic-bezier(0.36, 0, 0.66, -0.56)',
+      comment: 'https://easings.net/#easeInBack',
+    },
```
gerardo-rodriguez, line 31: These additions should be reviewed.

gerardo-rodriguez:
PR is ready for another review. Thanks, @tylersticka! 😄

src/objects/button-group/button-group.scss
```diff
@@ -23,3 +23,12 @@
...
 }
+
+.o-button-group__item {
+  position: relative;
+
+  // TODO This should not live here
+  .c-button {
+    inline-size: 100%;
+  }
+}
```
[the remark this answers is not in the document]
tylersticka: @gerardo-rodriguez Investigating this broke my brain, but the issue seems to be with the inline `padding` on some child elements but not all. The button has padding, [which affects the flexbox width](https://stackoverflow.com/questions/32099624/flexbox-item-widths-are-affected-by-padding), but the `__item` does not, so it takes up less space.

I found [a few different solutions](https://codepen.io/tylersticka/pen/YzjZwbe/23ff47f16f4ed7a5735e01d31f942a47?editors=1100)...

## Option A: Wrap both buttons in `__item`

```html
<div class="o-button-group__item"><!-- copy button, tooltip, etc. --></div>
<div class="o-button-group__item"><!-- draft email link --></div>
```

- **Pros:** Simple
- **Cons:** More markup

(This option feels the most straightforward to me.)

## Option B: Negate inline padding for children of the grow modifier

```css
.o-button-group--grow > * {
  flex: 1 1 0;
  padding-inline: 0;
}
```

- **Pros:** Simple, less markup
- **Cons:** Button content could run into edges at odd sizes

## Option C: Use `grid` instead

```css
.o-button-group--grow {
  display: grid;
  grid-template-columns: repeat(auto-fit, minmax(260px, 1fr));
}

.o-button-group--grow > .o-button-group__item {
  display: grid;
}
```

- **Pros:** Simplifies CSS, less markup
- **Cons:** Relies on an arbitrary "minimum" button width instead of wrapping based on content

## Re: `inline-size` question

No, I don't think we should use `inline-size` here. I think you should mimic whatever layout mechanism is being used by the parent.

```css
.o-button-group__item {
  display: flex;
  position: relative;
}

.o-button-group--grow .o-button-group__item > * {
  flex: 1;
}
```

src/prototypes/draft-email-textarea/demo/demo.scss
```diff
@@ -0,0 +1,141 @@
...
+    color.$brand-primary-darker,
+    $alpha: 0.4
+  )};
+
+///
+/// 1. Horizontally centers tooltip
+/// 2. Positions tooltip vertically in "end" position (below button)
+///
+
+.c-tooltip {
```
tylersticka, line 27: Because this isn't a _full_ component yet, can we precede all these class names with an underscore [per Harry Roberts' namespacing convention](https://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/#hack-namespaces-_)?

src/prototypes/draft-email-textarea/demo/demo.scss
tylersticka: If your goal is to prevent the text from breaking, would `white-space: nowrap` be more explanatory?

src/prototypes/draft-email-textarea/demo/demo.scss
```diff
@@ -0,0 +1,141 @@
...
+    size.$edge-large size.$edge-large 0 size.$edge-medium #{sass-color.change(
+        color.$brand-primary-darker,
+        $alpha: 0.2
+      )};
+  inline-size: max-content; // Is this the right way to handle this?
+  inset-inline-start: 50%; // 1
+  opacity: 0;
+  position: absolute; // 1
+  z-index: z-index.$alert;
+  inset-block-start: 2.45em; // 2
```
tylersticka, line 42: Instead of making this a hard value, could you position it to `100%` and then negate the border size using margin?

Example: https://codepen.io/tylersticka/pen/oNMZxBq/2c967ba0806e7ee98fec9ed5a6dfd295

src/objects/button-group/button-group.scss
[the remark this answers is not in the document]
gerardo-rodriguez: Thanks for digging into this, @tylersticka! I'm going to roll with **Option A**.

src/prototypes/draft-email-textarea/demo/demo.scss
gerardo-rodriguez: Great idea, thank you!

src/prototypes/draft-email-textarea/demo/demo.twig
```diff
@@ -0,0 +1,248 @@
...
+		<p>The size of an engagement varies based on project requirements, typically in the range of $XX,000 to $XXX,000. We are always happy to discuss ongoing retainer agreements and unique partnerships.</p>
+		<p>If you aren’t sure how to start, you can modify this example email:</p>
+
+		{# Draft email message #}
+		<textarea
+			id="email-draft"
+			class="c-input u-pad-1 js-elastic-textarea"
+			spellcheck="false"
+			data-message-subject="{{ draftSubject }}"
+		>{{ draftMessage }}</textarea>
```
gerardo-rodriguez, line 21: The `<textarea>` should be reviewed.

src/prototypes/draft-email-textarea/demo/demo.twig
```diff
@@ -0,0 +1,248 @@
...
+					{% include '@cloudfour/components/button/button.twig' with {
+						label: 'Draft email',
+						id: 'draft-email-btn',
+						class: 'c-button--secondary',
+						content_start_icon: 'envelope',
+						href: "mailto:info@cloudfour.com?subject=#{subject}&body=#{message}",
+					} only %}
+				</div>
+			{% endblock %}
+		{% endembed %}
```
gerardo-rodriguez, line 76: The buttons and tooltips should be reviewed.

gerardo-rodriguez:
@tylersticka The PR has been updated, and your feedback has been addressed. Thank you!

tylersticka (commented):
The experience LGTM, I'd love for a developer to review the code! Thanks, Gerardo!

src/prototypes/draft-email-textarea/demo/demo.scss
```diff
@@ -0,0 +1,142 @@
...
+/// 1. Horizontally centers tooltip
+/// 2. Positions tooltip vertically (below button)
+///
+
+._c-tooltip {
+  background: $tooltip-color-background;
+  border: ms.step(-4) solid $tooltip-color-background;
+  border-radius: size.$border-radius-medium;
+  box-shadow: 0 0 0 size.$edge-medium $tooltip-color-border,
+    size.$edge-large size.$edge-large 0 size.$edge-medium #{sass-color.change(
```
Paul-Hebert, line 33: Nit: there's a lot going on here. It might be nice to move the `sass-color` stuff to a variable declaration up above to simplify this line.

e.g.

```scss
$tooltip-color-shadow: #{sass-color.change(
    color.$brand-primary-darker,
    $alpha: 0.4
  )};
```

Paul-Hebert: (Then on this line you could use `$tooltip-color-shadow` without worrying about the color manipulation)

src/prototypes/draft-email-textarea/demo/demo.scss
```diff
@@ -0,0 +1,142 @@
...
+    color.$brand-primary-darker,
+    $alpha: 0.4
+  )};
+
+///
+/// 1. Horizontally centers tooltip
+/// 2. Positions tooltip vertically (below button)
+///
+
+._c-tooltip {
```
Paul-Hebert, line 28: Nit: I'm not used to this class syntax. Is it just because this is "prototype" code and not a real component? (If so, that's fine. I just wanted to make sure this was intentional)

src/prototypes/draft-email-textarea/demo/demo.scss
```diff
@@ -0,0 +1,142 @@
...
+  /// Background tooltip arrow (lighter color)
+  ///
+  /// 1. Combined with background arrow, matches border-width of tooltip body
+  /// 2. Position arrow above tooltip body
+  ///
+
+  &::after {
+    border-color: transparent transparent $tooltip-color-background;
+    border-width: 10px; // 1
+    inset-block-start: -28px; // 2
```
Paul-Hebert, line 84: 🤔 should we be using `rem`/`em` units here? Does this work well wth user scaling/zooming when we use px?

Paul-Hebert: This is a total nit, and I'm not sure if this is actually more readable or not so feel free to ignore. But I wonder if this could be made more DRY/readable using custom props? Something like this?

```scss
// Add two arrows to the top of tooltips
&:before,
&:after {
    border-color: transparent transparent var(--arrow-color, $tooltip-color-border);
    border-width: var(--arrow-size, 13px);
    inset-block-start: var(--arrow-offset, -35px);
}

// Customize the color, size, and placement of the second arrow
&:after {
    --arrow-color: $tooltip-color-background;
    --arrow-size: 10px;
    --arrow-offset: -28px;
}
```

(Some of the Sass variables may need to be wrapped like so to work. I can't remember for sure: `#{$sass-var}`)

Feel free to ignore this if it feels less readable. I'm not sure...

src/prototypes/draft-email-textarea/demo/demo.scss
```diff
@@ -0,0 +1,142 @@
...
+  }
+}
+
+///
+/// Tooltip intro animation
+///
+
+@keyframes tooltipIntro {
+  from {
+    animation-timing-function: ease.$out_sine;
```
Paul-Hebert, line 121: Oh wow, I didn't realize you could change the timing function for a specific keyframe. That's _very_ interesting...

src/prototypes/draft-email-textarea/demo/demo.scss
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  50% {
+    animation-timing-function: ease.$in_out_sine;
+    transform: translate(-50%, 15%);
```
Paul-Hebert, line 126: I was confused for a second about why the -50% horizontal translation was necessary. A comment might be helpful here.

src/prototypes/draft-email-textarea/js/draft-email-textarea.js
```diff
@@ -0,0 +1,129 @@
...
+  // The elements all will be a part of the same Gutenberg Custom HTML block.
+  // I chose to use IDs since the HTML and JS are scoped to the same Gutenberg
```
Paul-Hebert, line 6: Total nit: if we're building this as a block it's _possible_ for the block to be included multiple times on one page. If that were the case then we'd have issues with duplicate IDs. That said, I think that's a _really_ unlikely use case, so I'm fine ignoring it here.

src/prototypes/draft-email-textarea/js/draft-email-textarea.js
```diff
@@ -0,0 +1,129 @@
...
+    tooltipEl.classList.add('is-animating-intro');
+    // Clear any existing timeouts
+    if (hideTooltipDelayTimeoutId) clearTimeout(hideTooltipDelayTimeoutId);
+    // Hide the tooltip after a delay
+    hideTooltipDelayTimeoutId = setTimeout(() => {
+      tooltipEl.classList.remove('is-animating-intro');
+      tooltipEl.classList.add('is-animating-outro');
+    }, hideTooltipDelay);
+    // Listen for when the tooltip animation ends
+    tooltipEl.addEventListener('animationend', onTooltipAnimationEnd);
```
Paul-Hebert, line 54: Just a heads up, if it's possible for the tooltip to contain an element with an animation, this listener could be fired by the tooltip's child element finishing its animation. I don't _think_ that's an issue here, but I wanted to flag it in case it's possible we add animated children in the future, since this can be a tricky issue to debug.

src/prototypes/draft-email-textarea/js/draft-email-textarea.js
```diff
@@ -0,0 +1,129 @@
...
+  const requirements = [
+    copyBtn,
+    draftEmailBtn,
+    draftEl,
+    copySuccessTooltipEl,
+    copyFailTooltipEl,
+    navigator,
+    navigator.clipboard,
+    navigator.clipboard.writeText,
+  ];
```
Paul-Hebert, line 24: Nit: this might be a little clearer:
```suggestion
  const requiredElements = [
    copyBtn,
    draftEmailBtn,
    draftEl,
    copySuccessTooltipEl,
    copyFailTooltipEl,
  ];

  const requiredBrowserFunctionality = [
    navigator,
    navigator.clipboard,
    navigator.clipboard.writeText,
  ]

  const requirements = [
    ...requiredElements,
    ...requiredBrowserFunctionality
  ];
```

(At first glance I skimmed this and assumed it was only checking for the elements. Later on I wondered how we were checking for the necessary functionality.)

Paul-Hebert (commented):
@gerardo-rodriguez , overall this looks great! I left some comments inline, but none of them are blocking.

I did notice one visual issue that we may want to fix before merging though. On small screens in Firefox, there's a very noticeable border between the tooltip and its arrow:

image

Maybe we need to shift the arrow down or something? For some reason I see this on small screens but not large screens.

src/prototypes/draft-email-textarea/js/draft-email-textarea.js
gerardo-rodriguez: I like it!

src/prototypes/draft-email-textarea/js/draft-email-textarea.js
gerardo-rodriguez: Thanks, @Paul-Hebert. Valid concern.

As I understand, a Custom HTML block cannot be included multiple times. I'm not creating a "custom" reusable block. Instead, I'm only inlining HTML/CSS/JS in the Custom HTML block. The only way that there might be duplicates is if I copy-paste whatever is inside the Custom HTML block into another Custom HTML block.

This use case is a one-off for now, so I don't see a need to copy-paste into various Custom HTML blocks.

src/prototypes/draft-email-textarea/demo/demo.scss
gerardo-rodriguez: Yes! I didn't know this either, but it makes total sense! 🎉

src/prototypes/draft-email-textarea/demo/demo.scss
gerardo-rodriguez: > 🤔 should we be using rem/em units here? Does this work well wth user scaling/zooming when we use px?

OMG, of course, using `rem`/`em` makes more sense, so it can handle zooming. Thank you! 🙏🏽

src/prototypes/draft-email-textarea/demo/demo.scss
gerardo-rodriguez: Good call!

src/prototypes/draft-email-textarea/demo/demo.scss
gerardo-rodriguez: @Paul-Hebert Yes, indeed! 

Props to @tylersticka for suggesting this [underscore prefix formatting](https://github.com/cloudfour/cloudfour.com-patterns/pull/2109#discussion_r1065157734).

gerardo-rodriguez:
Thanks for the review and feedback, @Paul-Hebert. Much appreciated! 

> I did notice one visual issue that we may want to fix before merging though. On small screens in Firefox, there's a very noticeable border between the tooltip and its arrow:

Good catch. I've refactored the tooltip arrow to address this bug you found. I've also addressed your PR feedback. Please take another look, thank you! 😄

src/prototypes/draft-email-textarea/demo/demo.scss
```diff
@@ -0,0 +1,150 @@
...
+    animation-timing-function: ease.$out_sine;
+  }
+}
+
+///
+/// Tooltip intro animation
+///
+/// 1. Only the Y value is changing, but the X value of `-50%` is included
+///    because I can't only set the Y value without affecting the X value since
+///    there can only be a single `transform` property.
```
Paul-Hebert, line 124: This  comment definitely helps! The only thing I would add is why there's a `-50%` in the first place. Maybe something like this?

```suggestion
/// 1. Only the Y value is changing. 
/// Before the animation the horizontal translation is set to `-50%` to center the tooltip.
///    We don't want to overwrite the X transformation when we add the Y transformation.
///    Since there can only be a single `transform` property
///    we need to include the X value alongside the Y value..
```

src/prototypes/draft-email-textarea/demo/demo.scss
Paul-Hebert: I added another potential comment. Not sure if it's better or worse.

gerardo-rodriguez:
I've placed this PR back into "draft" mode. I just realized I did not handle reduced motion UX!

I'll have it ready for review again soon. 👍🏽

gerardo-rodriguez:
@Paul-Hebert This PR is ready for another review. Thank you!

Paul-Hebert:
@gerardo-rodriguez , this is looking great!

Unfortunately, I'm still seeing a very slight divider between the arrow and the tooltip in Safari on small screens

image

I wonder if it needs to be shifted another pixel down or something?

Paul-Hebert (approved):
Looks great! :shipit:

PR #2111, opened 2022-12-20
Highlight theme

### LINKED ISSUE #2107: Create Highlight theme in preparation for server-side rendering
We opted for [Prism](https://prismjs.com/) for our [syntax highlighting](https://cloudfour-patterns.netlify.app/?path=/docs/vendor-prism--bash-shell) because it was a lot more performant on the front-end than [Highlight.js](https://highlightjs.org/), and none of the server-side options we found at that time were of sufficient quality or usability.

Last week, [Chris Coyier wrote](https://chriscoyier.net/2022/12/15/this-wordpress-plugin-for-code-highlighting-kicks-ass/) about [a plugin](https://wordpress.org/plugins/syntax-highlighting-code-block/) that uses [highlight.php](https://github.com/scrivo/highlight.php) to apply syntax highlighting on the server. It appears to be well documented [and open source](https://github.com/westonruter/syntax-highlighting-code-block).

The only barrier to switching is that we have a tailor-made Prism theme we've received positive feedback on. It would be great as a first step to port this theme to Highlight, being careful to export the resultant stylesheet separately (so it can be included only on pages that need it, a feature of the plugin) and also being careful not to bundle highlight.js as a toolkit dependency (though it can be used in Storybook for the purposes of demonstrating the theme).

tylersticka:
## Overview

Introduces a custom [Highlight](https://highlightjs.org/) theme to enable us to use server side syntax highlighting in the future.

This is a breaking change because the client-side Prism dependency will override the server-side Highlight markup, so this drops the Prism theme and dependency from our patterns entirely.

This also includes the minimal styles necessary to support the WordPress Syntax-highlighting Code Block plugin.

The Liquid code sample was removed because Highlight does not support it. Some code samples were updated to test highlighting that Highlight supports that Prism did not.

On the WordPress side, implementation will involve:

- Updating patterns
- Removing Prism scripts and enqueuing
- Installing the aforementioned plugin
- Adding filters for `the_content` when `has_blocks` is false, `comment_text` and `render_block` for `jetpack/markdown` to manually apply the plugin logic

## Screenshots

Screenshot 2023-01-26 at 4 21 46 PM

Screenshot 2023-01-26 at 4 21 59 PM

Screenshot 2023-01-26 at 4 22 09 PM

## Testing

[Review the deploy preview docs and examples](https://deploy-preview-2111--cloudfour-patterns.netlify.app/?path=/docs/vendor-highlight--theme)

---

- See #2107


src/base/_defaults.scss
```diff
@@ -145,10 +145,13 @@ pre {
...
  * a link, we offset their color from surrounding content.
+ *
+ * We're disabling a selector rule here because the alternative is pretty
+ * ridiculously verbose. We should consider revising our selector rules.
  */
 
-*:not(pre):not(a) > code,
-*:not(pre):not(a) > kbd {
+/* stylelint-disable-next-line selector-not-notation */
+:not(pre, pre *, a) > :is(code, kbd) {
```
tylersticka, line 154: I could not figure out how to get our linter to be okay with `:not(pre *)`, so I decided to keep the less verbose selector and simplify everything down to one line.

src/vendor/highlight/_theme.scss
```diff
@@ -0,0 +1,85 @@
...
+    sasscolor.mix(color.$base-orange-lighter, color.$base-green-lighter),
+    $saturation: +10%
+  );
+}
+
+/// Syntax-Highlighted Code Block Plugin Customizations
+
+.shcb-language {
+  @include a11y.sr-only;
+}
```
tylersticka, line 85: I had experimented with a more interesting visual display, which I showed in our last design team meeting, but in practice there were a _lot_ of bugs to resolve. I decided it was best to maintain the current appearance and to consider enhancements for a later date.

src/vendor/highlight/_theme.scss
```diff
@@ -0,0 +1,85 @@
+@use '../../compiled/tokens/scss/color';
+@use '../../compiled/tokens/scss/opacity';
+@use '../../mixins/a11y';
+@use 'sass:color' as sasscolor;
+
+/// Cloud Four theme for Prism syntax highlighting.
```
Paul-Hebert, line 6:
```suggestion
/// Cloud Four theme for Highlight syntax highlighting.
```

src/vendor/highlight/_theme.scss
Paul-Hebert: That makes sense. Can we add a bit of this info in a comment? Something along these lines?

```suggestion
/// Syntax-Highlighted Code Block Plugin Customizations

///
/// The plugin adds some content showing the language the code is written in.
/// This is helpful info, and we explored an interesting visual display,
/// but in practice, there were a lot of bugs to resolve.
///
/// For now, we hide it visually and expose it to screen readers.
///
.shcb-language {
  @include a11y.sr-only;
}
```

src/vendor/highlight/demo/theme.twig
```diff
@@ -0,0 +1,5 @@
+<pre{% if shcb %}
+  aria-describedby="shcb-language-demo"
+  data-shcb-language-name="{{languageName}}"
+  data-shcb-language-slug="{{language}}"
+{% endif %}>{% if shcb %}<div>{% endif %}<code class="language-{{name}}">{% if html %}{{html|raw}}{% else %}{{content}}{% endif %}</code>{% if shcb %}</div><small class="shcb-language" id="shcb-language-demo"><span class="shcb-language__label">Code language:</span> <span class="shcb-language__name">{{languageName}}</span> <span class="shcb-language__paren">(</span><span class="shcb-language__slug">{{language}}</span><span class="shcb-language__paren">)</span></small>{% endif %}</pre>
```
Paul-Hebert, line 5: Non-blocking nit: If this is all on one line because of white space in the `pre` element, we could explore formatting it across multiple lines and using Twig's whitespace trimming functionality: https://symfony.com/blog/better-white-space-control-in-twig-templates

(But I think this is fine for docs)

Paul-Hebert (approved):
This looks great, nice job!

I left a few small comment/formatting suggestions inline.

src/vendor/highlight/demo/theme.twig
tylersticka: Totally. I'll admit I was tripping over the whitespace trimming a bit and decided to just be lazy about it for the sake of shipping sooner. 😅

Paul-Hebert (approved):
Latest commits look good!

PR #2129, opened 2023-02-07
Support WordPress/Timber comment content filters

tylersticka:
## Overview

Our comment component was outputting `comment_content` instead of `content`. While this works fine, it prevents any WordPress filters from applying to the comment content, which prevents us from easily enhancing comments with features like server-side syntax highlighting.

I've left the old `comment_content` property as a `default` so this shouldn't be a breaking change.

## Testing

Compare [the deploy preview comments docs](https://deploy-preview-2129--cloudfour-patterns.netlify.app/?path=/docs/components-comment--single) to [the live version](https://cloudfour-patterns.netlify.app/?path=/docs/components-comment--single). Though the randomized content may differ, functionally there should be no change.

---

- See https://github.com/cloudfour/cloudfour.com-patterns/issues/2107

### DIFF
--- a/.changeset/eight-experts-refuse.md
+++ b/.changeset/eight-experts-refuse.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Add support for a `content` property (in addition to `comment_content`) so the comment component will apply content filters in WordPress/Timber projects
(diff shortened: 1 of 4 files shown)


Paul-Hebert (approved):
:shipit:

PR #2133, opened 2023-02-10
Use flexbox gap instead of negative margins

tylersticka:
## Overview

When these two patterns were authored, support for `gap` properties in Flexbox elements was not great. Now [it's up to 92.8%](https://caniuse.com/flexbox-gap), so it feels like we can simplify these a bit.

## Testing

Review these layout objects for regressions:

- [x] [Button Group](https://deploy-preview-2133--cloudfour-patterns.netlify.app/?path=/docs/objects-button-group--default-story)
- [x] [Inline List](https://deploy-preview-2133--cloudfour-patterns.netlify.app/?path=/docs/objects-list--inline)

### DIFF
--- a/.changeset/curvy-ears-rule.md
+++ b/.changeset/curvy-ears-rule.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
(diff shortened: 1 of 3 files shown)


spaceninja (approved):
Code looks good and works as described in Chrome, Edge, Firefox, Safari, and Mobile Safari

PR #2138, opened 2023-02-16
Support more SHCB features

tylersticka:
## Overview

Adds support for the following optional features of the [Syntax-highlighting Code Block plugin](https://github.com/westonruter/syntax-highlighting-code-block):

- Highlighted lines
- Numbered lines
- Line wrapping

In order to test these features, I had to refactor the way the stories output markup. I moved those to a separate file and used JavaScript instead of Twig.

## Screenshots

With all three new options enabled:

Screenshot 2023-02-17 at 1 07 22 PM

## Testing

Review the stories in [the SHCB section of the deploy preview](https://deploy-preview-2138--cloudfour-patterns.netlify.app/?path=/docs/vendor-syntax-highlighting-code-block--basic):

- [x] For the "basic" demo, try toggling the line wrap option on and off.
- [x] For the "more features" demo, try toggling all the different options on and off.
- [x] For both demos, add the class `wp-block-code` (via Storybook controls) and confirm that the previously tested features still work fine.

Confirm that the features work as intended in:

- [x] Chrome
- [x] Edge
- [x] Firefox
- [x] Safari (iOS)
- [x] Safari (macOS)


spaceninja:
Screenshot 2023-02-16 at 4 24 28 PM

JSX, TSX, and YAML are all malformated in Chrome, with some of the lines (and line numbers) all appearing on the same line.

If this is just a bug in the Storybook demo, then it might not be a release blocker?

tylersticka:
@spaceninja Great catch.

I _think_ this has to do with how I'm faking the line separations. But just to be safe, I'm going to try doing some local testing in WordPress-land before merging anything.

spaceninja (dismissed):
Whew, that demo.ts file is _dense_.

The only question I'll ask is how much value it's providing to have this complex a set of code just to generate a Storybook demo page for our syntax highlighter.

Put yourself in the head of someone who has to come make a change to this code several years from now and doesn't have all the context you have now.

Would we be better served to just manually copy in a few markup examples from the blog?

I don't feel super strongly about this, and I'll defer to your judgement, but I think the question deserves asking.

Code looks good (if complex), and works as described (barring the bug I mentioned).

I'll approve on the assumption that you think this code is valuable and the bug doesn't matter, but feel free to re-tag me if you make any changes.

tylersticka:
> Would we be better served to just manually copy in a few markup examples from the blog?

A fair question.

tylersticka:
The more I think about this, the more I think Scott's instinct is right.

My goal was to make a Storybook demo that was comprehensive, so modifying Highlight's output to match that of the SHCB plugin made sense in theory. But in practice we now have a situation where there's a bug in the output, and it's unclear whether or not that's a bug of the demo or a bug of the styles. Which is a flaw in the approach.

Copying some sample output from SHCB directly may not be as comprehensive in terms of _languages_, but it will increase my confidence in the styles we've written, _and_ it should take a lot less code.

I'm going to move this back to a draft until I have a chance to address that feedback.

tylersticka:
Based on the previous feedback, I've made the following changes:

- I still preferred having the Highlight demo stuff in a separate file since it's rather JavaScript-y, but I dropped all of the SHCB-specific demos (which were the bulk of the complexity _and_ weren't working 100% well) from that file.
- I separated all of the SHCB stuff from the Highlight stuff just to separate concerns a bit and make the documentation a little more approachable.
- After confirming that the line wrapping issue Scott had pointed out was specific to my old demo implementation, I grabbed some precompiled examples to demonstrate SHCB. I only added options when doing so involved toggling a class rather than modifying markup.

I like this a lot better.

spaceninja (approved):
Looks good, and much easier to understand and maintain. Thank you!

PR #2140, opened 2023-02-21
Add primary button modifier

tylersticka:
## Overview

This adds a new `c-button--primary` modifier class that lends additional visual emphasis to a button. This was done with a future promotional content element in mind, where the existing buttons appears to blend in to adjacent text a bit too much.

My original instinct was to re-style the default button instead, but in practice I found it difficult to find a new "default" button style with enough contrast in our dark theme that did not blend in with Input Group. I decided a new modifier would be a better option.

~While interacting with Input Group, I rediscovered the issue where our button scale effects would conflict with adjacent elements: Hovering a button in an input group feels off, because the button loses its alignment when it scales up. This also came up in the past when discussing the possibility of adding tooltip patterns. I decided the filter effects were probably enough, but I adjusted the "secondary" button styles so that they would feel more visible there.~ (**Edit:** Paul made some really interesting alternative suggestions that made me re-think just straight-up _removing_ this effect. I've set a reminder to create an issue preserving some of those ideas, and for now I'm focusing this PR on the new modifier only.)

## Screenshots

Screenshot 2023-02-21 at 11 41 05 AM

## Testing

[Try out deploy preview](https://deploy-preview-2140--cloudfour-patterns.netlify.app/?path=/docs/components-button--styles)

### DIFF
--- a/src/components/button/button.stories.mdx
+++ b/src/components/button/button.stories.mdx
@@ -81,7 +81,7 @@ The `c-button` class may be applied to button and link elements. The button appe
 
 ## Styles
 
-Some buttons may not require as much emphasis as others. The `c-button--secondary` and `c-button--tertiary` modifiers reflect lesser importance with reduced contrast.
+The `c-button--primary` modifier may be used when a single call to action requires unique emphasis. The `c-button--secondary` and `c-button--tertiary` modifiers reflect lesser importance with reduced contrast.
 
 <Canvas>
   <Story
--- a/src/mixins/_button.scss
+++ b/src/mixins/_button.scss
@@ -2,7 +2,6 @@
 @use '../compiled/tokens/scss/color';
 @use '../compiled/tokens/scss/ease';
 @use '../compiled/tokens/scss/font-weight';
-@use '../compiled/tokens/scss/scale';
 @use '../compiled/tokens/scss/size';
 @use '../compiled/tokens/scss/transition';
 @use './focus';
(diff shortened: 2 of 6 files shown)


src/components/button/button.stories.mdx
```diff
@@ -81,7 +81,7 @@ The `c-button` class may be applied to button and link elements. The button appe
 
 ## Styles
 
-Some buttons may not require as much emphasis as others. The `c-button--secondary` and `c-button--tertiary` modifiers reflect lesser importance with reduced contrast.
+The `c-button--primary` modifier may be used when a single call to action requires unique emphasis. The `c-button--secondary` and `c-button--tertiary` modifiers reflect lesser importance with reduced contrast.
```
tylersticka, line 84: The first sentence felt unnecessary.

src/mixins/_button.scss
```diff
@@ -16,18 +15,26 @@
...
+  --theme-color-background-button-secondary: var(--theme-color-background-base);
   --theme-color-border-button-default: #{color.$brand-primary-darker};
-  --theme-color-border-button-secondary: #{color.$brand-primary-dark};
+  --theme-color-border-button-primary: #{color.$brand-primary-darker};
+  --theme-color-border-button-secondary: #{color.$brand-primary-darker};
+  --theme-color-text-button-default: #{color.$text-light-emphasis};
+  --theme-color-text-button-primary: #{color.$brand-primary};
   --theme-color-text-button-secondary: var(--theme-color-text-action);
   --theme-color-text-button-tertiary: var(--theme-color-text-action);
 }
```
tylersticka, line 40: I think there's a future opportunity to simplify this a bit, but for now I didn't want to rock the boat and make this a much larger PR.

src/mixins/_button.scss
```diff
@@ -57,6 +64,7 @@
   border-style: solid;
   border-width: size.$edge-control;
   color: color.$text-light-emphasis;
+  color: var(--theme-color-text-button-default);
```
tylersticka, line 67: I don't know if it's necessary anymore to have these non-custom-property fallbacks present, but I didn't feel confident enough in that to strip this out or dramatically change the way we are applying styles.

src/mixins/_button.scss
```diff
@@ -68,8 +76,7 @@
   position: relative;
   text-align: center;
   text-decoration: none;
-  transition: filter transition.$slow ease.$out,
-    transform transition.$quick ease.$out;
+  transition: filter transition.$slow ease.$out;
```
tylersticka, line 79: I liked the scale effects but I knew in my heart they were a pain point, and in many browsers the scaling never looks _exactly_ right anyway. Time to "kill my darlings," as they say.

tylersticka:
Specifically requesting @Paul-Hebert's design and dev feedback.

src/mixins/_button.scss
Paul-Hebert: For sure — I think you're right that we could remove most/all of these fallbacks but I agree that can happen later

Paul-Hebert (approved):
Code looks good, and I like the purple :+1:

I understand the logic behind removing the scaling, though it makes me a bit sad as well.

That said, the current filters alone feel a _touch_ bland to me. I'd love if we could inject a _tiny_ bit more fun and make people really wanna click 'em. 

Some potential ideas that may or may not be worth exploring:

- Cranking up the filter a bit. (maybe `brightness(130) saturate(110)`?) Is that too much? We'd need to test text contrast, too... 🤔 
- Could we scale the button's _content_ without scaling the actual button? This might allow us to keep a similar energy to the current buttons, while side-stepping the issues with adjacent content
- Giving the "white" background buttons (secondary and dark-mode primary) a subtle background color when hovered
- Using different BG shades for all hovered buttons
- Could we give buttons a gradient background when hovered?
- Other fun hover effects? Border animations? I dunno...

I don't want this to be a blocker, and I think these styles are fine, so if all those ideas sound terrible, feel free to proceed. But if we can find a way to inject a bit more personality, I think it could be a nice touch.

tylersticka:
@Paul-Hebert I like some of your ideas.

Looking at this now, I think I might just update this PR to _only_ include the new button style and revisit the transform effects another time.

Paul-Hebert (approved):
:shipit:

spaceninja (approved):

PR #2143, opened 2023-02-24
Add cloudy card modifier

tylersticka:
## Overview

Adds a `c-card--cloudy` modifier for very specific promotions.

Because this is a much more specific modifier, I declined to add it to the card controls and instead gave it its own separate demo.

Container edge artifacts drove me nuts.

## Screenshots

Screenshot 2023-02-23 at 4 40 55 PM

## Testing

https://deploy-preview-2143--cloudfour-patterns.netlify.app/?path=/docs/components-card--cloudy


src/components/card/demo/cloudy.twig
```diff
@@ -0,0 +1,26 @@
+{% embed '@cloudfour/components/card/card.twig' with {
+  tag_name: 'aside',
+  class: 'c-card--contained c-card--cloudy t-dark'
+} only %}
+  {% block content %}
+    <div class="o-rhythm o-rhythm--compact has-big-font-size">
```
tylersticka, line 6: Look, I know `has-big-font-size` is a WordPress utility class, I know I probably _should_ add an equivalent for our project. But this works _now_, and perfect is the enemy of good... right? Right???

src/components/card/card.scss
```diff
@@ -39,6 +40,20 @@ $_cover-gap: ms.step(1, 1rem);
 
 $_focus-overflow: (size.$edge-large * -1);
 
+/// The cloudy modifier adds clouds to the bottom right corner and also applies
+/// some fluid padding to accommodate. The clouds feel best when they are sized
+/// just a _tad_ larger than this fluid padding.
+///
+/// @type Number
+/// @access private
```
spaceninja, line 48: What are these two lines for?

src/components/card/card.scss
```diff
@@ -39,6 +40,20 @@ $_cover-gap: ms.step(1, 1rem);
...
+
+$_cloud-height: fluid.fluid-clamp(
```
spaceninja, line 50: Could we rename this variable to better reflect that it's being used for the `background-size` property, rather than a height, which I would expect to contain a single number? Perhaps something like `$_cloud-bg-size`?

```suggestion
$_cloud-bg-size: fluid.fluid-clamp(
```

src/components/card/card.scss
tylersticka: Part of the Sassdoc comment, specifying what the variable is for, what it's expected to be and whether or not it's intended for export via `@use`. This is consistent with other variables in this file.

At one time I had experimented with importing Sassdoc comments into our Storybook docs, which I still think is a neat idea, but a challenging one to prioritize. Now I'm just maintaining it for consistency here.

src/components/card/card.scss
spaceninja: Actually, now i'm super confused because I see you're using the same variable for `block-size`.

I don't understand how we can pass a comma-separated variable into `block-size`?

It's rendering fine in the browser, but that doesn't feel right to me, unless I'm missing something clever here.

spaceninja: Oh FFS.

I'm sorry, my bad — I misread this and didn't realize the output of this _IS_ a single value, the numbers you're setting are not a comma-separated list, they're the parameters for the clamp function. Sorry!

spaceninja (approved):

src/components/card/card.scss
tylersticka: > I don't understand how we can pass a comma-separated variable into block-size?

I don't think we are?

We're setting the variable to this:

```
fluid.fluid-clamp($min, $max, $min-width, $max-width)
```

Which is outputting this:

```
clamp(min, val, max)
```

Both the Sass function and `clamp` functions accept multiple arguments, but `clamp` represents a single value. Inspecting the preview, I only see one `clamp` in the `block-size`.

PR #2144, opened 2023-02-27
Add eyebrow block to card component

tylersticka:
## Overview

Adds a new optional block to include an "eyebrow" that visually precedes a heading in a card component, similar to the Article Header layout object. This will be used to include client logos in our case study listing (independent of the headlines, which are no longer simply the client name).

I moved our example logos to a more generic directory name so I could use an open source project as the example logo.

## Screenshots

Screenshot 2023-02-27 at 2 40 46 PM

## Testing

[Deploy Preview](https://deploy-preview-2144--cloudfour-patterns.netlify.app/?path=/docs/components-card--eyebrow)


src/components/card/card.stories.mdx
```diff
@@ -137,6 +138,22 @@ An optional `cover` block may be provided containing an image or other visual ob
   </Story>
 </Canvas>
 
+## With eyebrow
+
+If a card includes a heading, you may want to include additional elements that precede the heading visually without disrupting the element's outline. For example, a customer's logo in a card that links to a case study. Similar to [the Article Header object](/?path=/docs/objects-article-header--example), the contents of the `eyebrow` block will display above the heading.
```
spaceninja, line 143:
```suggestion
If a card includes a heading, you may want to include additional elements that precede the heading visually without disrupting the element's outline. For example, a customer's logo in a card that links to a case study. Similar to [the Article Header object](/?path=/docs/objects-article-header--example), the contents of the `eyebrow` block will render below the heading in the DOM, but display above the heading visually.
```

Minor suggestion to make it more clear what's happening

spaceninja (approved):

spaceninja (approved):

PR #2154, opened 2023-03-16
Fix form-related Jetpack style bugs

tylersticka:
## Overview

When updating our theme's patterns and applying them to Jetpack form blocks, I uncovered a few bugs:

- For some reason in Firefox and _only_ in the theme (not in Storybook), `select.c-input` (or comparable pattern in Jetpack forms) have their caret icon vertically stretched. This is because of a Chrome hack that I can no longer recreate the original bug for.
- We were missing a math operator in a `padding-inline-end` style for `select.c-input`, so it wasn't having any effect before.
- If an optional form field is submitted with no value, its `div.field-value` would be empty, which broke the layout. I added styles to not break the layout and to hide the name and value entirely if `:has` is supported.

## Testing

On the deploy preview, review the following components in Chrome and Firefox:
- [x] [`select.c-input`](https://deploy-preview-2154--cloudfour-patterns.netlify.app/?path=/story/components-input--select-element) should have the correct amount of right/inline-end padding
- [x] [`select.c-input`](https://deploy-preview-2154--cloudfour-patterns.netlify.app/?path=/story/components-input--select-element) should have a caret icon that is correctly sized and positioned
- [x] [Jetpack form dropdowns](https://deploy-preview-2154--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-jetpack-blocks--contact-form) should have the correct amount of right/inline-end padding
- [x] [Jetpack form dropdowns](https://deploy-preview-2154--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-jetpack-blocks--contact-form) should have a caret icon that is correctly sized and positioned
- [x] In Firefox, [the Jetpack success message](https://deploy-preview-2154--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-jetpack-blocks--contact-form-success) values that are empty should not break the layout
- [x] In Chrome, [the Jetpack success message](https://deploy-preview-2154--cloudfour-patterns.netlify.app/?path=/story/vendor-wordpress-jetpack-blocks--contact-form-success) names and values that are empty should be hidden


src/vendor/wordpress/styles/_jetpack-blocks.scss
```diff
@@ -154,19 +154,37 @@
     clear: left; // 1
     font-weight: font-weight.$medium; // 2
     margin-inline-end: 1ch; // 3
+
+    /// If there is no field value, don't bother displaying the field name.
+    &:has(+ .field-value:empty) {
+      display: none;
+    }
```
Paul-Hebert, line 161: Clever!

src/vendor/wordpress/styles/_jetpack-blocks.scss
```diff
@@ -154,19 +154,37 @@
...
+    /// Empty field values need to clear floats to maintain margin.
+    &:empty {
+      clear: left;
+      float: none;
+
+      /// If the browser supports `:has()`, we can hide the field value entirely
+      /// since the field name will be hidden as well.
+      @supports selector(:has(+ :empty)) {
+        display: none;
+      }
```
Paul-Hebert, line 178: I'm a little confused about why we don't always hide this? I'm assuming we want to keep this in sync with its label, but it seems like if it's empty we could just always hide it.

I'm guessing there's some layout reason related to how our floats work, so feel free to ignore. I just figured I'd mention it in case it sparked an idea.

I also kind of wonder if doing something like this would be good as a fallback?

```css
@support not selector(:has(+ empty)) {
  &::before {
    content: 'No Value'
  }
}
```

I'm not sure... just more food for thought!

Paul-Hebert: Oh, I'm guessing we still need to display the element so that its bottom margin applies? That makes sense.

Paul-Hebert (approved):
Looks good!

src/vendor/wordpress/styles/_jetpack-blocks.scss
tylersticka: Exactly, we need to keep its bottom margin around.

I thought about the pseudo element idea, but in the past I've received feedback that CSS `content` doesn't work with automated translation tools. I figured I'd keep this presentational.

Issue #2148, opened 2023-03-06
I would love a "Side Note" block


Paul-Hebert:
It can be nice to have an easy way to add asides or related diagrams to articles. For example here's a screen shot of side notes I've used in the past when [writing about technical topics outside of the Cloud Four blog](http://paulhebertdesigns.com/web_colors/):

<img width="1059" alt="image" src="https://user-images.githubusercontent.com/5798536/223187211-3967b127-4838-4089-a0fb-9710852752c0.png">

(I saw [Tufte CSS](https://edwardtufte.github.io/tufte-css/) over the weekend which also got me thinking about this.)

I wonder if we could make a block to support this use case? My assumption would be that on small screens it would switch to displaying inline in the text?

This might be simpler to build than footnote support while allowing a similar functionality.

PR #2158, opened 2023-04-06
Potential side note / aside design

tylersticka:
## Overview

A little prototype with some controls to experiment with a potential floating sidebar design, which @Paul-Hebert suggested in #2148.

This isn't intended to be merged, just to share a deploy preview for reference and discussion.

## Screenshots

On small screens, the side does not look much different from a paragraph with a silver background:

Screenshot 2023-04-06 at 9 58 32 AM

I set its `font-size` to be one step down in our modular scale, but it felt _really_ small to me, so I used a `max` function to only do so up to a minimum of `16px`. I'm open to feedback on this.

I wondered if it might feel less intrusive sometimes to use `<details>` and `<summary>`, so I added a control for specifying a summary:

Screenshot 2023-04-06 at 9 58 43 AM

Is that too boring looking? I'm not sure.

At very wide sizes, I use some absolute positioning and a lot of `calc` to nestle the side note into the margins. I also use a radial gradient to fake an inverted rounded corner, because I thought it felt a bit disconnected from its point in the article. (I mean, it's _still_ a lot more disconnected from the original footnote, but this way I think it's clear that it comes between two paragraphs.)

Screenshot 2023-04-06 at 9 59 05 AM

Like @Paul-Hebert's site, I added a modifier to allow the aside to align the opposite way:

Screenshot 2023-04-06 at 9 59 15 AM

The `<details>` version would work the same way at large sizes:

Screenshot 2023-04-06 at 9 59 37 AM

Screenshot 2023-04-06 at 9 59 48 AM

Gosh, that version is missing something. Does it need a rule between the summary and details? Should the summary be a medium weight? I'm not sure.

I'm also wondering now if the container is too confining on large screens. Maybe a border to the left or right would work? Might be worth trying and comparing.

## Testing

Try it out for yourself. I've not tested it outside of Arc (Chromium).

### DIFF
--- a/src/prototypes/asides/article.twig
+++ b/src/prototypes/asides/article.twig
@@ -1,0 +1,34 @@
+{% embed '@cloudfour/objects/container/container.twig' with {
+  class: '_proto-asides-r1 o-container--prose o-container--pad',
+  content_class: 'o-rhythm',
+  align: align,
+  summary: summary,
+  open: open
+} only %}
+  {% block content %}
+    <p>In the short run for those using a traditional web design process, we need to recognize that design doesn’t stop when something is handed off to developers. It continues until launch so designers can be involved in how the design is implemented in code.</p>
+
+    {% set aside_content -%}
+      <p>Or you could replace the traditional web design process with a process that incorporates in-browser mockups, but not every organization and designer can make that change. But if you’re interested in a process like that, stay tuned or… um… <a href="mailto:info@cloudfour.com">hire us</a>.</p>
+    {%- endset %}
+
+    <aside class="_c-aside {% if align %}_c-aside--align-{{align}}{% endif %}">
+      <div class="_c-aside__inner">
+        {% if summary %}
+          <details class="_c-aside__content"{% if open %} open{% endif %}>
+            <summary>{{summary}}</summary>
+            {{aside_content}}
+          </details>
+        {% else %}
+          <div class="_c-aside__content">
+            {{aside_content}}
+          </div>
+        {% endif %}
+      </div>
+    </aside>
+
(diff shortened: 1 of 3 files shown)


Paul-Hebert:
Yay, this is great! I didn't review the code but in terms of functionality I think this is great (and honestly would be fine adding this as a feature as-is while we fine-tune it.)

> Is that too boring looking? I'm not sure.

I do wonder if there's an opportunity to add some character to the `<summary>` while still having its meaning clear. Could something small like using a different arrow/icon or giving the arrow a color help?

> Gosh, that version is missing something. Does it need a rule between the summary and details? Should the summary be a medium weight? I'm not sure.

I wondered the same things. I think both options are worth exploring.

> I'm also wondering now if the container is too confining on large screens. Maybe a border to the left or right would work? Might be worth trying and comparing.

I wondered about this as well. I think the container is fun but I'm not sure I'd always want it. Maybe it should be an optional style on large screens? Or maybe we do a different treatment all the time like you mentioned.

---

Other random thoughts:

1. Should `summary` have `cursor: pointer`? I know some folks are pretty passionate that only links should get that style but in practice it seems like most users expect all/most interactive content to get that cursor?
2. I assume this can include non-paragraph content? e.g. could I put a list of links? Or a testimonial quote with an avatar?
3. On large screens, the current design makes it clear the side not sits _between_ paragraphs, but I wonder if that matches the intent of our usual use case. When I've used side notes in the past they've contained auxiliary info _about a specific paragraph_. I wonder if it should instead sit to the side of the paragraph above? (Obviously we can't do this on small screens. I'm also not sure how feasible it is on large screens with the way the DOM flows but I figured I'd mention it.)

Thanks for working on this Tyler!

PR #2167, opened 2023-05-15
Add basic, opinionated text size utilities

tylersticka:
## Overview

I've found myself using the WordPress `has-big-font-size` class in prototypes, so it seemed overdue to represent that in our own system (as `u-text-big`). In a recent marketing page prototype, I also found need for a `u-text-small` class that caps the size reduction to `16px`, so I took this chance to introduce both.

I chose to keep the "big" and "small" terms for these, because I think in this case their vagueness actually matches the responsive behavior. If we ever want to introduce utilities that were tied to steps of our modular scale, they shouldn't conflict with these.

## Screenshots

Screenshot 2023-05-15 at 4 01 13 PM

text-big-small

## Testing

[Try out the deploy preview](https://deploy-preview-2167--cloudfour-patterns.netlify.app/?path=/docs/utilities-text--big-and-small) in the following browsers:
- [x] Chrome
- [x] Edge
- [x] Firefox
- [x] Safari

### DIFF
--- a/.changeset/pink-foxes-guess.md
+++ b/.changeset/pink-foxes-guess.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Prevent the WordPress `has-small-font-size` utility class from reducing text below 16 pixels for consistency with `u-text-small`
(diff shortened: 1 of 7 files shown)


Paul-Hebert (approved):
LGTM!

PR #2168, opened 2023-05-16
Logo Group padding and justification options

tylersticka:
## Overview

The Logo Group was designed for the presentation of client logos on our work page, but while prototyping marketing content I encountered a use case for displaying a couple of logos left-aligned within a normal content flow.

Rather than introduce modifier classes that _remove_ functionality, I thought it was more appropriate to start from a less opinionated baseline and to tack on the features. That means that this is a breaking change that will require a small alteration to the template properties of the work page in our WordPress theme.

## Screenshots

Screenshot 2023-05-16 at 09-57-22 objects-logo-group--default-story

## Testing

[Try out the deploy preview](https://deploy-preview-2168--cloudfour-patterns.netlify.app/?path=/docs/objects-logo-group--default-story) in the following browsers:
- [x] Chrome
- [x] Edge
- [x] Firefox
- [x] Safari

### DIFF
--- a/src/objects/logo-group/logo-group.scss
+++ b/src/objects/logo-group/logo-group.scss
@@ -1,21 +1,41 @@
 @use '../../compiled/tokens/scss/size';
 @use '../../mixins/ms';
 
+/// The Logo Group container is where most of the magic happens. We use flex
+/// instead of grid because we want to support centered justification later (if
+/// we used grid, the bottom row would always be left/start justified).
 .o-logo-group {
   align-items: center;
   display: flex;
   flex-wrap: wrap;
+  gap: size.$spacing-gap-logo-group;
+}
+
+/// The initial use case for this pattern called for white space to surround the
+/// logos, but this makes the pattern more difficult to work with when left-
+/// justified or used in a more condensed layout, so we've moved this to a
+/// modifier class instead.
+.o-logo-group--pad {
+  padding: size.$spacing-gap-logo-group;
+}
+
+/// These modifiers set the horizontal alignment of the logos. Historically we
+/// would have used the term "align" for this, but because Flex and Grid both
+/// use "align" to refer to vertical alignment (at least in Western languages
+/// and writing modes), we're continuing the "justify" usage set in the Logo
+/// component.
+
+.o-logo-group--justify-center {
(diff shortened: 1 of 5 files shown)


src/objects/logo-group/logo-group.stories.mdx
```diff
@@ -1,29 +1,63 @@
...
+  const twigArgs =
+    Object.keys(args).length > 0
+      ? ` with ${JSON.stringify(args, null, 2)}`
+      : '';
+  return `{% embed '@cloudfour/objects/logo-group/logo-group.twig'${twigArgs} %}
+  {% block content %}
+    {# logos #}
+  {% endblock %}
+{% endembed %}`;
+};
```
tylersticka, line 16: I found the raw demo source was _mostly_ noise, so I made a custom template source function to be more helpful.

When we eventually update the project to the latest Storybook and Vite, I honestly think we should consider starting over when it comes to the source previews since this sort of thing is so widespread. But for now, it's nice to have a useful code sample.

src/objects/logo-group/logo-group.scss
```diff
@@ -1,21 +1,41 @@
...
 @use '../../mixins/ms';
 
+/// The Logo Group container is where most of the magic happens. We use flex
+/// instead of grid because we want to support centered justification later (if
+/// we used grid, the bottom row would always be left/start justified).
 .o-logo-group {
   align-items: center;
   display: flex;
   flex-wrap: wrap;
+  gap: size.$spacing-gap-logo-group;
```
tylersticka, line 11: This actually results in slightly less whitespace than the previous technique since there aren't margins to double up, but my eye prefers having consistent whitespace between _and_ around the items (compared to before, when there was half as much around) and I'm not bothered by the space reduction. But, I'm open to suggestion on this!

src/objects/logo-group/logo-group.stories.mdx
Paul-Hebert: I feel like this always ends up being a hard challenge when writing docs. Hopefully we can find a better pattern after the upgrade :+1:

Paul-Hebert (approved):
Makes sense — looks good!

PR #2171, opened 2023-05-25
Fixes to Media Link and Hype Group dimension issues

tylersticka:
## Overview

While implementing some of the newer patterns in our WordPress theme, I came across two bugs:

- A Media Link with content that exceeds its container would break out of a Grid or Flex container.
- A Hype Group image with dimensions smaller than the container would appear awkwardly cropped within.

## Screenshots

Before | After
--- | ---
Screenshot 2023-05-25 at 1 28 56 PM | Screenshot 2023-05-25 at 1 29 09 PM
Screenshot 2023-05-25 at 1 40 09 PM | Screenshot 2023-05-25 at 1 40 26 PM

## Testing

In [this story on the deploy preview](https://deploy-preview-2171--cloudfour-patterns.netlify.app/?path=/story/objects-hype-group--multiple):

- [x] Confirm that the Media Link never breaks out of its container or pushes the object too far to the right.
- [x] Lower the value of the `example_object_img_size` control and confirm that the image never appears "cropped" or like it stops awkwardly.

### DIFF
--- a/.changeset/cyan-terms-sip.md
+++ b/.changeset/cyan-terms-sip.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Prevent Media Links from breaking out of Grid or Flex containers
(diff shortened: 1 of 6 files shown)


Paul-Hebert (approved):
🚢

PR #2176, opened 2023-06-23
Revise blockquote-derived styles

### LINKED ISSUE #2161: Quote citation styles intended for WordPress blocks are leaking out to all blockquotes
## Overview

I noticed that some of our blockquotes had an em dash that could use some improved styles to account for line breaks. I was confused when I didn't see those styles in our defaults.

It turns out, it's a style that seems intended to target WordPress blocks _other_ than pullquotes that include a blockquote and citation:

https://github.com/cloudfour/cloudfour.com-patterns/blob/bbb148e480de16a96d930ccc930d5b3c509e596d/src/vendor/wordpress/styles/_core-blocks.scss#L412-L432

The issue is `:not(.wp-block-pullquote)`, which will select… well, just about anything.

tylersticka:
## Overview

Our revised marketing content contains a lot more testimonials than before. This exposed some issues with our current `blockquote` styles as well as dependencies, specifically:

- The default `blockquote` styles were probably _too_ lean: We were having to repeat the styles in comment content, markdown/classic blocks and the WordPress quote and pullquote blocks.
- In an attempt to preserve those lean, default styles, `:not` selectors were used to fence the WordPress block styles, but these were used too broadly, resulting in the opposite effect: The WordPress block styles were overriding _all_ `blockquote` elements as a result.
- The WordPress block styles were out of date, making them difficult to use.
- Citations did not break to a new line gracefully. 

This PR attempts to address that by:

- Expanding the default `blockquote` styles.
- Removing the comment, markdown and classic styles in favor of deferring to those defaults.
- Revising the WordPress quote styles to focus on the block features without leakage.

I've marked this as a "major" release because I know of several marketing pages with inline styles that will need to be simplified after this change takes effect.

### Default blockquote

Screenshot 2023-06-23 at 11 46 51 AM

### WordPress quote block

Screenshot 2023-06-23 at 11 47 39 AM

The "plain" style omits the usual appearance, presumably for situations where a blockquote is the right semantic choice but shouldn't be differentiated from surrounding content:

Screenshot 2023-06-23 at 11 47 48 AM

Text alignment options are supported:

Screenshot 2023-06-23 at 11 47 58 AM

If the blockquote is given a background color, it will now behave similarly to paragraphs and groups instead of requiring inline styles to omit borders or adjust padding:

Screenshot 2023-06-23 at 11 49 36 AM

### WordPress pullquote block

Screenshot 2023-06-23 at 11 50 11 AM

Supports text alignment options:

Screenshot 2023-06-23 at 11 50 33 AM

As well as general alignment options:

Screenshot 2023-06-23 at 11 51 17 AM

## Testing

TODO

---

- Fixes #2161

### DIFF
--- a/.changeset/tame-geese-provide.md
+++ b/.changeset/tame-geese-provide.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': major
+---
+
+Revised `blockquote` and WordPress Quote and Pullquote block styles to address issues of consistency and improve feature support. Content that relied on previous quote styles may need to be reviewed or updated.
--- a/src/base/_defaults.scss
+++ b/src/base/_defaults.scss
@@ -2,7 +2,9 @@
 @use '../compiled/tokens/scss/color';
 @use '../compiled/tokens/scss/size';
 @use '../mixins/border-radius';
+@use '../mixins/emdash';
 @use '../mixins/ms';
+@use '../mixins/spacing';
 @use '../mixins/theme';
 
 /**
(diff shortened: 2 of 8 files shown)


Paul-Hebert (approved):
Looks good! I like the pull quote alignment options :)

PR #2190, opened 2023-08-21
Mailchimp spacing fix

tylersticka:
## Overview

When using a Mailchimp form within an article, the heading appeared much closer to the preceding content than intended. This was because the vertical rhythm was not being applied to the first child, which was usually a heading that exceeded `1em` in size.

This applies the initial vertical padding to the first element as long as the Mailchimp form is not the `:first-child` of its container.

## Screenshots

Before | After
--- | ---
Screenshot 2023-08-21 at 4 09 44 PM | Screenshot 2023-08-21 at 4 09 55 PM

## Testing

- [x] Open [the Mailchimp Sign-up form story on the deploy preview](https://deploy-preview-2190--cloudfour-patterns.netlify.app/?path=/story/vendor-mailchimp--signup-form) in "Canvas" view.
- [x] Confirm that the Mailchimp form and the heading do not have top margin.
- [x] In the Storybook controls, toggle `show_example_content` to `true`. Confirm that the spacing that precedes the "Example Heading" is the same as the form heading.
- [x] In the Storybook controls, toggle `transform_heading` to `true`. Confirm that the spacing is the same as in the previous testing step.

### DIFF
--- a/.changeset/spicy-badgers-wait.md
+++ b/.changeset/spicy-badgers-wait.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Fixes Mailchimp forms appearing inconsistently spaced with adjacent content.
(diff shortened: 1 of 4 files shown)


Paul-Hebert (approved):
:shipit:

PR #2192, opened 2023-08-22
[NO MERGE] New Footer Prototypes

tylersticka:
## Overview

The intent is for this to replace the existing Ground Nav. It's okay for this to be a breaking change that foregoes the "PWA Stats" use case and changes the data requirements.

I have three revisions of prototypes that are all relevant for different reasons. Below I've highlighted those details, and included screenshots where I mark the key points of reference in each.

---

### [New Footer R1](https://deploy-preview-2192--cloudfour-patterns.netlify.app/?path=/story/prototypes-new-footer--r-1)

Screenshot 2023-08-22 at 9 46 37 AM

- Only prototype with actual "lead-in" promotional content.

---

### [New Footer R2](https://deploy-preview-2192--cloudfour-patterns.netlify.app/?path=/story/prototypes-new-footer--r-2)

Feature Count | Behavior | Screenshot
:---: | --- | ---
0 | No space is reserved for features | Screenshot 2023-08-22 at 9 49 29 AM
1 | The single feature has a "prose" maximum width | Screenshot 2023-08-22 at 9 49 16 AM
2 | The two features stack on small screens, on large screens they fill the default maximum width | Screenshot 2023-08-22 at 9 49 02 AM

- The "lead-in" content adapts to there being zero, one or two items therein.

---

### [New Footer R3](https://deploy-preview-2192--cloudfour-patterns.netlify.app/?path=/story/prototypes-new-footer--r-3)

Screenshot 2023-08-22 at 9 55 06 AM

- The main footer content (below the "lead-in") is much more worked out here. Also, the overall element adapts to themes better.
- The background uses blend modes so the illustration works well against any color.
- I'm really happy with the grid layout in this one.
- If we plan to start with the HTML/CSS of one of these prototypes, this is the one I'd recommend starting from.

---

### Unresolved Issues
- I would like the footer to work well even if the preceding element is silver (so its color would be white in those cases).
- The conditional border radius for the "lead-in" elements doesn't seem to work super reliably here. I've been thinking it might make sense to replace that mixin's technique with a solution that uses container queries: If need be, we can use this as a testing ground for that idea?

### DIFF
--- a/src/assets/illustrations/portland.svg
+++ b/src/assets/illustrations/portland.svg
@@ -1,67 +1,67 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="450" height="200" viewBox="0 0 450 200">
+<svg xmlns="http://www.w3.org/2000/svg" width="450" height="200" viewBox="0 0 450 200" preserveAspectRatio="xMaxYMax">
   <g>
-    <rect x="4" y="60" width="125" height="140" fill="#dae2e8"/>
-    <rect x="4" y="170" width="125" height="30" fill="#f2f5f7"/>
-    <rect x="0" y="60" width="129" height="15" fill="#dae2e8"/>
-    <rect x="4" y="75" width="125" height="3" fill="#c0cfd8"/>
+    <rect x="4" y="60" width="125" height="140" fill="#dae2e8" />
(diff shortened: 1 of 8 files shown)


tylersticka:
Closing since this feature has been implemented! (Thanks, @spaceninja! 🎉)

PR #2193, opened 2023-08-22
Fix some code snippets being too large in iOS

tylersticka:
## Overview

I was reading [one of our recent articles](https://cloudfour.com/thinks/progressively-enhanced-form-validation-part-3-validating-a-checkbox-group/) and noticed that some code snippets had very huge text.

This seems to be resolved by applying our `text-size-adjust` rule to `-webkit-text-size-adjust` as well. This matches [what caniuse says](https://caniuse.com/text-size-adjust) about support for this feature.

## Screenshots

Before | After
--- | ---
IMG_6489 | IMG_6490

## Testing

Unfortunately it's not super clear to me how to reproduce this outside of the context of a blog post, but at least review the code to verify that this seems like a not-very-risky change?

### DIFF
--- a/.changeset/perfect-needles-kick.md
+++ b/.changeset/perfect-needles-kick.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Fixes issue where some code snippets had very large text in iOS Safari.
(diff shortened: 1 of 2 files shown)


gerardo-rodriguez (approved):
Thank you for this fix. 🙏🏽  

I forgot to make a GH issue when [I first noticed it](https://cloudfour.slack.com/archives/C04AEEVK2/p1688147992205409). 😬

PR #2200, opened 2023-09-01
Radio component

### LINKED ISSUE #2157: Radio component
We have [a Checkbox component](https://cloudfour-patterns.netlify.app/?path=/docs/components-checkbox--enabled), but we do not have an accompanying Radio component.

This came up when extending our form styles to those supported by Jetpack's form blocks. [In the example story](https://cloudfour-patterns.netlify.app/?path=/docs/vendor-wordpress-jetpack-blocks--contact-form) can see that the radio blocks are unstyled:

Some requirements:

- [ ] Share visual conventions (dimensions, border, animation, etc.) with checkbox
- [ ] Default state
- [ ] `checked` state
- [ ] `disabled` state
- [ ] `checked` _and_ `disabled` state
- [ ] Light/default theme styles
- [ ] Dark theme styles
- [ ] Keep styles in a mixin within the `toggle-input.scss` file (where checkbox styles are)
- [ ] Apply mixin to `c-radio` and to relevant Jetpack form block styles (a la checkboxes)
- [ ] Add Storybook docs for Radio component (a la Checkbox component)

Normally I would label this as a design task, but I think there's enough prior design here to inform an implementation-led approach. Happy to provide suggestions and direction along the way, so please ask for guidance, thoughts, etc.!

tylersticka:
## Overview

Creates a new Radio component for styling `radio` type inputs and applies those same styles to the Jetpack contact form CSS for consistency with checkboxes.

## Screenshots

Before | After
--- | ---
Screenshot 2023-09-01 at 9 35 44 AM | Screenshot 2023-09-01 at 9 35 20 AM

Screenshot 2023-09-01 at 9 34 58 AM

## Testing

Using the deploy preview, test in the following browsers:

- [ ] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Safari (Desktop)
- [ ] Safari (iOS)

---

- Fixes #2157

### DIFF
--- a/src/components/radio/radio.twig
+++ b/src/components/radio/radio.twig
@@ -1,0 +1,23 @@
+{% set attributes %}
+  type="radio"
+  class="c-radio{% if class %} {{ class }}{% endif %}"
+  {% for attribute_name in [
+    'value',
+    'id',
+    'name',
+    'autocomplete'
+  ] %}
+    {% set attribute_value = attribute(_context, attribute_name) %}
+    {% if attribute_value %}{{ attribute_name }}="{{ attribute_value }}"{% endif %}
+  {% endfor %}
+  {% for attribute_name in [
+    'checked',
+    'required',
+    'disabled'
+  ] %}
+    {% set attribute_value = attribute(_context, attribute_name) %}
+    {% if attribute_value %}{{ attribute_name }}{% endif %}
+  {% endfor %}
+{% endset %}
+
+<input {{ attributes }}>
--- a/src/components/radio/radio.twig
+++ b/src/components/radio/radio.twig
@@ -1,3 +1,8 @@
+{#
+This all might be a little overengineered? Leaving in for now for consistency
+with the checkbox component.
+#}
+
 {% set attributes %}
   type="radio"
   class="c-radio{% if class %} {{ class }}{% endif %}"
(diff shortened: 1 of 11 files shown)


src/components/radio/radio.twig
```diff
@@ -0,0 +1,23 @@
...
+    'checked',
+    'required',
+    'disabled'
+  ] %}
+    {% set attribute_value = attribute(_context, attribute_name) %}
+    {% if attribute_value %}{{ attribute_name }}{% endif %}
+  {% endfor %}
+{% endset %}
+
+<input {{ attributes }}>
```
Paul-Hebert, line 23: Part of me wonders if the complexity of this component is worth it.

Would it be better to just tell component consumers to drop a specific class on a radio input? It seems like this component could become a hindrance if we needed to add an unsupported attribute or make changes in the future.

Paul-Hebert: (Not a strong preference/blocker)

Paul-Hebert (approved):
LGTM! One non-blocking nit inline

src/components/radio/radio.twig
tylersticka: I wondered the same thing, but since this was copy/pasted from the checkbox, the thought of downgrading that component for parity made me a bit weary.

Maybe in the future we could combine the two into one simpler component?

tylersticka: @Paul-Hebert I've added a comment to both the checkbox and radio templates to acknowledge this suggestion.

Paul-Hebert: Sounds good!

PR #2202, opened 2023-09-06
Better Ground Nav Options

spaceninja:
## Overview

This PR makes a couple changes to make it easier to use the Ground Nav template in WordPress:

- Renames `features` to `feature_count`
- Adds a `block` for both features so the content can be set

## Testing

Review the Ground Nav page on the preview deploy. It should still show the appropriate number of features in each story.


Paul-Hebert (approved):
LGTM

PR #2203, opened 2023-09-07
Add Block Size Utilities

spaceninja:
## Overview

In a recent WordPress change I had use for a `u-block-size-full` util but it didn't exist. This adds both "full" and "auto" size utils for the block direction.

## Screenshots

Screenshot 2023-09-07 at 8 39 45 AM

## Testing

1. Review the new "Block Size" story in Utilities > Size on the preview deploy.

### DIFF
--- a/src/utilities/size/size.stories.mdx
+++ b/src/utilities/size/size.stories.mdx
@@ -39,8 +39,8 @@ Utilities beginning with `u-size-inline-` specify the `inline-size` of an elemen
 
 Utilities beginning with `u-size-block-` specify the `block-size` of an element:
 
-- `u-size-inline-full` sets the `inline-size` to `100%` of the container.
-- `u-size-inline-auto` sets the `inline-size` to the natural value for the element.
+- `u-size-block-full` sets the `block-size` to `100%` of the container.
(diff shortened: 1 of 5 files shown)


src/utilities/size/size.stories.mdx
```diff
@@ -34,6 +35,17 @@ Utilities beginning with `u-size-inline-` specify the `inline-size` of an elemen
   <Story name="Inline size (Width)">{() => inlineSizeDemo()}</Story>
 </Canvas>
 
+## Block size
+
+Utilities beginning with `u-size-block-` specify the `block-size` of an element:
+
+- `u-size-inline-full` sets the `inline-size` to `100%` of the container.
+- `u-size-inline-auto` sets the `inline-size` to the natural value for the element.
```
Paul-Hebert, line 43:
```suggestion
- `u-size-block-full` sets the `block-size` to `100%` of the container.
- `u-size-block-auto` sets the `block-size` to the natural value for the element.
```

Should these docs be talking about block size, not inline size?

Paul-Hebert (approved):
🚢

PR #2205, opened 2023-09-07
Chore/ground nav promos width

spaceninja:
## Overview

While testing the updated Ground Nav in WordPress, I noticed that if the promo content was too short, and there was only one promo showing, it should shrink and look odd. This PR remedies that by applying a min-width to the promo container.

## Screenshots

Screenshot 2023-09-07 at 11 34 07 AM

## Testing

Review the "One Feature" story on the Ground Nav component on the preview environment. It should look appropriate even though it now has short content.

### DIFF
--- a/src/utilities/size/demo/block-size.twig
+++ b/src/utilities/size/demo/block-size.twig
@@ -4,7 +4,7 @@
       full (in a 150px tall container)
     </div>
   </div>
-  <div style="height:150px;">
+  <div style="height:150px;" class="t-alternate">
     <div class="u-size-block-auto u-pad-n1 u-rounded t-dark t-alternate u-text-center">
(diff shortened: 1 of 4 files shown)


src/utilities/size/demo/block-size.twig
```diff
@@ -4,7 +4,7 @@
       full (in a 150px tall container)
     </div>
   </div>
-  <div style="height:150px;">
+  <div style="height:150px;" class="t-alternate">
```
spaceninja, line 7: Unrelated change that didn't get included in the last PR that makes the block size utilities demo easier to understand.

Paul-Hebert (dismissed):
LGTM!

PR #2208, opened 2023-09-12
Bugfix: Ground Nav Feature Widths

spaceninja:
## Overview

This PR fixes a bug where the width of the features for the Ground Nav wasn't quite right. Now, when there are two features, they'll match the width of the Ground Nav content, and when there's one feature, it'll match the width of a `fill-pad` container in the prose-width content above the Ground Nav.

## Screenshots

**Two Features**

Screenshot 2023-09-12 at 11 24 03 AM

**One Feature**

Screenshot 2023-09-12 at 11 23 38 AM

**Note:** See comment below for updated screenshots.

## Testing

1. On the preview deploy, review the Ground Nav stories (particularly the "Cloud Four" and "One Feature" stories) at multiple viewport sizes in the following browsers:
    - [x] Chrome
    - [x] Edge
    - [x] Firefox
    - [x] Safari
    - [ ] Mobile Safari
2. Then test again by changing the content of the features to very short strings to ensure the features don't collapse when the content isn't full-width.

### DIFF
--- a/src/components/ground-nav/demo/ground-nav-demo.twig
+++ b/src/components/ground-nav/demo/ground-nav-demo.twig
@@ -1,0 +1,77 @@
+<div class="o-container o-container--pad o-container--prose">
+<div class="o-container__content o-rhythm o-rhythm--default o-rhythm--generous-headings">
+
+<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Aliquam ultrices sagittis orci a scelerisque purus semper.</p>
+
+<article class="c-card c-card--contained o-container__fill-pad">
+  <div class="c-card__content">
+    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Neque vitae tempus quam pellentesque nec nam aliquam sem et. Proin gravida hendrerit lectus a.</p>
+  </div>
+</article>
+
+</div>
+</div>
+
+{% embed '@cloudfour/components/ground-nav/ground-nav.twig' %}
+  {% block feature_one %}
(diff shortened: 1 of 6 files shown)


src/components/ground-nav/ground-nav.twig
spaceninja, line 1: Deleted code relocated to `ground-nav-demo.twig`

src/components/ground-nav/demo/ground-nav-demo.twig
spaceninja, line 1: This is a demo file that uses the main Ground Nav template, but adds some fake page content above to verify alignment. The feature blocks below were relocated from `ground-nav.twig` since they're demo code.

tylersticka:
At large sizes, the content within the features should align with the content outside:

Screenshot 2023-09-12 at 11 30 54 AM

This seems to be a partial regression from the version already in `main`.

spaceninja:
Per Tyler's feedback this has been updated with a revised layout to ensure proper alignment.

## Single Feature:

Note the content of the single feature aligns with the page's prose content and the outer edge of the feature aligns with the callout box in the prose area.

Screenshot 2023-09-13 at 1 51 19 PM

## Multiple Features:

Note the content of the multiple features aligns with the footer's content, and the outer edge of the feature aligns with the footer's padding.

Screenshot 2023-09-13 at 1 51 42 PM

tylersticka (commented):
Looks great! Thanks for addressing those little details, they make a big difference.

Deferring merge approval to @cloudfour/dev

Paul-Hebert:
Is it okay that the signup form isn't full width if the other content is shorter?

image

Paul-Hebert (approved):
I left one question about an edge case that doesn't seem like a blocker. Looks good besides that!

spaceninja:
> Is it okay that the signup form isn't full width if the other content is shorter?

Good catch! Fixed.

Paul-Hebert (approved):

PR #2209, opened 2023-09-13
Restrict Button scaling in Input Groups

### LINKED ISSUE #2141: Explore alternate button pointer effects
Our current button animation includes a scale effect on `:hover` and `:active`:

Image

The goal of this is to introduce just a little bit of surprise and whimsy into the typical static button styles.

This currently introduces some awkwardness for interfaces that rely on the button positions being static. For example, an input group where an input and button appears to _detach_ on `:hover` and `:active`:

Image Image

In a recent PR (#2140) I almost removed these styles entirely, but @Paul-Hebert had some interesting suggestions for alternate explorations: https://github.com/cloudfour/cloudfour.com-patterns/pull/2140#pullrequestreview-1308095888

> Some potential ideas that may or may not be worth exploring:
> 
> * Cranking up the filter a bit. (maybe `brightness(130) saturate(110)`?) Is that too much? We'd need to test text contrast, too... 🤔
> * Could we scale the button's _content_ without scaling the actual button? This might allow us to keep a similar energy to the current buttons, while side-stepping the issues with adjacent content
> * Giving the "white" background buttons (secondary and dark-mode primary) a subtle background color when hovered
> * Using different BG shades for all hovered buttons
> * Could we give buttons a gradient background when hovered?
> * Other fun hover effects? Border animations? I dunno...

tylersticka:
## Overview

Our newfound emphasis on newsletter signups has increased the prominence of our Input Group layout object, which has made a lingering issue more visible: On hover/active, our button scaling effects make the button appear disconnected from its adjacent input.

At one time I considered removing the scaling effects entirely, but @gerardo-rodriguez convinced me to reconsider. (See #2141)

I explored some alternative effects, but found they didn't work as well across the different interactive elements (pagination links, alert closure, etc.) that rely on the same tokens.

So I decided to expose those scale values as custom properties, so I could reset them _just_ for the Input Group object.

**TL;DR:** Buttons (and similar interactive elements) will continue to scale in most cases, except in Input Groups, when only the filter effect will apply.

## Screenshots

State | Before | After
--- | --- | ---
`:hover` | Screenshot 2023-09-13 at 2 26 54 PM | Screenshot 2023-09-13 at 2 26 22 PM
`:active` | Screenshot 2023-09-13 at 2 27 00 PM | Screenshot 2023-09-13 at 2 26 30 PM

## Testing

Try out deploy preview stories for Button and Input Group in:

- [ ] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] Safari

---

- Closes #2141

### DIFF
--- a/.changeset/dry-tables-leave.md
+++ b/.changeset/dry-tables-leave.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Prevent Buttons from feeling disconnected from Input Groups when `:hover` or `:active` states are applied
--- a/src/mixins/_button.scss
+++ b/src/mixins/_button.scss
@@ -90,16 +90,20 @@
   /**
    * Transform, opacity and filter changes are performant and will work across
    * many more stylistic modifiers than specific color shifts.
+   *
+   * We optionally support custom properties to allow the amount of scaling to
+   * be adjusted by other patterns. For example, an Input Group might disable
+   * the scaling so the button always appeared joined to adjacent elements.
    */
 
   &:hover {
     filter: brightness(brightness.$control-brighten);
-    transform: scale(scale.$effect-grow);
+    transform: scale(var(--scale-effect-grow, #{scale.$effect-grow}));
   }
 
   &:active {
     filter: brightness(brightness.$control-dim);
-    transform: scale(scale.$effect-shrink);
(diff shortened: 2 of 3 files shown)


Paul-Hebert (approved):
:shipit:

PR #2218, opened 2024-01-02
Add brand icon for Bluesky

tylersticka:
## Overview

[Bluesky updated their icon](https://blueskyweb.xyz/blog/12-21-2023-butterfly) and it works in monochrome, so we can add it to our brand icons.

(They use `bsky.app` and `blueskyweb.xyz` domains for different stuff so this may not work auto-magically in our theme without a new conditional.)

## Screenshots

Screenshot 2024-01-02 at 3 47 18 PM


Paul-Hebert (approved):
LGTM :+1:

PR #2222, opened 2025-02-05
Fixed ugly default borders on WordPress table blocks

tylersticka:
## Overview

At some point, WordPress shipped an update that applied a default `border-width` on table block `td` and `th` elements unless a block style was applied.

This adds `border-width: 0` to those elements by default so they'll be consistent with our table component again.

## Screenshots

Before | After
--- | ---
Screenshot 2025-02-05 at 10 23 30 AM | Screenshot 2025-02-05 at 10 24 06 AM

## Testing

For whatever reason, this issue isn't symptomatic with our imported Gutenberg styles. But it wouldn't hurt to check our stories for the WordPress table block on the deploy preview to make sure this didn't introduce regressions.

### DIFF
--- a/.changeset/green-windows-march.md
+++ b/.changeset/green-windows-march.md
@@ -1,0 +1,5 @@
+---
+'@cloudfour/patterns': patch
+---
+
+Fixed ugly default borders on WordPress table blocks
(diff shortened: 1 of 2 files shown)


tylersticka:
Note: One CI job is failing, but it seems to be due to a browser process issue? The same tests succeeded in three other CI jobs. 🤷

Paul-Hebert (approved):
Thanks WP! 🧌 

(Actually, thank you @tylersticka!)

That CI error looks familiar. Let me see if I can dig up how to fix it.

spaceninja (approved):

PR #2225, opened 2025-02-05
Add `text-wrap` to typographic elements

tylersticka:
## Overview

Sets some default `text-wrap` styles for ~~paragraphs, list items, figure captions,~~ headings and subheadings.

I marked this as a breaking change because it will conflict with the existing techniques in our theme for reducing orphans and widows. Those workarounds will need to be removed when applying this update.

## Screenshots

Before | After
--- | ---
Screenshot 2025-02-05 at 2 55 08 PM | Screenshot 2025-02-05 at 2 55 18 PM

### DIFF
--- a/.changeset/loud-paws-tease.md
+++ b/.changeset/loud-paws-tease.md
@@ -2,6 +2,4 @@
 '@cloudfour/patterns': major
 ---
 
-Improved text wrapping for headings and subheadings.
-
-This is a breaking change because it may disrupt existing techniques for avoiding orphans and widows in text (such as inserting `&nbsp;` or `<span class="u-inline-block">`). It's recommended to remove these workarounds.
+Improved text wrapping for headings and subheadings with `text-wrap`. Removes the `u-rescue-orphan` utility, which is now redundant.
--- a/.changeset/loud-paws-tease.md
+++ b/.changeset/loud-paws-tease.md
@@ -2,6 +2,6 @@
 '@cloudfour/patterns': major
 ---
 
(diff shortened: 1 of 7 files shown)


spaceninja (dismissed):
LGTM

MDN notes a performance impact from `text-wrap:pretty`:

> For longer sections of text, text-wrap: pretty can be used. Note that pretty has a negative effect on performance, so it should be only used for longer blocks of text when the layout is more important than speed.

I assume that you've considered this, but wanted to mention it just in case.

tylersticka:
> I assume that you've considered this, but wanted to mention it just in case.

That is a generous assumption! 😆 

I've sorta been taking it for granted, but my usage here does seem broader than in some other web dev sites I'm looking into.

I may reduce this down to just "improve text wrapping for headings." That part's a lot easier to see as a net positive compared to what we were doing before.

tylersticka:
I decided to scale back the scope of this change to just the `text-wrap: balance`. I think this portion is more noticeable, has a built-in performance failsafe (in that it won't take effect past a certain number of lines), and will reduce markup complexity in numerous places throughout our theme.

This also isolates the "breaking" part of the change, since it's only breaking because of the workarounds we are using in headings. If we want to revisit adding more `text-wrap` in the future, it can be a feature update.

Thanks for keeping me honest, @spaceninja!

tylersticka:
Converting to a draft so I can also remove the `u-rescue-orphan` utility I forgot we had.

spaceninja (approved):

PR #2232, opened 2025-02-25
Upgrade to Node 22

spaceninja:
This PR upgrades from Node 16 to Node 22 to keep up-to-date.


spaceninja:
Hrm. This is more complex than expected. Closing for now, will tackle again in the future when I have a bit more bandwidth.

